PR #39, opened 2020-03-19
Run as alpine docker image and server static files with gunicorn

h4llow3En:
First of all, I love this project! 

I still have some cleaning up to do but I wanted to know your opinion on this.

- [x] refactored the Dockerfile
- [x] Because it runs in a docker container I configured the static files to be serverd by gunicorn
- [x] only add relevant files to container
- [x] DB migration now runs on every startup
- [x] some cleaning up
- [x] reconfigure docker-compose examples to single docker-compose with an docker-compose.override.yml
- [x] mount staticfiles and mediafiles


h4llow3En:
Herko mentions this static files serving in their documentation
https://devcenter.heroku.com/articles/django-assets

vabene1111:
thanks for submitting this! I will take a look at it next week. Two questions:

1. did you already consider the changes i recently made with the different base image and removing the compose file from vcs ? I am not entirely sure by looking at the commit.

2. If we restructure it we could completely move it to a pre build image since that would be much cleaner than everybody having to build it themselves. There is already a repo but i have stopped it from building for now since something i cant remember promted me to not use a pre build image https://hub.docker.com/repository/docker/vabene1111/recipes I have triggered it to build the latest version again in case you want to test it.

Definitely thanks for the help, i am no docker expert and mostly just get things working so any help is appreciated.

h4llow3En:
About your first question: Yes. I forked this repo yesterday. So maybe I changed it back? In my opinion it is more intuitive to just adjust some lines in a `docker-compose.override.yml` than manually copying a `docker-compose.yml`.

The docker-compose should be working at itself. So you only have to clone and run `docker-compose up`. Everything else should be additional configuration.
(https://docs.docker.com/compose/extends/#understanding-multiple-compose-files)

I would appreciate a prebuild image. Since in my version of the docker-compose only the "static" content ist mounted you have to rebuild the image to make changes to the code. This is more release based. Also you/we build the docker image on every push with a CI e.g. Travis-CI so the `latest` tag is the latest commit and every release has its own release tag.

README.md
```diff
@@ -63,9 +62,7 @@ Otherwise simply follow the instructions for any django based deployment
 
 ## Updating
 0. Before updating it is recommended to **backup your database**
-1. Stop the container using `docker-compose down`.
```
h4llow3En: there is no explicit `docker-compose down` needed. A `docker-compose up` recreated the container if necessary. For example the database would not be recreated and keeps running

README.md
```diff
@@ -51,7 +51,7 @@ When clicking submit, every recipe containing the word will be updated (tags are
 2. Choose one of the included configurations [here](https://github.com/vabene1111/recipes/tree/develop/docs/docker).
 3. Copy (if needed) the `docker-compose.override.yml.template` to `docker-compose.override.yml` and uncomment the configurations you need.
 4. Start the container (`docker-compose up -d`)
-5. Create a default user by executing into the container with `docker-compose exec web_recipes sh` and run `python3 manage.py createsuperuser`.
+5. Create a default user by running `docker-compose exec -it <web_recipes> createsuperuser`. **Replace "<web_recipes>" with your instance name which can be found via running `docker ps`**
```
h4llow3En: Because of using a venv I simplified the superuser creation. Otherwise you would have to run `source venv/bin/activate` before executing `python manage.py createsuperuser`

README.md
vabene1111: are you sure of this ? i had mixed experiences when running docker compose up again when the containers where already up and also restart sometimes posed problems since it appeared to not behave 100% the same as down and up. But maybe that was just a bug ...

README.md
vabene1111: why use venv in a dockerized environment ? there is only one host running in the container anyway. since docker compose names the containers i dont think the replacing part is necessary (see update.sh)

vabene1111:
I will merge this and than change a few things as i prefer the setup with the different examples in the docs folder (mainly because there might be additional examples over time which will at some point become to many for one file and additionally because many other docker based projects handle it this way and thus makes it easy to understand).

i will also take a look at the static/media file serving if it makes sense to move this to gunicorn.

README.md
h4llow3En: There may be issues if you change for example the container names. But otherwise there should be no problems. It recreates if necessary

README.md
h4llow3En: It is a cleaner more bugfree approach.
https://vsupalov.com/virtualenv-in-docker/

h4llow3En:
As you wish, I only offered my opinion. :slightly_smiling_face: 

Moving static file serving to gunicorn means one less container running in the stack.

README.md
vabene1111: ok it appears to have some benefits. even though i find them rather slim. there is not much complexity added so i guess its fine.

vabene1111:
As is said, i am no docker expert so you opinion is very helpful, thank you. I was just holding off on deciding if venv was a goof idea but you explained that above so i will try to get this merged and adapted asap.

vabene1111:
ok so while merging it struc me that the nginx containers serverd another purpose other than serving static/media files: they are the proxy for gunicorn which, as the developers themself state, should not be run as the edge server (https://serverfault.com/questions/331256/why-do-i-need-nginx-and-something-like-gunicorn)

I suppose you did this on purpouse, why do you feel like this isn't needed ? I mean this application isn't meant for large scale deployments, but i guess the recommendation to have a proxy still holds ?!

h4llow3En:
I will use modified examples of the post from server fault

`||` will represent the start of docker environments

The recommendations are met if:
- you run the container behind Traefik 
  `The outside world <||> Traefik <-> The socket <-> Gunicorn`
- you run the container behind a global nginx in docker
  `The outside world <||> Nginx <-> The socket <-> Gunicorn`
- you run this container plain with port forwarding and have a nginx/apache/... running on your system serving as a proxy
  `The outside world <-> what ever is here <-> The socket <||> Gunicorn`

The recommendations are not met if:
- you run this container plain with port forwarding and you connect to it with no other proxy layer.

This is only my opinion but I think that you can document it as a recommendation if you intent on running the last option but otherwise you would end up with something like:
`The outside world (<-> maybe system proxy ) <||> Proxy <-> Proxy <-> The socket <-> Gunicorn`

vabene1111:
Ok perfect, thats what i was thinking about ... i think i will leave the nginx container in the plain example with the coment to remove it if its not needed due to a system wide proxy.

Thank you for all the input, just dont want to make to many changes to this in the future so wanna get this right now.

h4llow3En:
You're welcome

vabene1111:
So i am pretty sure you tested your changes but when i am trying to build the container i get the following error with `lxml`. I had the same previously, thats the reason i stopped using an alpine based image (also because many people recommended not to use them for python based applications).

Just wanted to make sure i am not missing anything, otherwise i would just revert the image to the slim-buster edition.

```
Running setup.py install for lxml: finished with status 'error'
    ERROR: Command errored out with exit status 1:
     command: /opt/recipes/venv/bin/python -u -c 'import sys, setuptools, tokeni                                                     ze; sys.argv[0] = '"'"'/tmp/pip-install-q88hkzrr/lxml/setup.py'"'"'; __file__='"                                                     '"'/tmp/pip-install-q88hkzrr/lxml/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'                                                     "', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close(                                                     );exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-recor                                                     d-h1xc9yub/install-record.txt --single-version-externally-managed --compile --in                                                     stall-headers /opt/recipes/venv/include/site/python3.8/lxml
         cwd: /tmp/pip-install-q88hkzrr/lxml/
    Complete output (85 lines):
    Building lxml version 4.5.0.
    Building without Cython.
    Using build configuration of libxslt 1.1.34
    Building against libxml2/libxslt in the following directory: /lib
    running install
    running build
    running build_py
    creating build
    creating build/lib.linux-x86_64-3.8
    creating build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/pyclasslookup.py -> build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/builder.py -> build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/sax.py -> build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/ElementInclude.py -> build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/__init__.py -> build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/_elementpath.py -> build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/usedoctest.py -> build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/doctestcompare.py -> build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/cssselect.py -> build/lib.linux-x86_64-3.8/lxml
    creating build/lib.linux-x86_64-3.8/lxml/includes
    copying src/lxml/includes/__init__.py -> build/lib.linux-x86_64-3.8/lxml/inc                                                     ludes
    creating build/lib.linux-x86_64-3.8/lxml/html
    copying src/lxml/html/builder.py -> build/lib.linux-x86_64-3.8/lxml/html
    copying src/lxml/html/defs.py -> build/lib.linux-x86_64-3.8/lxml/html
    copying src/lxml/html/html5parser.py -> build/lib.linux-x86_64-3.8/lxml/html
    copying src/lxml/html/diff.py -> build/lib.linux-x86_64-3.8/lxml/html
    copying src/lxml/html/__init__.py -> build/lib.linux-x86_64-3.8/lxml/html
    copying src/lxml/html/usedoctest.py -> build/lib.linux-x86_64-3.8/lxml/html
    copying src/lxml/html/clean.py -> build/lib.linux-x86_64-3.8/lxml/html
    copying src/lxml/html/formfill.py -> build/lib.linux-x86_64-3.8/lxml/html
    copying src/lxml/html/_html5builder.py -> build/lib.linux-x86_64-3.8/lxml/ht                                                     ml
    copying src/lxml/html/soupparser.py -> build/lib.linux-x86_64-3.8/lxml/html
    copying src/lxml/html/_diffcommand.py -> build/lib.linux-x86_64-3.8/lxml/htm                                                     l
    copying src/lxml/html/_setmixin.py -> build/lib.linux-x86_64-3.8/lxml/html
    copying src/lxml/html/ElementSoup.py -> build/lib.linux-x86_64-3.8/lxml/html
    creating build/lib.linux-x86_64-3.8/lxml/isoschematron
    copying src/lxml/isoschematron/__init__.py -> build/lib.linux-x86_64-3.8/lxm                                                     l/isoschematron
    copying src/lxml/etree.h -> build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/etree_api.h -> build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/lxml.etree.h -> build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/lxml.etree_api.h -> build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/includes/relaxng.pxd -> build/lib.linux-x86_64-3.8/lxml/inc                                                     ludes
    copying src/lxml/includes/htmlparser.pxd -> build/lib.linux-x86_64-3.8/lxml/                                                     includes
    copying src/lxml/includes/xslt.pxd -> build/lib.linux-x86_64-3.8/lxml/includ                                                     es
    copying src/lxml/includes/schematron.pxd -> build/lib.linux-x86_64-3.8/lxml/                                                     includes
    copying src/lxml/includes/xpath.pxd -> build/lib.linux-x86_64-3.8/lxml/inclu                                                     des
    copying src/lxml/includes/xmlerror.pxd -> build/lib.linux-x86_64-3.8/lxml/in                                                     cludes
    copying src/lxml/includes/config.pxd -> build/lib.linux-x86_64-3.8/lxml/incl                                                     udes
    copying src/lxml/includes/xmlschema.pxd -> build/lib.linux-x86_64-3.8/lxml/i                                                     ncludes
    copying src/lxml/includes/__init__.pxd -> build/lib.linux-x86_64-3.8/lxml/in                                                     cludes
    copying src/lxml/includes/uri.pxd -> build/lib.linux-x86_64-3.8/lxml/include                                                     s
    copying src/lxml/includes/xmlparser.pxd -> build/lib.linux-x86_64-3.8/lxml/i                                                     ncludes
    copying src/lxml/includes/dtdvalid.pxd -> build/lib.linux-x86_64-3.8/lxml/in                                                     cludes
    copying src/lxml/includes/xinclude.pxd -> build/lib.linux-x86_64-3.8/lxml/in                                                     cludes
    copying src/lxml/includes/etreepublic.pxd -> build/lib.linux-x86_64-3.8/lxml                                                     /includes
    copying src/lxml/includes/tree.pxd -> build/lib.linux-x86_64-3.8/lxml/includ                                                     es
    copying src/lxml/includes/c14n.pxd -> build/lib.linux-x86_64-3.8/lxml/includ                                                     es
    copying src/lxml/includes/lxml-version.h -> build/lib.linux-x86_64-3.8/lxml/                                                     includes
    copying src/lxml/includes/etree_defs.h -> build/lib.linux-x86_64-3.8/lxml/in                                                     cludes
    creating build/lib.linux-x86_64-3.8/lxml/isoschematron/resources
    creating build/lib.linux-x86_64-3.8/lxml/isoschematron/resources/rng
    copying src/lxml/isoschematron/resources/rng/iso-schematron.rng -> build/lib                                                     .linux-x86_64-3.8/lxml/isoschematron/resources/rng
    creating build/lib.linux-x86_64-3.8/lxml/isoschematron/resources/xsl
    copying src/lxml/isoschematron/resources/xsl/RNG2Schtrn.xsl -> build/lib.lin                                                     ux-x86_64-3.8/lxml/isoschematron/resources/xsl
    copying src/lxml/isoschematron/resources/xsl/XSD2Schtrn.xsl -> build/lib.lin                                                     ux-x86_64-3.8/lxml/isoschematron/resources/xsl
    creating build/lib.linux-x86_64-3.8/lxml/isoschematron/resources/xsl/iso-sch                                                     ematron-xslt1
    copying src/lxml/isoschematron/resources/xsl/iso-schematron-xslt1/iso_schema                                                     tron_message.xsl -> build/lib.linux-x86_64-3.8/lxml/isoschematron/resources/xsl/                                                     iso-schematron-xslt1
    copying src/lxml/isoschematron/resources/xsl/iso-schematron-xslt1/iso_dsdl_i                                                     nclude.xsl -> build/lib.linux-x86_64-3.8/lxml/isoschematron/resources/xsl/iso-sc                                                     hematron-xslt1
    copying src/lxml/isoschematron/resources/xsl/iso-schematron-xslt1/iso_svrl_f                                                     or_xslt1.xsl -> build/lib.linux-x86_64-3.8/lxml/isoschematron/resources/xsl/iso-                                                     schematron-xslt1
    copying src/lxml/isoschematron/resources/xsl/iso-schematron-xslt1/iso_schema                                                     tron_skeleton_for_xslt1.xsl -> build/lib.linux-x86_64-3.8/lxml/isoschematron/res                                                     ources/xsl/iso-schematron-xslt1
    copying src/lxml/isoschematron/resources/xsl/iso-schematron-xslt1/iso_abstra                                                     ct_expand.xsl -> build/lib.linux-x86_64-3.8/lxml/isoschematron/resources/xsl/iso                                                     -schematron-xslt1
    copying src/lxml/isoschematron/resources/xsl/iso-schematron-xslt1/readme.txt                                                      -> build/lib.linux-x86_64-3.8/lxml/isoschematron/resources/xsl/iso-schematron-x                                                     slt1
    running build_ext
    building 'lxml.etree' extension
    creating build/temp.linux-x86_64-3.8
    creating build/temp.linux-x86_64-3.8/src
    creating build/temp.linux-x86_64-3.8/src/lxml
    gcc -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -DTHREAD                                                     _STACK_SIZE=0x100000 -fPIC -DCYTHON_CLINE_IN_TRACEBACK=0 -I/usr/include/libxml2                                                      -Isrc -Isrc/lxml/includes -I/opt/recipes/venv/include -I/usr/local/include/pytho                                                     n3.8 -c src/lxml/etree.c -o build/temp.linux-x86_64-3.8/src/lxml/etree.o -w
    gcc: fatal error: Killed signal terminated program as
    compilation terminated.
    Compile failed: command 'gcc' failed with exit status 1
    creating tmp
    cc -I/usr/include/libxml2 -I/usr/include/libxml2 -c /tmp/xmlXPathInitujol2qi                                                     j.c -o tmp/xmlXPathInitujol2qij.o
    cc tmp/xmlXPathInitujol2qij.o -L/lib -lxml2 -o a.out
    error: command 'gcc' failed with exit status 1
    ----------------------------------------
ERROR: Command errored out with exit status 1: /opt/recipes/venv/bin/python -u -                                                     c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-q88hkzr                                                     r/lxml/setup.py'"'"'; __file__='"'"'/tmp/pip-install-q88hkzrr/lxml/setup.py'"'"'                                                     ;f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'                                                     \r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))                                                     ' install --record /tmp/pip-record-h1xc9yub/install-record.txt --single-version-                                                     externally-managed --compile --install-headers /opt/recipes/venv/include/site/py                                                     thon3.8/lxml Check the logs for full command output.
WARNING: You are using pip version 19.2.3, however version 20.0.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

vabene1111:
interestingly enough the build i ran on docker hub ran trough (not sure if you can view that) https://hub.docker.com/repository/registry-1.docker.io/vabene1111/recipes/builds/b4447474-c206-4f00-985c-b93fdb70bedd

h4llow3En:
Very strange... I tested building this with Docker for Mac and Docker on Linux.
The only thing I did not test was Docker for Windows...
It build everytime without any issues...

vabene1111:
Interesting .. i am building on linux ubuntu (` 4.15.0-65-generic #74-Ubuntu SMP Tue Sep 17 17:06:04 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux`).

vabene1111:
updating the kernel seems to have fixed the problem .. weird but ok

h4llow3En:
I used Kernel 5.0.0-38-generic on Ubuntu and 5.5.10-arch1-1 on arch to test this. So this might be the solution.

vabene1111:
So everything is basically working. One problem is left: the boot.sh script creates the `staticfiles` directory and therefor cannot access the files inside. i am not sure if this is related to the warning in the docker documentation (https://docs.docker.com/engine/reference/builder/#user) that if no primary group exists the root user will be taken because the recipes user has a group (recipes).

Maybe you have an idea on why this happens.

docker logs (latest version):
```
Updating database
web_recipes_1  |
web_recipes_1  | Operations to perform:
web_recipes_1  |   Apply all migrations: admin, auth, contenttypes, cookbook, sessions
web_recipes_1  | Running migrations:
web_recipes_1  |   No migrations to apply.
web_recipes_1  | Traceback (most recent call last):
web_recipes_1  |   File "manage.py", line 15, in <module>
web_recipes_1  |     execute_from_command_line(sys.argv)
web_recipes_1  |   File "/opt/recipes/venv/lib/python3.8/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
web_recipes_1  |     utility.execute()
web_recipes_1  |   File "/opt/recipes/venv/lib/python3.8/site-packages/django/core/management/__init__.py", line 395, in execute
web_recipes_1  |     self.fetch_command(subcommand).run_from_argv(self.argv)
web_recipes_1  |   File "/opt/recipes/venv/lib/python3.8/site-packages/django/core/management/base.py", line 328, in run_from_argv
web_recipes_1  |     self.execute(*args, **cmd_options)
web_recipes_1  |   File "/opt/recipes/venv/lib/python3.8/site-packages/django/core/management/base.py", line 369, in execute
web_recipes_1  |     output = self.handle(*args, **options)
web_recipes_1  |   File "/opt/recipes/venv/lib/python3.8/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 187, in handle
web_recipes_1  |     collected = self.collect()
web_recipes_1  |   File "/opt/recipes/venv/lib/python3.8/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 113, in collect
web_recipes_1  |     handler(path, prefixed_path, storage)
web_recipes_1  |   File "/opt/recipes/venv/lib/python3.8/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 348, in copy_file
web_recipes_1  |     self.storage.save(prefixed_path, source_file)
web_recipes_1  |   File "/opt/recipes/venv/lib/python3.8/site-packages/django/core/files/storage.py", line 52, in save
web_recipes_1  |     return self._save(name, content)
web_recipes_1  |   File "/opt/recipes/venv/lib/python3.8/site-packages/django/core/files/storage.py", line 248, in _save
web_recipes_1  |     os.makedirs(directory, exist_ok=True)
web_recipes_1  |   File "/usr/local/lib/python3.8/os.py", line 223, in makedirs
web_recipes_1  |     mkdir(name, mode)
web_recipes_1  | PermissionError: [Errno 13] Permission denied: '/opt/recipes/staticfiles/autocomplete_light'
web_recipes_1  | Done
```
directory
```
/opt/recipes $ ls -al
total 56
drwxr-xr-x    1 recipes  recipes       4096 Mar 24 18:01 .
drwxr-xr-x    1 root     root          4096 Mar 24 17:50 ..
-rwxr-xr-x    1 recipes  recipes        225 Mar 24 17:50 boot.sh
drwxr-xr-x    1 recipes  recipes       4096 Mar 24 18:01 cookbook
-rw-r--r--    1 recipes  recipes        539 Mar 24 17:50 manage.py
drwxr-xr-x    3 root     root          4096 Mar 24 16:47 mediafiles
drwxr-xr-x    1 recipes  recipes       4096 Mar 24 18:01 recipes
-rw-r--r--    1 recipes  recipes        270 Mar 24 17:50 requirements.txt
-rwxr-xr-x    1 recipes  recipes        106 Mar 24 17:50 setup.sh
drwxr-xr-x    2 root     root          4096 Mar 24 17:27 staticfiles
drwxr-xr-x    5 root     root          4096 Mar 24 17:51 venv
/opt/recipes $
```

h4llow3En:
I stumbled over this issue. If the folders dont exist while running `docker-compose up` they are created because they have to be mounted. This is an issue by docker itself.
If a path does not exist it will be created as root user and if the path on the host points to a file, and that file is missing, docker will create a directory in that location.
Its a bit annoying. You wouldn't have this issue if you would run the container with a root user but this is never a good idea...

Best solution I found: Either create these files as your user before running docker-compose or run `chown -R 1000:1000 staticfiles mediafiles`

vabene1111:
hmm but this isn't really a solution for normal people setting this up if we aim for simplicity .. i will try if i can create and own the folders at the time of building the image (in the Dockerfile), maybe that helps

h4llow3En:
Not really because mounting overrides it. You could keep those folders in git ignoring all content except a .gitkeep in both directories...

So touch a .gitkeep in both directories an add something like that to .gitignore :
```
staticfiles
!staticfiles/.gitkeep
mediafiles
!mediafiles/.gitkeep
```

h4llow3En:
I may have a solution. If this works I would make a separate PR

h4llow3En:
I have seed you worked a bit on it. So I don't want to destroy your structure again. Simply add this to boot.sh after `#!/bin/sh`

```
# Change ownership if needed
if [ ! -n "$(find staticfiles -user "$(id -u)" -print -prune -o -prune)" ]
then
    echo "Changing ownership of staticfiles"
    chown -R recipes:recipes staticfiles
fi
if [ ! -n "$(find mediafiles -user "$(id -u)" -print -prune -o -prune)" ]
then
    echo "Changing ownership of mediafiles"
    chown -R recipes:recipes mediafiles
fi
```

vabene1111:
perfect! i screwed some stuff up and will revert it but i can add this. thanks for helping

vabene1111:
it appears that the script is not working as expected 

```
 chown: staticfiles: Operation not permitted
web_recipes_1  | chown: staticfiles: Operation not permitted
web_recipes_1  | chown: mediafiles/recipes/278.jpeg: Operation not permitted
web_recipes_1  | chown: mediafiles/recipes/495.jpeg: Operation not permitted
web_recipes_1  | chown: mediafiles/recipes/439_1R4zMu2.jpeg: Operation not permitted
web_recipes_1  | chown: mediafiles/recipes/507.jpeg: Operation not permitted
web_recipes_1  | chown: mediafiles/recipes/365_pqlg01d.png: Operation not permitted
web_recipes_1  | chown: mediafiles/recipes/395.jpeg: Operation not permitted
web_recipes_1  | chown: mediafiles/recipes: Operation not permitted
web_recipes_1  | chown: mediafiles/recipes: Operation not permitted
web_recipes_1  | chown: mediafiles: Operation not permitted
web_recipes_1  | chown: mediafiles: Operation not permitted
web_recipes_1  | Changing ownership of staticfiles
web_recipes_1  | Changing ownership of mediafiles
web_recipes_1  | Updating database
```

i have no problems running it as root, i dont see that this is too much of problem, but if you have any other ideas feel free to  submit them :)

h4llow3En:
If you feel confident running this as root, change the user back.
Just remove in the Dockerfile:
- The user creation (Line 7/8)
- the `--chown=recipes:recipes` from the copy in line 13
- line 22 the change of the user

And remove the chown from boot.sh as wenn ;)

vabene1111:
its definitely better without root but ok, i will release this for now and maybe we find a solution later

PR #45, opened 2020-03-30
Add buttons to add a meal plan to a specific point in time

tourn:
Hey, hope you don't mind me barging in :) I've been meaning to replace my own homegrown unfinished recipe manager and stumbled upon yours. I might run this for myself with some modifications and figured you might like some of them. 

This adds some buttons to the meal planner that open the new-meal-plan form with the `date` and `meal` fields prefilled.

image

### DIFF
--- a/cookbook/templates/meal_plan.html
+++ b/cookbook/templates/meal_plan.html
@@ -8,6 +8,25 @@
 {% endblock %}
 
 {% block content %}
+    <style>
+        .mealplan-cell .mealplan-add-button{
+            text-align: center;
+            display: block;
+        }
+
+        @media (hover: hover) {
+            .mealplan-cell .mealplan-add-button{
+                visibility: hidden;
+                float: right;
+                display: inline;
+            }
+
+            .mealplan-cell:hover .mealplan-add-button{
+                visibility: initial;
+            }
+        }
+
+    </style>
 
     <h3>
         {% trans 'Meal-Plan' %} <a href="{% url 'new_meal_plan' %}"><i class="fas fa-plus-circle"></i></a>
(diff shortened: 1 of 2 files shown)


vabene1111:
Hey, thanks for your PR. The meal plan generally is pretty unpolished/unfinished as it is a feature i personally dont use and have just implement since someone wished for it so contributions/suggestions from people actually using it are very welcome.

I kinda feel like this make the thing pretty full of buttons, i am no java script wizard but what do you think about adding the buttons in a way that they only show up when you hover over the cell ? Not sure though if this would work on mobile ...

If something like this isn't possible what do you think about centering them so they are more in line with the table headers ?

I have no time to look at the code right now and will give feedback (if needed) probably tomorrow.

tourn:
Sure, I can make the buttons a bit less prominent or hide them using CSS unless you hover, excluding touch devices based on screen size (although that one always is a bit iffy).
Personally I don't mind the buttons as they are, because as soon as you start filling up your plan it's full of buttons/links for the actual menus anyway.

I'll post an update if I come up with something that looks nicer when i get the time.

tourn:
Okay, so i've added some CSS. I've jammed it right into the meal_plan.html template since i'm quite new to django and couldn't find any other existing CSS except for the minified themes, so any pointers on doing that nicer are appreciated.

This is what it looks like on devices that have a mouse, with the pointer where the plus sign is showing

image

And on touch devices, it looks like this:

image

vabene1111:
i really like this 👍 

You can put CSS in the base template which is included (or actually extended by) the meal_plan.html but since this css is only used on the meal_plan page its probably best/correct to put it there.

PR #46, opened 2020-04-06
Updated character limit for emojis in keywords

cazier:
I'm unable to select a country's flag as an emoji in the keyword editor, because the field has a maximum length of 1 character.

As far as I can tell, having read [this page on Emojipedia](https://blog.emojipedia.org/emoji-flags-explained/), this is because, as a fallback, flags are encoded as a minimum of 2 characters, both on devices that do and don't display the actual flag icons. I was able to "fix" this by changing https://github.com/vabene1111/recipes/blob/4cf6a3b219680aeb13500e63a5333f3c7515fd54/cookbook/models.py#L93 to `max_length=2`, which allows me to use _almost_ all flags.

Unfortunately, it does introduce a bit of a different issue, in that flags like the Welsh flag require 14(!) characters. So I'm not sure what the perfect value for `max_length` really is. I think that 2, which allows very close to nearly every flag is probably sufficient, but I do apologize in advance to fans of Welsh cuisine.

Please note also, that this does mean that it is now possible to have more than one emoji character, if both of those characters are _not_, a flag, but, again, I think that's probably okay...

I didn't really think this needed an official pull request, but made one that has both the change and a db migration.


vabene1111:
Thanks for the PR.

The length limitation was also an issue sometimes when changing the emoji as the picker plugin sometimes messed up with invisible characters. 

If i think about it i dont even really see a problem in removing the length limitation (or setting it to something like 16) because in the end its the users decision if they want to put many emojis in a keyword.

vabene1111:
ok i have tested a little and it seems like having multiple emojis does not break anything so i set the limit to 16 characters 🏴󠁧󠁢󠁷󠁬󠁳󠁿

I will close this PR because if i merge it we will have two migrations but thank you very much for pointing out this issue and taking the time to create a PR!

PR #51, opened 2020-04-13
added 'webdav_root' option

pataya23:
otherwise I get an error (webdav3.exceptions.RemoteResourceNotFound: Remote resource: </path> not found)

### DIFF
--- a/cookbook/provider/nextcloud.py
+++ b/cookbook/provider/nextcloud.py
@@ -16,9 +16,10 @@ class Nextcloud(Provider):
     @staticmethod
     def get_client(storage):
         options = {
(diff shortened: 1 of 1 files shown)


vabene1111:
what is the nextcloud/owncloud/webdav server (and version) you are using ? because it works for me with my nextcloud 18.02

But i will test it and if it works for my version as well i see no reason to change it

pataya23:
same here, nextcloud with 18.0.2. interesting...

vabene1111:
ok, i will quickly test it and if it works i see no reason to keep the old Version as the `webdav_root` option seems to be exactly for this.

vabene1111:
ok i have tested it with my nextcloud and it seems to work as expected, thanks for the PR!

PR #57, opened 2020-04-13
Allow mealplan items to have no recipes

tourn:
This allows leaving the recipe blank in the mealplan. Additionally, the first line of the note is displayed next to the recipe name. This allows to plan for meals that are not really recipes, like leftovers etc.

image

In this example the top left entry is just a note with no recipe.


vabene1111:
So from looking at the code this always displays the first line of the note, even if a recipe is linked is that correct ? 

What do you think about adding an additional field, e.g. `title` and make it so that either the title or the recipe must be set (in a form validator). I think this would be a little more intuitive as a user would not really guess the first line feature without reading some kind of instructions and allow for more flexibility with the note content.

tourn:
Yep, that would be more intuitive for sure. This was just my quick sunday evening hack to get what I want :) I'll look into implementing this in a cleaner way soon.

vabene1111:
Sure, your solution is very good as well and the least intrusive one by the means of code changes that could break anything but i think its worth going the extra few steps here.

PR #58, opened 2020-04-13
Add serving count to recipe

tourn:
This adds a number field for the number of servings creating a recipe.

image

This number is then used as a base for scaling the ingredient amounts in the recipe view. This improves scaling for recipes that have more than 1 serving.

image

image


vabene1111:
i have often thought about adding a feature like this but in the end decided against it. 

Why: All the people i know eat very different amounts of food and most times i open a recipe the "portions" or "serving count" or whatever it is called information is wrong to a degree where it is more misleading to exist than it helps. 

I see that other people might find this useful and i really like your implementation with adding it to the calculator but i kinda dont want to add it or at least would love to have it as an optional field (which would break the integration with the calculator) ...

I will have to think about this a little ...

tourn:
Of course, sadly, every recipe has a different idea of how much a serving is.
But since this is supposed to be my personal recipe database I can always adjust the ingredients (or the serving count!) according to my experience.
Since the calculator feature is already there, the serving count just makes it super convenient to add a recipe that says "4 servings" so the calculator works correctly.  
 Without my patch the calculator goes from 1 (= 4 servings) to 2(= 8 servings).
 With the patch the caculator starts at 4 (= 4 servings) and can be adjusted to any number, like 2 = 2 servings.

If you don't want to use it that's fine, i'm going to run it on my own fork anyway. If we agree on an implementation and keep the same code base then that's preferrable, though.

vabene1111:
yes definitely, as long as its possible i want to support everyone wishes without them having to deviate from upstream so they can benefit from newly made changes.

I totally understand your reasoning but i am still not really sure. When thinking about this the last time i thought about adding a feature were every user could submit a serving count and some kind of average would be build but i think that would break features like the calculator integration.

Another thing that would be possible is for every user to add his own serving counts where the recipe would be created without a number of serving and a user could then add the serving count. That would also integrate nicely with a feature i thought about that was something like a cooing logbook where you could, after cooking a meal, add a quick entry with some notes. This hasn't been implemented and thats what the comments are for currently but i think it would be nice (for example one could note which sides were used).

auanasgheps:
Hey, what's the status on this one? 
I'm a new user and I find confusing the current calculator, because there's no place to state how many portions will be served with such amount, and the calculator will just double the figure, not scale it.

DaFlowah:
I'd like to see this included.
Might need to brainstorm a bit how to integrate non-fractionable ingredients (like eggs: if a recipe requires an egg and it's for two servings, would be difficult to add 1.5 eggs. I love the Internation Unit System, but let's face the truth: no one weights eggs in grams, or any other imaginary unit systems).

tourn:
I'm currently running this on my own fork, but I'd rather not pull users
away from the main repository, especially since I'm slow to merge in new
updates and migrating the existing database is not always trivial.

Thinking about it, it shouldn't be too intrusive to merge my code into here
and add a configuration option to hide the serving count on the recipe
(basically setting it always to 1 by default). I might be able to look into
it over the holidays, but no promises. If anyone else feels up to it, feel
free to go ahead.

Regarding the fractions: Personally I don't really mind when a scaled
recipe gets weird numbers. I'd rather see 1.5 eggs on my shopping list than
trying to come up with some rounding mechanism that makes things more
complicated. (In that specific case I'd know to get 2 small eggs, or a
really big one instead)

auanasgheps:
Thanks @tourn for working on this.
I agree to have to see something like 1.5 eggs in the list.
Maybe we could add an option to flag ingredients that can't be easily divided, then show the exact calculation with a warning message "Hey, probably this doesn't make much sense because you said it can't be divided".

vabene1111:
Thanks @tourn for stepping in here. 

We discussed why the PR isn't merged yet in the discussion and i had some ideas on how to get the different standpoints together. 

To be honest my stance on the whole topic changed over the time i have been developing this application. When i started (and still for me personally) and dont see any use in serving counts as they are basically always wrong (if someone else eats, if you have more or less hunger, ...). But as this application has gained popularity i have to admint that many people really need this feature and it is definitly high on my priority list.

I will take @tourn PR as a base and add a few things (mainly that giving a serving count in cook log can caluclate an average serving over time or is at least displayed somewhere) and so that per user servings are possible. 

The rest should be straight forward (also i still love you approach that the calculator shows the servings directly). One thing that we need to take into account is the scaling in the meal plan and shopping list need to be adjusted as well if we change the base of the scaling from now always one to then the number of servings.

tourn:
Thanks for picking this up!

> One thing that we need to take into account is the scaling in the meal plan and shopping list need to be adjusted as well if we change the base of the scaling from now always one to then the number of servings.

If I understand you correctly, this is already taken care of. The meal plan and shopping list already scale correctly based on the configured serving count. The only awkward thing about it is that the current meal planning UI cannot pre-fill the correct number of servings for a recipe:

image

So if in the above Example I drag the Süsskartoffel-Auflauf into my meal plan, it will be planned 4 servings because that's its default serving count. If I manually set the multiplier (should probably be called "Number of servings" in this case) to say, 2, it will plan the same recipe with 2 servings. So it already works well in my opinion, it's just not very obvious/discoverable/user friendly.
This is also correctly transferred to the shopping list.

> that per user servings are possible.

This is a great idea. It might be worthwhile to keep the serving count on the recipe, just to have a default value, and using the personal "override" serving count of the user if they created one.

vabene1111:
ok i wasn't that deep into what you do and tough meal plan and shopping lists would need changing.

I would definitely keep the default and just add the user based on top of that. This might also be done directly at the model level with a custom method overriding the serving count if there is one custom to the user.

Do you want to take a look at bringing this PR to the current version and implement the few missing QOL features or should i do this ?

tourn:
All right, the PR is now up to date with the current develop branch.

- renamed the recipe multiplier label to "Serving Count"
- Displaying default serving count as a badge on the recipe cards (might have change it from "4" to "4 servings" to make it clearer)
- Displaying serving count on Meal-Plan-Entry popup

image

Things that handle serving count with this PR

- "Add to Shopping" from recipe view
- Meal Plan
  - including "Open Shopping list"
- Shopping list

Things I leave up to @vabene1111 
- The by-user serving override
- A config to hide all the serving count things (they all default to 1 and everything should work as before)
- Updating translations
  - the "Edit Meal-Plan" form still shows "Recipe Multiplier" as label
  - probably many more, especially in foreign languages
 
@vabene1111 let me know if that works for you or if there's anything else I can help you with.

Cheers!

vabene1111:
Sounds really good 👍 I hope I will find some time to merge this and add the few missing pieces soon.

Thank you very much for your help

PR #80, opened 2020-06-01
Made recipe images hyperlinked

cazier:
Attached is a suggested change to make it so you can click on the picture of the recipe (or the blank image supplied) to proceed to the instructions. As of right now, you have to click on the name of the recipe, and the image has no hyperlink. (Note, this only works at all if using the "Search Style: Large" setting.)

I'm honestly not super well-versed on Django so it was a bit of a stumble through. I also broke out the added elements with variables since much of the tags are reused. Obviously those could be simplified into one line (in [cookbook/tables.py](https://github.com/vabene1111/recipes/blob/445c01bddc6bbf3f195d6cec67dcd18e656629ed/cookbook/tables.py#L11)) as so:

```
    def render(self, value):
        if value.url:
            return mark_safe(f'<img src="{value.url}" alt="{_("Recipe Image")}" class="card-img" style="object-fit: cover; height:130px">')
        return mark_safe(f'<img src="{static("recipe_no_image.svg")}" alt="{_("Recipe Image")}" class="card-img d-none d-lg-block" style="object-fit: inherit; height:130px">')
```

Finally, if you were more interested in just having the changes in one file, there is also the option to edit [cookbook/templates/recipes_table.html](https://github.com/vabene1111/recipes/blob/445c01bddc6bbf3f195d6cec67dcd18e656629ed/cookbook/templates/recipes_table.html#L16) as follows:

```
<div class="col-md-4">
    <a href="/view/recipes/{{ row.cells.id }}">
        {% if row.cells.image|length > 1 %}
            <img src=" {{ row.cells.image }}" alt="{% trans 'Recipe Image' %}"
                    class="card-img" style="object-fit: cover;height: 130px">
        {% else %}
            <img src="{% static 'recipe_no_image.svg' %}"
                    alt="{% trans 'Recipe Image' %}"
                    class="card-img d-none d-lg-block"
                    style="object-fit: inherit; height: 130px">

        {% endif %}
    </a>
</div>
```

Both add links, but one has more of a requirement on hard-coding the URL, while the other adds HTML to the Python code, which I didn't know if it was all that desirable...

Thoughts? Github doesn't really give me the option to show all three different possibilities without making different branches for each, but I'd be happy to provide them that way, if you'd like.

### DIFF
--- a/cookbook/tables.py
+++ b/cookbook/tables.py
@@ -5,12 +5,22 @@ from django_tables2.utils import A  # alias for Accessor
 
 from .models import *
 
+from django.utils.safestring import mark_safe
+from django.templatetags.static import static
 
-class ImageUrlColumn(tables.Column):
+class ImageUrlColumn(tables.LinkColumn):
     def render(self, value):
         if value.url:
-            return value.url
-        return None
+            src = value.url
+            _class = ''
+            style = 'object-fit: cover;'
+        
+        else:
+            src = static("recipe_no_image.svg")
+            _class = 'd-none d-lg-block'
+            style = 'object-fit: inherit;'
+            
+        return mark_safe(f'<img src="{src}" alt="{_("Recipe Image")}" class="card-img {_class}" style="{style} height:130px">')
 
 
 class RecipeTableSmall(tables.Table):
@@ -30,7 +40,7 @@ class RecipeTable(tables.Table):
     name = tables.LinkColumn('view_recipe', args=[A('id')])
     all_tags = tables.Column(
         attrs={'td': {'class': 'd-none d-lg-table-cell'}, 'th': {'class': 'd-none d-lg-table-cell'}})
-    image = ImageUrlColumn()
+    image = ImageUrlColumn('view_recipe', args=[A('id')])
 
     class Meta:
         model = Recipe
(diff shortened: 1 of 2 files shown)


vabene1111:
I really like this idea, thanks for the PR! 

I have decided to go with option 3. The reason beein that django tables tries to be really smart with its columns and does not call the render method for fields without values. This leads to the `recipe_no_image.svg` not beieng renderd correctly when using option one and is also the reason for the stupid ` {% if row.cells.image|length > 1 %}` "hack". There is most likely a better Solution but this works for now.

I also solved the URL hard coding problem, take a look at b8821f1f7285064084aead6e9c033872d2fab484 and you will see that this can very easily be done with django, in case you want to play a little more with it.

cazier:
I see what you mean about the URL hardcoding. That's pretty nifty. Having mostly done tiny projects with things like Flask, the crazy amount of configuration and added features in Django is nuts...

Thanks for a great project! I'm a huge fan of it!

vabene1111:
Glad you like it :) I can definitly recommend trying out/learning django, it is incredibly powerful, has a great documentation and community and once you get you head wrapped around a few edges it is really easy.

Issue #81, opened 2020-06-01
Reverse Proxy/External Authentication


cazier:
Feel free to file this issue under the "only affects a few people" category, but I was wondering if you'd be interested in adding functionality to log in using, as Django describes it, [REMOTE_USER](https://docs.djangoproject.com/en/3.0/howto/auth-remote-user/).

My current set up uses nginx as a reverse proxy, and [Authelia](https://github.com/authelia/authelia) as an SSO provider. In short, if I try to use a subdomain that is protected by Authelia, it requires me to a) be logged in, and b) have access (configured in Authelia) for that site.

Django provides the functionality to log-in via a reverse proxy header by using a specific Authentication Backend, and reading the variable passed through the header, simply by adding a few lines in the `settings.py` file and creating a custom middleware. (Though the last step can be optional.)

Is this something you would be willing to implement upstream? I'm happy to create a pull request, but, while I think it's quite basic, I am aware it's not something that would necessarily be applicable to all users.

Thanks!

vabene1111:
That sounds very interesting and also like something many self hosting enthusiasts might already have setup. As i dont have a setup of my own with authelia so feel free to create a PR with something that works for your setup. I think it would be best to have as much of it configurable trough the `.env` file as possible so that people with other auth frameworks can use it as well.

I can then review the code and read up on the docs if there are possible security or feature limitations and then we can merge it.

vabene1111:
so in 0df86b940f2265decb96d75c57a3738c35a2776a i added an online setup page as some people apparently dont have shell access to their containers. I have added a condition so that the page should not be shown if remote userers are activated but please make sure to thoroughly validate this (https://github.com/vabene1111/recipes/commit/0df86b940f2265decb96d75c57a3738c35a2776a#diff-3a3a6e110e53cfdd2cdab24960feb6eeR265).

PR #88, opened 2020-06-03
Adding Reverse Proxy Authentication

cazier:
See #81.

This Pull Request will allow you to log in to recipes by looking for a matching request header. If there is no matching header, Django automatically falls back to using the built-in/standard User/Password login functionality of `django.contrib.auth.backends.ModelBackend` (i.e., what is currently used.)

If the user _does not exist in recipes_, but does have access granted via the authentication server, Django silently lets the user access the site, but will not _create_ that user in the database:

image

I tried to add some documentation/deployment details as well, but, given this also requires a functioning authentication server, which could have lots of different implementations, it's rather sparse. I'm happy to expand upon that, though, if needed.

Let me know if there any issues!

### DIFF
--- a/docs/docker/nginx-proxy with proxy authentication/README.md
+++ b/docs/docker/nginx-proxy with proxy authentication/README.md
@@ -1,0 +1,15 @@
+This is a further example combining the power of nginx with the reverse proxy authentication service, [Authelia](https://github.com/authelia/authelia).
+
+Please refer to the appropriate documentation on how to setup the reverse proxy, authentication, and networks.
+
+Ensure users have been configured for Authelia, and that the endpoint that recipes is pointed to is protected, but available.
+
+There is a good guide to the other additional files that need to be added to your Nginx set up at the [Authelia Docs](https://docs.authelia.com/deployment/supported-proxies/nginx.html).
+
+Remember to add the appropriate environment variables to `.env` file:
+```
+VIRTUAL_HOST=
+LETSENCRYPT_HOST=
+LETSENCRYPT_EMAIL=
+PROXY_HEADER=
+```
--- a/docs/docker/nginx-proxy with proxy authentication/docker-compose.yml
+++ b/docs/docker/nginx-proxy with proxy authentication/docker-compose.yml
@@ -1,0 +1,43 @@
+version: "3"
+services:
+  db_recipes:
+    restart: always
+    image: postgres:11-alpine
+    volumes:
+      - ./postgresql:/var/lib/postgresql/data
+    env_file:
+      - ./.env
+    networks:
+      - default
+
+  web_recipes:
+    image: vabene1111/recipes
+    restart: always
+    env_file:
+      - ./.env
+    volumes:
+      - ./staticfiles:/opt/recipes/staticfiles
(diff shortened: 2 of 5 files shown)


vabene1111:
since i cannot really test this setup, have you tested this (i.e. by accessing several different pages, creating recipes, keywords, comments, etc.) ? 

Since this is optional for a user to implement its ok if things dont work but i feel like there might be a little to many things that break with this. Since you plan on using this anyway maybe you can just run your setup for a few days/weeks and then give feedback on how well it works and maybe we can add that to the documentation as well ?

cazier:
I've been using this set up for about two months, doing most of all the things you've mentioned, except for adding comments, but had been pushing [my own build](https://github.com/bcazier-dockerhub/recipes) to Dockerhub that just replaced the two files that needed changes.

Django is really quite convenient in that it allows you to just slot in the user via the proxy header rather than via a login page. Technically, it doesn't require the supplied user to exist, but since I'm limiting the number of people that have access to this domain, I just add those (via the Django Admin page) ahead of time.

I'm happy to keep testing it out this way for a while, but I would like to vouch for the fact that it has worked over the last few months without any issues!

vabene1111:
No if you have already been using it for this long we can definitely integrate it. 

Just out of curiosity because i find this fascinating: When you create a recipe with your setup how is the `created_by` associated ? Because it is a required field of the recipe model to have an associated user ? Does django choose from one of the user models you manually created ? 

Sry for all the questions, if find this whole setup very technically interesting and also if i understand it better i might improve some features to work better on "userless" setups (e.g. showing a proper name/label for users without objects).

cazier:
Not a problem; I totally understand the questions!

This recipe was made with a user logged in via the proxy:
image

The proxy supplied the username `brendan` which _does_ correspond to an existing user in the recipes database. (I initially started using the built in authentication, but when I got to having a number of services running on my server, I tried to switch as many as possible over to an SSO option.) Looking at the user database, you can see this user exists, and has the table filled out:

image

For the sake of being thorough, though, when logging in as a user that exists in the SSO, but not recipes, Django automatically creates that user in the users database, but does not assign that user a group. (Sorry for all the inline screenshots...)

image

This `system` user is "logged in" and can see the `/search/` page, but as they're not members of any group, they can't actually see or edit any recipes, or any of the keywords, books, units, etc., and the following error pops-up:

image

Per the Django documentation, any user logged in via the proxy header inherits all the same permissions details as a regular user, but as they are not automatically assigned a group (even the `guest` group), they don't end up being able to access anything.

For me, personally, that works. As of now I'm creating a user in my authentication database, and specifically assigning them access to the services I want them to access. Therefore if they are supposed to have access to the recipes, I specifically create a user for them. Plus, it appears to me that, as it inherits the properties of the `ModelBackend` user, it looks like the existing error messages apply.

cazier:
I did, however, push another commit because I apparently deleted that `from os import getenv` line which is necessary to pull the header name from the .env file. Oops!

vabene1111:
ok, thank you very much for all the detailed information on how this feature works. I will merge this with the next release. I will probably change it so that there is a environment variable which adds the auth setting but other than that i think its fine.

PR #154, opened 2020-08-19
Instructions for manual installation

Mwoua:


docs/manual_install/Readme.md
```diff
@@ -0,0 +1,147 @@
...
+
+These intructions are inspired from a standard django/gunicorn/postgresql instructions ([for example](https://www.digitalocean.com/community/tutorials/how-to-set-up-django-with-postgres-nginx-and-gunicorn-on-ubuntu-16-04))
+
+**Important note:** Be sure to use pyton3.8 and pip related to python 3.8. Depending on your distribution calling `python` or `pip` will use python2 instead of pyton 3.8.
+
+## Prerequisites
+
+*Optional*: create a virtual env and activate it
+
+Download the latest release from <https://github.com/vabene1111/recipes/releases>
```
vabene1111, line 11: is there a specific reason you recommend downloading the release instead of cloning the repo using git (which makes further updates easier). 

If there is anything i would need to change in the repo structure i can do that. Or if its just some `.gitignore` entries that would not be a problem either

vabene1111:
Very good instructions, thanks for providing them! Will be merged once we discussed the one comment i made.

docs/manual_install/Readme.md
Mwoua: I didnt know how stable was the branch. And when I started I hope that the version displayed in the settings would not be 0.0.0. But no real reason.

If the branch is stable enough, i can change this line to a `git clone https://github.com/vabene1111/recipes.git`

Issue #143, opened 2020-07-29
Improve install process


vabene1111:
improve install instructions so its not so easy to forget stuff like .env or nginx.d 

ideas by @tdvantine in #140  https://pastebin.com/VDmXK7Dt

NessInOnett:
I agree, some improvements could be made to the installation process.  I haven't tested this project out yet, just exploring what self-hostable software is available for recipe management.  Installation had me hung up for a bit

vabene1111:
can you explain what exactly caused you difficulties ? the problem with writing a software is that you always know how everything works and thus its hard to see where others might have difficulties ?

Also please tell me if you had experience with docker before.

NessInOnett:
Nothing major.. just maybe add more complete steps. 

I'm familiar with docker.. less so with docker compose.  I use it for a couple of services, but most of my stuff uses regular docker.  I just think the documentation could benefit from more of a step-by-step process.   

> Choose one of the included configurations here.

> **<choose one, and then what?>**

> Download the environment (config) file template . . . . 

I plan to install the software tomorrow, and I expect what I'll end up doing is cloning the entire repo, cd'ing into the configuration directory of my choice (plain), and proceeding from there.  I don't know for certain if that's the correct way to do it, but I'm sure it will work.   That's where I think extra instructions could be helpful.

I checked dockerhub to see if I could use that instead of docker-compose.  The package is available there, but no documentation at all and no mention of which configuration flavor it is (plain, nginx-proxy, etc) .. so I came back here to re-read the docker-compose instructions.

tdvantine:
If you look at the pastebin above you should get a general idea of how the process works. Those instructions are based off a different style of deployment (zip files) but if you have a GUI easy peasy, CLI is a little more "fun" but not hard.

vabene1111:
Ok, thanks for the feedback. I personally alsways use docker compose as its (for me at least) just a clean and simple way to store the configuration of my containers. Generally a docker compose config can be converted into a docker command and the other way around if you know what each means. 

If you want to do a plain docker installation and setup the database and proxy container yourself than that should work without any problems. If you want you can even submit the exact commands so i can add them to the instructions.

stewartadam:
> If you want to do a plain docker installation and setup the database and proxy container yourself than that should work without any problems. If you want you can even submit the exact commands so i can add them to the instructions.

[...]

PR #172, opened 2020-08-30
Permit MEDIA_URL and STATIC_URL to be set from environment variables

stewartadam:
Addresses some of the issues brought up in #143.


vabene1111:
this looks good but in order for people to notice this can you add the two settings to the `.env` file and write a short comment on what they do (maybe with a link to the django doc) ?

the value itself can be left commented out but just so people know how it works

PR #199, opened 2020-10-16
Add nutritional values

sebimarkgraf:
Current WIP for https://github.com/vabene1111/recipes/issues/179
Approach which allows users to define types of nutritional values.

This makes implementing an automatic import even harder but allows for custom nutrition models and tracking of micronutritiens.

Any feedback is welcome!


vabene1111:
This looks very interesting 👍  i am no nutrition expert so no idea what micronutritiens are but i guess some people will like want to use this.

What do you think about the general idea of adding the nutritional values to the food items instead of the recipe so that, once added, nutritions can be calculated automatically ? Or maybe both is needed recipes and food ? 

Would also need some kind of unit conversion i guess 🤔 just and idea anyway, looks good for now !

sebimarkgraf:
Maybe, you are right about the calculation from the ingredients. 
Could be harder to implement right away, but is probably more versatile and easier to use then adding the nutritions to every recipe.

We could add a page similar to the current ingredient conversion, that allows to give the nutritions for every ingredient.
That could be then automated with imports for different ingredients.


But for the moment I would stick with per recipe nutritions. They are usually provided and the community can give feedback if the feature is useful. If determined useful we could add another option to calculate the nutritions instead.

Either way we would use the same nutrition view and therefore I would try to implement something nice for that and add it to the PR.

vabene1111:
definitly implement the per recipe thing as its needed anyway (e.g. if you dont want to include cooking oil or whatever in the ingredients but it adds to the nutritions) 👍

sebimarkgraf:
I've updated my approach to a fixed number of nutritions. In my opinion this is easier to maintain and build upon in the future.

I've tried adding nutritions with my old approach and it was labor intensive while not providing to  much benefits.
It's pretty easy to add more nutrition information on request and it is easier to query external databases / calculate if we know the nutritions that can be added.

I am going to update the edit recipe to allow for nutritions and then try to get this ready for merging.

sebimarkgraf:
@vabene1111 I guess this should be the functionality for a first merge. 

I went with Calories, Carbohydrates, Fats and Protein. Other nutritions could be added on request.

The editing works and the visual is alright, but if I remember correctly you wanted to revisit the show recipe page anyway at some point, right?

How should I test the changes?

cookbook/models.py
```diff
@@ -194,6 +205,7 @@ class Recipe(models.Model):
     working_time = models.IntegerField(default=0)
     waiting_time = models.IntegerField(default=0)
     internal = models.BooleanField(default=False)
+    nutrition = models.ForeignKey(NutritionInformation, blank=True, null=True, on_delete=models.PROTECT)
```
vabene1111, line 208: why is this PROTECT ? shouldn't it be CASACE (if the recipe is delete, delete the nutritional info as well ?) maybe i am mixing something up, dont have time to test it atm

cookbook/views/api.py
```diff
@@ -282,6 +282,10 @@ def get_queryset(self):
         queryset = ViewLog.objects.filter(created_by=self.request.user).all()[:5]
         return queryset
 
+class NutritionInformationViewSet(viewsets.ModelViewSet):
```
vabene1111, line 285: if there is no call to this api anyway as the information is always returned with the recipe there is no need to add the ViewSet. 

If we want to have a seperate API endpoint for just nutritional information we should add the appropriate permission classes and filters

vabene1111:
first of all thank you very much. 

I would call myself a beginner at testing so i just do whatever feels to be the correct way from all the information i have at hand. 

I usually make sure that all API calls for  a given route work, in your case that would be the old tests for edit recipe plus a few more for creating/updating/deleting nutritional values. 

To be honest the most important thing at the moment with the tests is making sure that permissions are applied correctly, but having more tests is always good! 

If you dont feel comfortable writing the tests i can do that to.

i have added two comments to the code, otherwise everything looks fine from my first impression (dont have much time atm but will take a good look later)

cookbook/models.py
sebimarkgraf: Ah, sure! Mix it up somehow

cookbook/views/api.py
sebimarkgraf: I removed it

sebimarkgraf:
I have included your comments.
Not sure about the testing, never did testing of Django applications but I can try to add tests to the recipe editing.

vabene1111:
i wont have time to merge this for at least another two weeks so i will leave this open, if you want to add testing feel free if not i will do it when merging (or tell myself that i will do it later 😂 ) honestly i haven't done testing in django or in fact any other language before this project and it has really surprised me how well it works/how much you can improve when properly thinking about testing (but maybe you know this already so ...).

kushfest:
It would be awesome to add sodium to the list. That's a really common value to track for heart health/blood pressure. While we're at it, this list would cover the most common nutritional values:

    Serving Size
    Sugar
    Carbohydrate
    Cholesterol
    Fiber
    Protein
    Sodium
    Saturated Fat
    Unsaturated Fat
    Total Fat

vabene1111:
can you open a seperate issue where you list the most common nutirtional values .. maybe i will just make it user configurable 🤔

kushfest:
Gladly! user configurable would be a great solution. Most people look for the same nutritional values, but every situation is different. Like people with high blood pressure will pay more attention to sodium content and added sugars, while calories may factor in more for people looking at weight management, etc.

PR #211, opened 2020-10-29
Username field consistency

Tmaxxrox97:
Changed Username field label from "Name" to "Username" on new user registration page as mentioned in https://github.com/vabene1111/recipes/issues/209


Tmaxxrox97:
This is my first time creating a PR so let me know if I didn't do something correctly.

vabene1111:
thanks for fixing this, makes more sense this way! Also you did everything correctly with the PR, very good!

PR #213, opened 2020-11-03
Show random recipes in meal planner

tourn:
Hi again!

Since nobody implemented random recipes (#5) yet, I've put together my own. It doesn't do any fancy tag filtering yet, but shows 5 random recipes on loading the meal plan and adds a button to load another 5.

Feel free to reject this if you don't like the way it's done, I just wanted to share what I'm running on my fork.

Cheers!

image


vabene1111:
this looks very cool ! 

i will take a quick look at the new dependency when i have time and then this will be released with the next update.

PR #227, opened 2020-11-12
Fix typo

Scrimples:


vabene1111:
Thank you for providing this fix. As the english locale is not really translated but used as the default string in all the templates this should be fixed directly in the HTML. I can do this so no worries, i will leave this PR open until i have fixed it.

vabene1111:
fixed directly in source code

Issue #275, opened 2020-12-20
Recipe Update Error


vabene1111:
500 error when editing some recipes
`There was an error updating the recipe!{"steps":[{"ingredients":[{},{},{},{},{},{"unit":{"name":["This field may not be blank."]}},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{}]}]}`

Probably related to import not populating some fields correctly

_Originally posted by @edcous in https://github.com/vabene1111/recipes/discussions/274_

vabene1111:
the issue is related to units or ingredients being objects with empty text when the importer does not populate these fields appropriatly

vabene1111:
actually inserts a unit with an empty name into the db. What needs to be done:

- [ ] fix the importer so that null is used instead of this "dead" unit
- [ ] write migration to null out all ingredients that use these dead units (and possibly foods)
- [ ] set database constraint to disallow empty names
- [ ] test if everything still works and no old recipes are broken

PR #277, opened 2020-12-21
Improve text to ingredient parsing

l0c4lh057:
The previous implementation of parsing ingredients was very simple. I now wrote a parser
that I would consider good. It takes care of several edge cases and notations.

- Supports fraction unicode (½, ¼, ⅜, ...)
- Supports notations like `1½` and `1 1/2`
- Supports unit directly after the amount without space inbetween (`2g`, `2½g`)
- Supports notes (`5g onion (cubed)` -> amount: 5, unit: g, ingredient: onion, note: cubed)
- Supports notes (`5g onion, cubed` -> amount: 5, unit: g, ingredient: onion, note: cubed)
- Does not break when both commas and brackets exist

### DIFF
--- a/cookbook/helper/ingredient_parser.py
+++ b/cookbook/helper/ingredient_parser.py
@@ -2,130 +2,130 @@ import unicodedata
 import string
 
 def parse_fraction(x):
-	if len(x) == 1 and "fraction" in unicodedata.decomposition(x):
-		frac_split = unicodedata.decomposition(x[-1:]).split()
-		return float((frac_split[1]).replace("003", "")) / float((frac_split[3]).replace("003", ""))
-	else:
-		frac_split = x.split("/")
-		if not len(frac_split) == 2:
-			raise ValueError
-		try:
-			return int(frac_split[0]) / int(frac_split[1])
-		except ZeroDivisionError:
-			raise ValueError
+    if len(x) == 1 and "fraction" in unicodedata.decomposition(x):
+        frac_split = unicodedata.decomposition(x[-1:]).split()
+        return float((frac_split[1]).replace("003", "")) / float((frac_split[3]).replace("003", ""))
+    else:
+        frac_split = x.split("/")
+        if not len(frac_split) == 2:
+            raise ValueError
+        try:
+            return int(frac_split[0]) / int(frac_split[1])
+        except ZeroDivisionError:
+            raise ValueError
 
 def parse_amount(x):
-	amount = 0
-	unit = ""
-	
-	did_check_frac = False
-	end = 0
-	while end < len(x) and (x[end] in string.digits or ((x[end] == "." or x[end] == ",") and end + 1 < len(x) and x[end+1] in string.digits)):
-		end += 1
-	if end > 0:
-		amount = float(x[:end].replace(",", "."))
-	else:
-		amount = parse_fraction(x[0])
-		end += 1
-		did_check_frac = True
-	if end < len(x):
-		if did_check_frac:
-			unit = x[end:]
-		else:
-			try:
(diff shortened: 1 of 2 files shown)


l0c4lh057:
I also wrote a test for the parse function how I'd expect it to work. It includes some edge cases: https://anonfiles.com/recepc11p3/testIngredientParser_py

Feel free to look over it and say if you would expect any of them to give a different result. I also included cases in which it gets detected wrong that I think is impossible to fix, so even in the expectations I wrote there are examples that you would not want like that in real life but that I don't think can be improved.

There is just one case in which I did not know what to do and that is when only a vague amount is given like `2-3`.

l0c4lh057:
I'd recommend testing if I did not break anything else, but I tried importing recipes from 3 or 4 different websites and they all seemed to still work as expected. I did not notice any difference (besides notes getting detected too of course)

What I tested importing (and also used to base lots of edge cases on):
- https://www.ica.se/recept/kanelbullar-5342/
- https://www.chefkoch.de/rezepte/1305881235547779/Kuerbisstrudel.html
- https://www.jamieoliver.com/recipes/chocolate-recipes/scrumptious-yule-log/
- some other recipe on jamieoliver.com that I can't find anymore

l0c4lh057:
i just tested what it gives with something like `2-3` as amount and it seems that it just takes the first number
grafik

l0c4lh057:
has issue with multiple ingredients being in one entry separated by comma, but that is something not used a lot and usually only for stuff like salt and pepper
grafik

vabene1111:
This looks really really good, thank you very much! For some reason GitHub did not run the normal test suite for this PR (only the CodQL checks) but there should be some tests verifying if the importer works. Have you run those tests ? 

I suspect, since the tests are pretty basic/bad that they wont run successfully as you changed the structure of where stuff goes (the note recognizing is awesome by the way)

i guess i will have to re-write the import tests to be more useful than simply checking if the length of the result is correct but you part looks really good.

If you want you can add the test suite you already wrote to the automatic tests included in this repository, otherwise i will do this. 

Will probably take a while for me to get this merged as i would do it together with the rewrite of the normal tests but i am looking forward to getting this included!

l0c4lh057:
I have no experience at all with automated tests but I'll give my best to figure out how to do that

vabene1111:
you dont need to, i can do it when merging :) but if you want feel free to give it a try. I heard for years how awesome test driven development and all that fancy stuff is and nerver actually used it. 

With this project i did my first tiny steps towards some decent test coverage and i must admit that i really liked it, it actually helped my find a few edge cases that i would otherwise not have found.

The python file you linked above is already pretty much all you need. Basically you open the test directoy, find the file where the other parser related tests are and add another function to it that tests your new parts. Its basically just python code where you add a few assert statements which then get checked.

If you need any help feel free to ask and if you don't feel like writing the tests dont worry, i can do so when i have some more time again.

l0c4lh057:
I guess I'll just leave that up to you.
One thing I just noted: I am using empty strings as default values. Idk if that is true but my feeling is that you used None as default value. If that is an issue just let me know, I think I could easily change that.

vabene1111:
i am honestly not sure any more. I have changed that too many times to support empty fields on some things that i previously considered required that i will have to look into this. 

There is also an issue related to this i guess #275 

i will have to take a deep look into this and then i will fix it in both cases

vabene1111:
if you are interested, i added your test in https://github.com/vabene1111/recipes/commit/840f5ec60d00c5ec9470395122312bb5cb231a89 and it was super easy as you already did basically everything that needed to be done.

The other issue will be looked at independently

PR #279, opened 2020-12-22
Minor improvements

### LINKED ISSUE #278: Unintuitive keywords label in website import
### Version
Version: latest master clone (2020-12-20, just shows 0.0.0 in dev setup)

### Bug description
When importing a website the checkbox to import all keywords and not just those that already exist is a bit weird. The text above `All Keywords` is the label to the checkbox that is one line below whereas the text right next to the checkbox is no label. That was a bit irritating to me because usually you can just click the text next to a checkbox to check it which did not work here but when looking at the source code there was a label. I did not even consider the title as label until I saw it. Imo the text right to the checkbox should be the label. The title can be a label too but doesn't have to be (idk if there are any accessibility guidelines saying something about multiple labels for one checkbox)

grafik

l0c4lh057:
- Add min (`1`) and max (`999`) to the servings spinner on the recipe view
- Made the text next to the checkbox the label instead of the title for the `All Keywords` check on website import (fixes #278)
- Add the `show` class to the advanced search element if you searched with the advanced search

### DIFF
--- a/cookbook/templates/index.html
+++ b/cookbook/templates/index.html
@@ -62,7 +62,7 @@
                     </div>
 
                     <div class="row">
-                        <div class="collapse col-md-12" id="collapse_adv_search">
+                        <div class="collapse col-md-12{% if filter.data.keywords or filter.data.foods or filter.data.internal and not filter.data.internal == "unknown" %} show{% endif %}" id="collapse_adv_search">
                             <div style="margin-top: 1vh">
                                 {{ filter.form.keywords | as_crispy_field }}
                             </div>
--- a/cookbook/templates/recipe_view.html
+++ b/cookbook/templates/recipe_view.html
@@ -109,7 +109,7 @@
                         <div class="col col-md-3">
 
                             <div class="input-group d-print-none">
-                                <input type="number" value="1" maxlength="3" class="form-control"
+                                <input type="number" value="1" min="1" max="999" class="form-control"
                                        v-model="ingredient_factor"/>
                                 <div class="input-group-append">
                                     <span class="input-group-text"><i class="fas fa-calculator"></i></span>
--- a/cookbook/templates/url_import.html
+++ b/cookbook/templates/url_import.html
@@ -194,9 +194,9 @@
                 </div>
 
                 <div class="form-group">
-                    <label for="id_all_keywords">{% trans 'All Keywords' %}</label><br/>
+                    {% trans 'All Keywords' %}<br/>
                     <input id="id_all_keywords" type="checkbox"
-                           v-model="all_keywords"> {% trans 'Import all Keywords not only the ones already existing.' %}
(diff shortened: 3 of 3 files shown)


l0c4lh057:
Pretty sure there has to be an easier way to only include some commits in a pull request. After my first pull request I created a new branch to be able to make a second PR because they are unrelated and PRs are based on the branch. I then made these changes, pushed them and then in the overview when creating a new PR I noticed that the commits from my other PR are also included. Since the new commits did not change anything that was changed in my old commits I think it should be possible to just remove them but I really don't know. I then created a new branch that I based on the last commit that is not from me, remade and recommited the changes and then create a PR but I'm sure there has to be a better way. If you know anything please let me know. (Besides not forgetting about that when creating the new branch in the first place)

vabene1111:
haha i had the same Problem with PRs in the past many time .. in the end i always went for the option you described and i am not really sure if there is a better way. This makes it especially complicated if a second PR is using something from a first one.

Regarding the changes:
Sounds really good, the only thing i would change is allow the scaling calculator to also go to 0.XX since sometimes it might be desired to cook half of a recipe (this is probably only relevant as long as it is the way it currently is where the scaling factor is not based on servings but on a fixed "recipe" amount, adding servings was discussed at length in the other open PR but i haven't gotten around to implementing it).

i hope that i will find some time over the christmas holiday to merge your PRs but i cant guarantee it

l0c4lh057:
Yeah you are right, I did not think about that. It should also be possible when the servings attribute is implemented, e.g. for something like cake or tarte most often 1 serving is used to indicate that you will get one of them with the ingredients and not how many serving that actually are. But maybe you only want to bake a small version of the cake.

Is there anything that allows you to specify custom step sizes? I'd like to make it that if the value is 1 and you go down it will decrease by 0.1 but it would increase by 1. Just changing the step size based on the value would not work, it also depends on the direction of the change. I tried overwriting the element's stepUp and stepDown function but apparently they are not used when clicking the arrows or using the arrow up/down keys. I don't want to create a custom component for this and I don't want to use a value other than 1 for increasing if the value already is bigger than 1.

l0c4lh057:
I'll just leave it as is for now because the previous version did not allow non-integer values either. You can input them manually and it will scale correctly but it will tell you that it is an invalid value. The same also applies to this input with min and max (at least for me on firefox, who knows if other browser decide to behave differently).

vabene1111:
i would say it is perfectly reasonable to manually type in the value if you dont want a full integer 👍

l0c4lh057:
That's fair. Imo it shouldn't say that the value is invalid tho. But I can't find anything that would allow this. If you set the step property then only `min + whole multiples of step` and `max` are valid values. But apparently you can't set something like a step used for the validator and a step used by the browser to decide how much to increase/decrease so that steps inbetween are allowed too.

PR #297, opened 2020-12-30
Assign authenticated user to created_by

lipschultz:
Fixes https://github.com/vabene1111/recipes/issues/296


vabene1111:
thanks for fixing this!

PR #299, opened 2020-12-31
Add latvian language

melkypie:


vabene1111:
thanks for providing the translations. Sry that i did not accept you on transifex earlier. 

I have added what you did to transifex, if you want to translate stuff in the future just send me a join request on transifex (i sadly cannot simply add you) and i will add you to the team :)

PR #316, opened 2021-01-06
Improve documentation formatting

KyleWJohnston:
This mostly fixes documentation formatting, but there are a few minor changes to the new Docker commands and templates.

### DIFF
--- a/docs/install/docker.md
+++ b/docs/install/docker.md
@@ -2,7 +2,7 @@
     Setting up this application using Docker is recommended. This does not mean that other options are bad, just that
     support is much easier for this setup.
 
-It is possible to install this application using many docker configurations.
+It is possible to install this application using many Docker configurations.
 
 Please read the instructions/notes on each example carefully and decide if this is the way for you.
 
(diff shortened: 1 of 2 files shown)


docs/install/docker.md
```diff
@@ -60,6 +61,8 @@ services:
   web_recipes:
     image: vabene1111/recipes
     restart: always
+    ports:
+      - 80:8080
```
vabene1111, line 65: this is not correct. If the web_recipes container already exposts port 80 nginx can no longer do so

vabene1111 (changes requested):
i have added a small comment, if you could quickly remove these two lines i can merge this and the documentation will be in good shape again

vabene1111:
I will also review the other branches in the evening to make sure they dont suffer from similar issues.

docs/install/docker.md
KyleWJohnston: This makes sense, thank you. I did some further testing and think I've figured it (Docker basics) out.

For example, to expose on port 3000, you could use `docker run -d -p 3000:8080 ...` or, for Docker Compose,
```
  nginx_recipes:
    ports:
    - 3000:80
```

It worked in my tests.

vabene1111 (approved):

vabene1111:
Thank you 👍 i will have a look at your other mentioned issue in the PR asap

PR #321, opened 2021-01-06
Sticky Navbar

neferin12:
I wanted to propose a change for the navbar, so that it sticks to the top of the window like this when you're scrolling down: 

recipes jp-studios de_search_ (2)

I find that useful, as I don't have to scroll up and in my opinion it looks better to have it there all the time, but since its styling that might be just my subjective opinion. What do you think about it?


vabene1111:
I like the idea. 

What do you think about making this user customizable ? It would proably be really easy to add a setting for that.

I will leave this PR open and when i merge it add a setting so that everyone can decide how he likes it.

neferin12:
Sounds like a good idea. I added this as an option to the settings, if you want me to i can also push it

vabene1111:
if you already did it feel free to push it, i haven't gotten around to doing so :)

neferin12:
Ok, I pushed it, but a little disclaimer: I never worked with django beyond of some simple playing around with models, and adding this setting was a bit learning by doing, so I hope I did nothing wrong or unconventional

vabene1111:
i have looked over everything and it looks perfect ! 

i will test it and then it will be in the next release.

PR #322, opened 2021-01-07
Fix ingredient parser to allow Plural-suffixes

jakobwenzel:
The new Ingredient Parser failed to parse ingredients like "400 g Karotte(n)", as it will try to match closing brackets with opening brackets only at the start of tokens.

This PR tries to fix that.

Also, the testcases always succeeded, as the wrong function was called and the result was checked for inequality instead of equality.


vabene1111:
hmm i tought i fixed the test, i noticed some problems there too 🤔 
Thanks for fixing both issues !

PR #326, opened 2021-01-09
Unique constraint on RecipeBookEntry

bloomcake:
Added a unique constraint on `RecipeBookEntry` table to avoid duplicate entries.


vabene1111:
Thank you!

PR #328, opened 2021-01-10
Small cleanup

bloomcake:
the small "cleanup" probably turned into a somewhat larger one, sorry 

actually I didn't plan to do this, but somehow I couldn't stop myself. so I ran through the project once with PEP 8.

maybe a style guide with suitable configurations for the project can be developed additionally. looking forward to your feedback!


vabene1111:
i did not really have a problem with the existing format an tought PyCharm automatically applied PEP 8 when reformatting but this looks good! Thank you very much

vabene1111:
i am going to merge this so it does not get outdated but maybe we should open a seperate issue where we discuss on how to best integrate style checking into the normal DEV wokflow. 

I usually just use the standard IDE auto-format and as i do not have any strong feelings about formatting dont really care about the details that much, so it needs to be something simple.

vabene1111:
Ok so i have just tried to review the code and it appears that there are some code changes and also few formatting changes that i really dont like visually like the formatting of the urls.py file.

Besides that i cannot really, with good conscience, merge a PR with thousands of changes without reviewing every single one as i would be responsible for pushing potentially malicious code to an unknown number of people running this application.

I suspect that you have used some kind of automatic tool to convert all those files to the new Format but that you have also made some changes manually is that correct ?

Do you think its possible to just tell me the tool and the important manual changes or would that be a lot of work ?

bloomcake:
I have made the changes myself except for the imports. with `flake8` I have analyzed the code and then adapted. the imports (especially the order) I have changed with `isort`

If you want I can split the merge request into three:
- all imports changes from isort
- all changes with flak8 (from pep8)
- all manual changes/adjustments from me

enrico-kaack:
A possible automated approach could be a git pre-commit hook to run the linting https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks

vabene1111:
If you have done everything manually than it is to much work to just throw away 🤔 
I will take the time to review all changes and then merge them, thanks for all that work 👍

vabene1111 (approved):

bloomcake:
Thanks! I will split it in the future

Issue #358, opened 2021-01-27
Develop tag in docker hub


cesarblancg:
Hi! I saw your project on the docker hub. For the develop tag, the target is "linux/amd64", and the default tag has others like "linux/arm/v7". It's possible to add arm v7 to the develop tag? I'm using a rpi and I wanna try the latest version. 
It is possible to do?

vabene1111:
i will close this  as our discussion is moved to the PR

PR #359, opened 2021-01-27
Update docker-publish-dev.yml

cesarblancg:
Added more target platforms to develop tags on docker hub.

#358

### DIFF
--- a/.github/workflows/docker-publish-dev.yml
+++ b/.github/workflows/docker-publish-dev.yml
@@ -18,9 +18,11 @@ jobs:
             VERSION_NUMBER = 'develop'
             BUILD_REF = '${{ github.sha }}'
           write-mode: overwrite
-      - name: Publish to Registry
-        uses: elgohr/Publish-Docker-Github-Action@2.13
+      - name: Build and publish image
+        uses: ilteoood/docker_buildx@master
         with:
-          name: vabene1111/recipes
-          username: ${{ secrets.DOCKER_USERNAME }}
-          password: ${{ secrets.DOCKER_PASSWORD }}
+          publish: true
+          imageName: vabene1111/recipes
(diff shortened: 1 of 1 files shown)


vabene1111:
This is a nice idea but to be honest i often push breaking changes to develop so its not really recommended to use that branch. 

Apart from that i also use develop to deploy to my server for testing and buildx for all plattforms takes something like an hour or so to build so it would make testing even harder for me than it already is.

What do you think of a beta branch ? basically develop but i only merge it when its not completely broken 😂

cesarblancg:
I like the idea of a beta branch.

vabene1111:
ok then i will implement that 👍

vabene1111:
ok so i added a beta branch and used the code you provided in this PR, in about an hour the release should be out. I would really love some feedback so feel free to post ideas, wishes, feedback, etc. from what you see on the beta branch so i can improve on it. 

I currently dont have a good way of making beta changelogs as i dont really feel like always making  a new release but most things should be obvious. 

Also: Even tough it is a "offical" beta release i cannot guarantee that things will be working all the time. There might be broken versions and backpatching is currently NOT supported or at least at your own hand.

vabene1111:
Ok something is broken in the buildx action, either its something i broke with the dockerfile or the action is broken 

https://github.com/vabene1111/recipes/runs/1783285924

need to figure this out first

vabene1111:
ok so for whatever reason the  first run failed but the second one worked. Beta image is now released on docker hub.

https://hub.docker.com/layers/vabene1111/recipes/beta/images/sha256-99566832ed0fa5e212d56609183a8627e747bddc79abd0c0f622df1d75aad78f?context=explore

if i start forgetting about it feel free to let me know :)

PR #364, opened 2021-01-29
Update boot.sh

cschoeni:


docker-compose.yaml
```diff
@@ -0,0 +1,40 @@
...
+    restart: always
+    env_file:
+      - ./.env
+    volumes:
+      - staticfiles:/opt/recipes/staticfiles
+      - nginx_config:/opt/recipes/nginx/conf.d
+      - ./mediafiles:/opt/recipes/mediafiles
+    depends_on:
+      - db_recipes
+    command: ["./wait-for-it.sh", "db:5432", "--", "python", "app.py"]
```
cesarblancg, line 22: I think you should change "db:5432" by the name of the database container and in the .sh file get the port using the variable set on the .env file.

vabene1111:
this looks really interesting but might i ask why do you need all the dependency changes in the Dockerfile ?

vabene1111:
closing due to inactivity, db wait has been added by a more recent PR and dependencies have changed since

Issue #385, opened 2021-02-03
Icon web app on Apple devices


cesarblancg:
### Version
Please provide your current version (can be found on the system page since v0.8.4)
Version:  beta

### Bug description
The Icon web app on iPad pro doesn't appear. 

I searched info and founded this:
_To specify multiple icons for different device resolutions—for example, support both iPhone and iPad devices—add a sizes attribute to each link element as follows_
font:
https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html

vabene1111:
thanks for closing, i thought github did it automatically as it was mentioned in the merged PR (it used to do that anyway 🤔 )

PR #386, opened 2021-02-03
Solving Apple web app issue

cesarblancg:
#385


vabene1111:
i hate apple, they always make stupid stuff like this ... but thanks for fixing this 👍

PR #398, opened 2021-02-14
Fix ingredient parsing for fractions using '/'

l0c4lh057:
Even though ingredients like '1 1/2 something' already worked fine and got converted to 1.5 something
I just came across a recipe using '1/2' as the whole amount without any whole number before that.
Apparently I overlooked that case before so I now also fixed that.


vabene1111:
awesome, thank you very much. The ingredient parser is extremly helpful and has allowed many cool features 👍

PR #401, opened 2021-02-15
Update .env.template

tdvantine:
Matched the db user to that of the install instructions, fixed some grammar/spelling errors and simplified wording...hopefully.


vabene1111:
thank you, thats definitly clearer 👍

PR #425, opened 2021-02-19
Add support for DATABASE_URL env var

tourn:
I'm running my own instance of Tandoor using [Dokku](https://dokku.com/), which injects the database login in the DATABASE_URL env var. This is a standard used by several PaaS (Heroku does it too, probably others as well)

This PR allows configuring the database via the DATABASE_URL instead of the whole POSTGRES_HOST, POSTGRES_PORT etc.

If you're not OK with pulling in a new dependency for this I could probably write it by hand using a regex.

Anyway, this is the last bit of code preventing me from running the official version instead of my own branch, so i'd like seeing this included in some form. I might also throw in an installation guide for running Tandoor on Dokku if people are interested.

Cheers!


vabene1111:
Thank you for providing this. While i like the simplicity and think this feature is important i do think that not having an extra dependency for this would be preferable.

I have not really looked into how this works, do you think it would be a lot of effort to do this without a dependency ?

tourn:
The DATABASE_URL is basically just a string in the format of `engine://username:password@host:port/dbname`. Shouldn't be too hard to pull apart using a regex. I'll look into it.

vabene1111:
perfect, thank you!

tourn:
So I've written the regex and it should be working, however my production installation seems to fall back to the default sqlite database anyway. This didn't happen using the dj-database-url lib. I'm having trouble getting my dev environment running since some requirements apparently need to compile some rust code, which makes this a pain to test for me right now.

If you find at a glance what's wrong that'd be great, otherwise I'll come back to this later when i've got my dev setup in order.

vabene1111:
i dont currently have time but just leave this open and i will take a look at it as soon as i get some time for it.

vabene1111:
i have looked at your code and everything looks perfectly fine. If you add a database string and a engine this should work.

i have not pulled the code but run parts of the script and they seem to work. Have you verified the environment variables are set correctly (sry dump question, i sure you have but just to make sure).

Do we also need to provide a way for the user to configure the name of the variable that should be used for the connection string or is this usually configurable with PaaS providers ?

tourn:
Okay, I finally managed to resolve my build issues and get back to this. Turned out to be a stupid little mistake that is now fixed. This PR is ready to go now.

> Do we also need to provide a way for the user to configure the name of the variable that should be used for the connection string or is this usually configurable with PaaS providers ?

Not to my knowledge, no. This could be easily added later if it's needed, though.

Issue #443, opened 2021-02-21
Sub-Path not included in JavaScript files


albcp:
### Version
Version: 0.14.3

### Bug description
The URLs from *reverse.js* do not contains the Sub-Path. Because of this, when JavaScript sends a request or include a URL into a template, the endpoint does not contains the Sub-Path.

This can be tested on the *recipe_view* page of any recipe. This page does not include the Sub-Path in the hrefs from the 3 dot menu and sends two requests to *api/*.

vabene1111:
hmm interesting. do you mean sub domain by sub path ? e.g. subdomain.mydomain.com ?

albcp:
No, I mean after the domain, e.g. mydomain.com/recipes/

As you can see in the screenshot, api requests are sent to mydomain.com/api/ instead of mydomain.com/recipes/api/.
![2021-02-22_18-16](https://user-images.githubusercontent.com/3170731/108744971-dbbbc100-753a-11eb-81a7-62fc0070e5be.png)

It was working in version 13.0 with the environment variables STATIC_URL, MEDIA_URL and SCRIPT_NAME.

vabene1111:
ah ok .. i was using a different system then to get the urls on the client. 

I dont really know how the url generation system works when the system is run under a subpath. I will see if and how i can fix this.

PR #450, opened 2021-02-22
Add environment variable to include Sub-Path in resolve.js.

albcp:
#443

### DIFF
--- a/recipes/settings.py
+++ b/recipes/settings.py
@@ -228,7 +228,7 @@ LANGUAGES = [
 # path for django_js_reverse to generate the javascript file containing all urls. Only done because the default command (collectstatic_js_reverse) fails to update the manifest
 JS_REVERSE_OUTPUT_PATH = os.path.join(BASE_DIR, "cookbook/static/django_js_reverse")
 
-JS_REVERSE_SCRIPT_PREFIX = os.getenv('JS_REVERSE_SCRIPT_PREFIX', '')
+JS_REVERSE_SCRIPT_PREFIX = os.getenv('JS_REVERSE_SCRIPT_PREFIX', os.getenv('SCRIPT_NAME', ''))
 
 STATIC_URL = os.getenv('STATIC_URL', '/static/')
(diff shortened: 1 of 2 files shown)


recipes/settings.py
```diff
@@ -228,6 +228,8 @@
 # path for django_js_reverse to generate the javascript file containing all urls. Only done because the default command (collectstatic_js_reverse) fails to update the manifest
 JS_REVERSE_OUTPUT_PATH = os.path.join(BASE_DIR, "cookbook/static/django_js_reverse")
 
+JS_REVERSE_SCRIPT_PREFIX = os.getenv('JS_REVERSE_SCRIPT_PREFIX', '')
```
stewartadam, line 231: Given most users installing to a sub-directory probably want this to match SCRIPT_PATH, would it make sense to turn this to `os.getenv('JS_REVERSE_SCRIPT_PREFIX', os.getenv('SCRIPT_PATH', ''))`

albcp: Did you mean SCRIPT_NAME? I have tried it as you said but with SCRIPT_NAME and it works fine.
```suggestion
JS_REVERSE_SCRIPT_PREFIX = os.getenv('JS_REVERSE_SCRIPT_PREFIX', os.getenv('SCRIPT_PATH', ''))

```

albcp: Did you mean SCRIPT_NAME? I have tried it as you said but with SCRIPT_NAME and it works fine.
```suggestion
JS_REVERSE_SCRIPT_PREFIX = os.getenv('JS_REVERSE_SCRIPT_PREFIX', os.getenv('SCRIPT_NAME', ''))
```

vabene1111:
i feel like this should be done in one sweep together with the possibility to configure a base URL form which the other options are then generated but since i dont currently have the time to implement this i guess that this solution should fix the issue for the time being. 

I will merge this but not put anything in the documentation so this remains an temporary and not really "official" solution until i have found something better (just so that everyone who does it knows that it might change in the future)

PR #464, opened 2021-02-25
Ensure Wheel is installed & Update manual install instructions

tdvantine:


tdvantine:
@vabene1111 wheel needs to be installed first before running the other install candidates, however I do not know right off the top of my head how to install wheel before building the other requirements. Feel free to take this over if it is quick and easy, otherwise I will happily figure it out over the next few days.

vabene1111:
thank you very much. I am honestly a little confused by this. Neither the docker build nor install on my local machines has ever had problems with installing the requirments. Why do you think that wheel is required ?

vabene1111:
also i have no idea on how to specifiy the install order (except by the order the packages are listed in the file), as far as i know pip resolves all dependencies automatically and installs them in the correct order

tdvantine:
Sorry didn't realize you had replied here. Well my system didnt already have wheel installed and while installing the requirements.txt it threw an error saying it couldn't complete because wheel wasn't installed. That said, it may be an edge case where wheel is typically installed, just on my system it wasn't and needed to be.

vabene1111:
interesting, i will look up how this works as i thought it would install all dependencies automatically. I will leave this PR open to make sure i dont forget it :D

tdvantine:
Just so you are up to date, I am writing the new instructions and will be testing them out over the next few days. 

As for wheel, I THINK I know what's up with that. When using a venv wheel isn't included and thus needs to be installed with reqs. That said I haven't tested that theory but will be over the next few days here.

vabene1111:
thank you very much. 

I use venvs all the time for dev and they always come with wheels or at least i never had to install it. The docker container uses a venv as well you can see it setup in the Dockerfile https://github.com/vabene1111/recipes/blob/develop/Dockerfile

But maybe venv creation is not the same on each plattfrom so this is probably a good theory to test properly

vabene1111:
ok so i am adding wheels starting with 0.15.0

I have it in the dockerfile directly not the requirements as it seemed like the requirements weren't processed in the order they are listed.

PR #474, opened 2021-03-03
Allow recipes to be imported from json directly

Patralos:
To allow import from password protected websites, I added an option to allow direct import of ld+json structure.


vabene1111:
i really like the idea 👍 its also not adding much complexity so that is good as well.

I have some other things i need to get done but will get this merged and tested for the next release i hope (no idea when that will be)

thank you

vabene1111:
manually merged due to merge conflict, thank you very much !

Issue #465, opened 2021-02-26
Add support for swedish food magazine Alltommat


roflcoopter:
**Is your feature request related to a problem? Please describe.**
In Sweden there is a popular food magazine called Alltommat (AllAboutFood) where they publish lots of recipes.

**Describe the solution you'd like**
It would be great if the importer could support the format of these recipes.
An example would be this recipe:
https://alltommat.se/recept/stekt-rodbetsgnocchi-med-brynt-valnotssmor/

Looking at the HTML it does seem to have a good structure to grab the data from.

**Describe alternatives you've considered**
None

**Additional context**
None

vabene1111:
the source does contain ld+json so it should be parsable. I will need to verify if requests are blocked or any information is stored in a non standard way but this should be possible

roflcoopter:
Thanks for looking into it! Let me know if i can do anything to help

smilerz:
@roflcoopter  The pull above will mostly fix the import - but that site does not include the units/measures in their ingredient list.

You can add a feature request at [recipe-scrapers](https://github.com/hhursev/recipe-scrapers) to have a custom scraper for that site added and when it gets merged there hopefully the new URL import will be merged here.

roflcoopter:
Great! Yeah the units are embedded in the amounts... not ideal.
Any way i can test this out for you or should i just wait for the PR to get merged?

smilerz:
If you know how to merge branches you can merge url_import_recipes branch (from [here](https://github.com/smilerz/recipes/tree/url_import_recipes)).  That won't make any database changes so should be safe.

Issue #462, opened 2021-02-25
Add Import Support for taste.com.au


BradleyDS2:
Could you please add support for taste.com.au

They are a recipe site, so the structure should be there, however when I try to import I receive:
"The requested page could not be found.

Information

Only websites containing ld+json or microdata information can currently be imported. Most big recipe pages support this. If you site cannot be imported but you think it probably has some kind of structured data feel free to post an example in the github issues."
Encountered with url https://www.taste.com.au/recipes/creamy-bacon-carbonara/12c27b1e-5fb6-48c9-ac42-82a37cc8ab30

Thank you.

vabene1111:
i will take a look and see what can be done 👍

Issue #291, opened 2020-12-29
URL Import Not Working


ClydeDroid:
### Version
0.12.1

### Issue
When I try to import a recipe from a URL, it seems to not be able to load any pages, even ones with valid ld+json data (eg https://www.allrecipes.com/recipe/60598/vegetarian-korma/). I just get the error `The requested page could not be found.`:
![image](https://user-images.githubusercontent.com/9121162/103309752-c7d86080-49ca-11eb-9e67-5fe303923442.png)

### `.env`
Please include your `.env` config file (**make sure to remove/replace all secrets**)
```
# only set this to true when testing/debugging
# when unset: 1 (true) - dont unset this, just for development
DEBUG=1

# hosts the application can run under e.g. recipes.mydomain.com,cooking.mydomain.com,...
ALLOWED_HOSTS=*

# random secret key, use for example base64 /dev/urandom | head -c50 to generate one
SECRET_KEY=...

# your default timezone
TIMEZONE=America/Los_Angeles

# add only a database password if you want to run with the default postgres, otherwise change settings accordingly
DB_ENGINE=django.db.backends.postgresql
POSTGRES_HOST=db_recipes
POSTGRES_PORT=5432
POSTGRES_USER=djangodb
POSTGRES_PASSWORD=...
POSTGRES_DB=djangodb

# the default value for the user preference 'fractions' (enable/disable fraction support)
# when unset: 0 (disabled)
FRACTION_PREF_DEFAULT=1

# the default value for the user preference 'comments' (enable/disable commenting system)
# when unset: 1 (true)
COMMENT_PREF_DEFAULT=1

# Users can set a amount of time after which the shopping list is refreshed when they are in viewing mode
# This is the minimum interval users can set. Setting this to low will allow users to refresh very frequently which
# might cause high load on the server. (Technically they can obviously refresh as often as they want with their own scripts)
SHOPPING_MIN_AUTOSYNC_INTERVAL=5

# If staticfiles are stored at a different location uncomment and change accordingly
# STATIC_URL=/static/

# If mediafiles are stored at a different location uncomment and change accordingly
# MEDIA_URL=/media/

# Serve mediafiles directly using gunicorn. Basically everyone recommends not doing this. Please use any of the examples
# provided that include an additional nxginx container to handle media file serving.
# If you know what you are doing turn this back on (1) to serve media files using djangos serve() method.
# when unset: 1 (true) - this is temporary until an appropriate amount of time has passed for everyone to migrate
[...]

PR #475, opened 2021-03-04
Refactored URL Import

### LINKED ISSUE #376: Serious Eats Recipe Import
Adding support to import recipes from seriouseats.com Example recipe URL: https://www.seriouseats.com/recipes/2011/12/serious-eats-halal-cart-style-chicken-and-rice-white-sauce-recipe.html

Import recipes for this domain similar to other recipe sites.

smilerz:
Added library recipe-scraper (maintained https://github.com/hhursev/recipe-scrapers) and refactored the import helper to use the library.
Leveraging community maintained scraper will increase coverage and keep scrapers more up to date.  Also adds capability to do web based scraping in addition to ld+json.

This pull also fixes #376, #291, #465, and #462.


vabene1111:
thank you very much 👍  i have seen this package before and while i hate that it is required (as more pages should adhere to the standard) i think it is a great solution.

I will need to do some very thorough testing and then should probably be able to integrate it. Will take some time tough as there are some other things first in line that need to get done and i dont have much time for recipes atm

PR #480, opened 2021-03-06
Update Italian Translation

auanasgheps:
Italian translation updates.
Manual PR since Transifex doesn't do it.


vabene1111:
thank you! i will see if i can get transifex to open the PR manually if that works i will discard this PR. Any way this will be in the next update :)

PR #499, opened 2021-03-19
Add iOS webclip icon support

### LINKED ISSUE #432: iOS WebClip support
There is a way to add a 'WebClip' in iOS that opens any website as if it was a native iOS app. Users just have to 'pin' the website to their Homescreen when inside safari. It already works for Tandoor, but it could be better: it is missing a proper icon. You can also specify wether it should be opened fullscreen, with/without safari UI and more.

**Additional context**
I found some apple documentation on this, along with other website explaining how to add this functionallity.

Apple Documentaion:
* https://developer.apple.com/documentation/devicemanagement/webclip
* https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html

Tutorials:
* https://webdesign.tutsplus.com/articles/quick-tip-give-your-website-an-ios-home-screen-icon--webdesign-10067
* https://www.create.net/support/adding-an-ios-home-screen-icon-for-your-website

sebimarkgraf:
Fixes #432 
Icons are generated using https://realfavicongenerator.net/
Using SVGs as favicons is not really supported.

Mac allows this, but the icon needs to be specified differently as "mask-icon".


sebimarkgraf:
And I tested the iOS webclip on an iPhone 11 Pro and an iPad.
Tested it in Chrome and Safari on Mac, but I am not sure how to pin the webpage to the dock on Mac.

vabene1111:
thank you very much, this is awesome i could not have tested this if i had implemented it.

PR #501, opened 2021-03-19
fix json direct import when wrapped in @graph

smilerz:
also added description import when using direct json import


vabene1111:
thank you very much.

PR #510, opened 2021-03-28
Cheftap import tweak

ModdedGamers:
When I submitted my testing .zip, I didn't submit my normal export, because that has some personal data in the recipes. Instead, I made one by had, which was a mistake. The folder contained inside is also called cheftap_export. This fixes the code and the docs for ChefTap imports.


vabene1111:
a thank you very much !

PR #534, opened 2021-04-09
fix backend detection

smilerz:
RecipeFilter was only checking for postgresql_psycopg2 - added postgresql


vabene1111:
i thought i already added that a while ago 🤔 thanks for the PR, will get it merged asap

PR #566, opened 2021-04-18
Nginx with subfolder

smilerz:
this change allows nginx to mount tandoor to a subfolder when proxying.  no change in config is required within the application to enable this behavior allowing seamless switching between hosted (nginx) and local debug


smilerz:
this breaks the api - will update and resubmit

PR #571, opened 2021-04-20
Nginx

smilerz:
This PR allows for nginx to be setup at a subfolder and includes the following modifications:
- .env.template: add variable script_name that can be used (django_rest config in settings already references this variable)
- custom tag created to build a SCRIPT_NAME (subfolder), BASE_PATH (full base URL including subfolder)
- base.html two values are stored in localStorage:  script_name and base_path
- base.html: service worker scope is modified to include any value from script_name
- base.ts (api_generator) retrieves values from local storage to build BASE_PATH
- wsgi.py rewritten to check for a HEADER from nginx to determine if default urls should be used or a subfolder pre-pended.
- an example nginx conf to utilize subfolder configuration


vabene1111:
very cool, thanks. I will take a look at what this does exactly and ship it with the next update.

vabene1111:
Ok so i just reviewed the code and have two questions

1. the vetur.config.js is a local artefact of your build environment right ? 
2. the new nginx config is already in the volume automatically mounted into nginx. wouldn't that get automatically activated and cause issues ? (haven't tried it yet, just making sure ?)

smilerz:
yes, the vetur config is local.  I attempted to add it to .gitignore and it never seemed to take.

and you are correct on the nginx config.  it should be rename recipe.subfolder.conf.sample

vabene1111:
i am pretty sure that nginx discovers all kinds of files in its directory. 

I guess i will just create a docs page for it and tell the user to put it in the volume if sub paths should be used

smilerz:
I'm using a swag nginx container and that's how they configured the sample configs for various apps.  but I can't say with certainty that it would work for everything.

Mine nginx config works now so providing instructions works for me.  :)

vabene1111:
haha ok, i will try to verify it further 😂

smilerz:
I should point out that I’m not using the mounted folder for nginx  configuration.  My setup is a bit different.  So to keep it simple approach it any why you’d like.

PR #574, opened 2021-04-21
Full text search

smilerz:
full text search is working with v2 of search.
search box now searches across recipe name, description, food and keyword (existing additional filters still in place)
search box also supports complex filters 
e.g. -egg flour or (sugar and butter) would return recipes with
flour in name, description, keyword or food
OR
recipes with sugar and butter in one of those 4 fields
but no recipes with egg in any of those 4 fields

I tried multiple approaches with icontains and/or trigrams - they all made search much worse.


smilerz:
here is an article that goes into painful detail on related matters:
http://logan.tw/posts/2017/12/30/full-text-search-with-django-and-postgresql/#update-search-vectors

smilerz:
I finally figured out how to get unaccent and trigrams included in the full text search.  latest push includes those features

smilerz:
I am refactoring the full-text search to allow more user configuration - should be done soon.

PR #578, opened 2021-04-26
New Importer - RecetteTek

itsmegb:
This PR adds an importer for RecetteTek - [https://www.recettetek.com/en/](https://www.recettetek.com/en/)

It is currently possible to import either a single recipe or import all exported recipes at once.

RecetteTek allows a user to export their recipes as either an '.RTK' file or an '.HTML' file, this importer supports only the RTK format.

It will also attempt to download the original image that RecetteTek initially imported, but if that fails it will not import any image. The recipe export from RecetteTek does include the original images locally, but I have yet to look at the best way to import those.

A lot of the fields that RecetteTek have are free-form, so things like serving size, prep time, cooking time etc will fail to import if there is anything other than an integer in the field. (Maybe a parser could be built for this for use by all importers, similar to the ingredient parser?)

Unlike some of the other importers, I've split out a lot of the value parsing into their own try/catch blocks to get as much info as possible, I'm sure there is probably a better way to do this though :)

RecetteTek also allows videos to be inserted as a recipes instructions, in this case there will be no ingredients or description.

### DIFF
--- a/cookbook/forms.py
+++ b/cookbook/forms.py
@@ -112,6 +112,7 @@ class ImportExportBase(forms.Form):
     SAFRON = 'SAFRON'
     CHEFTAP = 'CHEFTAP'
     PEPPERPLATE = 'PEPPERPLATE'
+    RECETTETEK = 'RECETTETEK'
     RECIPESAGE = 'RECIPESAGE'
     DOMESTICA = 'DOMESTICA'
     MEALMASTER = 'MEALMASTER'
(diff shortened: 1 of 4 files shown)


vabene1111:
thanks for the PR, sounds really good from the description. i will give it a test and try to get it into the next release

itsmegb:
I've updated the image importing, it will now retrieve the image that is in the export file first and if that fails it will then attempt to download the image from the original location.
I found that this works better, as some of the original image URLs no longer exist or are unreachable for some reason.

itsmegb:
As an additional question, is there a recommended location to store the original source url of the recipe?

For now i've just copied what other importers have done and placed it at the bottom of a step, however, I noticed that the recipe DB object contains two fields: Link and Cors Link
What is the purpose of these fields? Should/Could i store the original URL in one of those?

vabene1111:
You have done it correctly, i have no special field for "source URL" like many other recipe managers have as i do not really think that this is very useful. 

The link and cors link fields are leftovers of the old external recipe system that just produced links to external recipes.

itsmegb:
i made the decision to append the url to the end of the description field, i feel that makes more sense than having it added to the first step in the recipe.

Is there a way to add the ingredients directly to the recipe rather than one of the steps?

vabene1111:
No ingredients are always linked to a step. Also all instructions are so you will need to add at least one step to the imported recipe.

I recommend adding the source URL to the text as the description will be shown on the search page from the next update on which will probably not look good. Also it will be consistent with all other importers

itsmegb:
No problem, i've moved the source URL out of the description and back into the step.

PR #591, opened 2021-04-30
New Importer - Reciepe Keeper

itsmegb:
A new importer for Recipe Keeper - https://www.recipekeeperonline.com/

Most items are implemented with the exception of Prep/Cook time. Recipe Keeper exports the time using [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Durations) duration format and I can see that at least one other importer also is looking to import times in this format (Mealie), so I will likely write a helper function to assist with converting these times into an integer representation of minutes.

- [ ] Remove the step number from the step text (markdown resets it to 1 on every step)
- [ ] Import prep/cook times
- [ ] Import quantity/servings


vabene1111:
Good idea with the helper! Maybe we could do it similarly to the ingredient parser where a general helper function first recognizes the format and then parses it into the standard minutes format from tandoor

hay-kot:
> Recipe Keeper exports the time using ISO 8601 duration format

Mealie dev here. I have a little helper function for this that may be of use. [Check it out](https://github.com/hay-kot/mealie/blob/be378cb20c6d5e6c8fca940cecd892a61742842d/mealie/services/scraper/cleaner.py#L165-L246)

itsmegb:
Amazing, Thanks @hay-kot, i'll take a look.

I did also find the [isodate](https://github.com/gweis/isodate) library that parses all of the ISO 8601 format strings. It may be overkill though, as only the duration is needed.

I've started to work on something, so I'll see how I get on.

vabene1111:
merging this and adding the required todos myself, thanks for contributing this importer !

itsmegb:
No problem, sorry i didn't get more done! Life got in the way!

vabene1111:
no worries, you did all the hard lifting

PR #598, opened 2021-05-02
updated to handle new behavior of recipe_scrapers

smilerz:
They changed a lot on how the scraper works, so I had to rework some of the logic.  it is really repetitive now, but it works and all the tests pass now.


vabene1111:
thanks for fixing the remaining issues !

Issue #325, opened 2021-01-09
Ingredients scaling should not go to 0 or negative.


charleszlu:
### Version
Version: 0.13.0

### Bug description
When scaling ingredients, the counter should not be able to go to 0 or negative.

vabene1111:
you are correct. There was a PR not long ago but i think merging servings reverted some of the changes. 

While this is nice its not really something important so only classified as enhancement, will fix it when i get around to it.

Note to future self: Make sure 0.XY is supported

PR #604, opened 2021-05-05
Fix servings input field text direction

### LINKED ISSUE #526: Weird Behaviour of Servings Field
### Version
v0.15.0 but already existed before

### Bug description
When trying to type a new servings number in the servings field the field behaves weirdly depending on the browser.
Numbers are added at the wrong place or cannot be deleted.

### Potential Solution
I assume this is due to the "rtl" text direction. As this problem is more noticeable on mobile phones, a fix could include setting
the text direction to "ltr" again and change the styling by adding the spin buttons manually.
One could try to integrate something like this or write a custom version: https://vuejsexamples.com/a-customizable-number-input-spinner-component-for-vuejs/

sebimarkgraf:
This fixes #526 and #325

I guess there is no reason to set the text direction to rtl.
To keep the styling I added the text-align: right.

The search didn't show any css that depends on the rtl attribute on purpose, so I guess this should not have styling side affects.

### DIFF
--- a/vue/src/apps/RecipeView/RecipeView.vue
+++ b/vue/src/apps/RecipeView/RecipeView.vue
@@ -54,7 +54,7 @@
             </div>
             <div class="my-auto" style="padding-right: 4px">
               <input style="text-align: right; border-width:0px;border:none; padding:0px; padding-left: 0.5vw; padding-right: 8px; max-width: 80px"
-                     value="1" maxlength="3"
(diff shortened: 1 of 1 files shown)


vabene1111:
Interesting, i thought i tried this .. but whatever, thanks for fixing, this will be included in the next release. 

Also sorry for this taking so long for me to fix, really busy right now irl

PR #616, opened 2021-05-17
Update authentication.md

lbarnett83:
Added a semi-colon that while missing would cause a reverse proxy failure.


vabene1111:
thanks for fixing, will be include in next release

PR #622, opened 2021-05-20
updated url import and test due to updated recipe_scraper

smilerz:
one of the tests - thespruceeats.com - is still broken due to upstream bug, which has been reported.


vabene1111:
ah thanks, i wanted to look into fixing this today ..

PR #633, opened 2021-05-28
bumped recipe_scrapers and updated expected results

smilerz:
url imports all pass now.

doublecheck the file path change you made - i had to fix it to work on linux  are paths different between linux/windows?


vabene1111:
merged, thanks, as discussed i will fix the windows paths probably tomorrow

PR #673, opened 2021-06-08
Create image_processing.py

MaxJa4:
Added central function to rescale and compress images for recipes (or in general).
Switched from previously used PNG format to 75% JPEG format for a 5-10x file size reduction with hardly any quality loss.

### DIFF
--- a/cookbook/helper/image_processing.py
+++ b/cookbook/helper/image_processing.py
@@ -1,0 +1,15 @@
+from PIL import Image
+from io import BytesIO
+
+
+def rescale_image(image_object, base_width=720):
+    img = Image.open(image_object)
+    icc_profile = img.info.get('icc_profile')  # remember color profile to not mess up colors
+    width_percent = (base_width / float(img.size[0]))
+    height = int((float(img.size[1]) * float(width_percent)))
+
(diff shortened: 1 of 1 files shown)


MaxJa4:
Feel free to move this file to a more suitable location as you see fit.

vabene1111:
thanks very much, i will take a look at implementing this accross the application

MaxJa4:
Maybe a one time execution after the next update containing this feature would be a good idea to compress all already existing pictures.

I'm gonna make a little test and do that with my instance manually to see how much smaller the overview page gets and how faster it loads (the first time, since caching is active on my page).

vabene1111:
i am always worried with file migrations but you are generally right .. we will see what can be done after installing

MaxJa4:
That's true... maybe implementing it in one version and migrating the existing files (just PNGs) in the follow version reduces that migration risk a little since the new compression has been tested live already (which should be fine but you never know).

MaxJa4:
Previous format (like there were rescaled before; 70% PNG; reload without caching):
image

New format (75% JPEG optimized; reload without caching):
image

This was on the /search/ site (my default site). I'd call it a success :)

vabene1111:
i think that sounds very good 👍

PR #751, opened 2021-07-07
Update manual setup

tandy-1000:
I just went through a manual install on DietPi / Debian Unstable, and found that the manual needed some improving, this method works, whereas following the original manual was not a success for me.

The only line I would say I am unsure about is this one:
"Give the user permissions: `chown -R recipes:www-data /var/www/recipes`"
As I am not sure whether these permissions are necessary for this to work.

Hope this is useful, let me know if I can make further improvements.


vabene1111:
Thank you very much. I will review this and get it merged.

I intentionally left the manual install vague as there are simply to many options to deploy this but from a first very quick look your changes seem to make sense

PR #777, opened 2021-07-26
New keywords

smilerz:
finally got my development box back in working order.

This is a rebase of my keyword branch on your current develop branch.

Everything passes tests - do you know what wasn't working so I can work on bug fixes?


vabene1111:
ok so its been a few days but i remeber the following

- you can use feature/fulltext-search as it contains both the keywords rework and the search rework and has all migrations already merged (except 2 new ones i think)
- my main concern was and is that the tree serialization and all the functions associated with it need to have proper tests to prevent tennant leaks (for example it needs to be made sure that you cannot see someone elses keywords but guessing ids and adding sub keywords etc.)
- this is especially important in the automatic methods that try to find a root node based on names or ids, since treebeard uses raw queries i do not know if `django-scopes` can filter those and we need to make absoultuly sure that no tennancy rule is broken 
- this is also valid for the search rework: we need to make sure that all indices etc. are build so that  they are individual for each space (i think this is already the case but we need to make sure)

other than that i dont remeber any pressing issues, just a few TODOs you left in the code and i think i added some, but i think after tenency support is verified we can do a proper review of all functions and see what needs to be done.

PR #787, opened 2021-07-31
Search and keywords

smilerz:
this has been tested against both sqlite and postgres passing all tests.
tests include all API actions cross-space to ensure space boundaries are respected.


smilerz:
@vabene1111 - merge with -X theirs to accept all the incoming changes.

smilerz:
one more note: I moved the manual schemas (so far just recipes and keywords) to schema.py

vabene1111:
> one more note: I moved the manual schemas (so far just recipes and keywords) to schema.py

these are the ones used for the API endpoint documentation right ? 

apart from that i hope to get to reviewing and merging this soon.

cookbook/migrations/0121_auto_20210518_1638.py
```diff
@@ -18,6 +18,6 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='userpreference',
             name='use_fractions',
-            field=models.BooleanField(default=False),
+            field=models.BooleanField(default=True),
```
vabene1111, line 21: as only the Americans use fractions as far as i know and decimals are far superior in my opinion i would like to leave the default as decimals.

cookbook/migrations/0145_alter_userpreference_use_fractions.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('cookbook', '0144_alter_userpreference_search_style'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='userpreference',
+            name='use_fractions',
+            field=models.BooleanField(default=False),
```
vabene1111, line 16: ok i see, the above change is reverted. Leaving this unreviewed/open comment to verify behavior

cookbook/serializer.py
```diff
@@ -45,7 +47,7 @@ def to_internal_value(self, data):
 class SpaceFilterSerializer(serializers.ListSerializer):
 
     def to_representation(self, data):
-        if type(data) == QuerySet and data.query.is_sliced:
+        if (type(data) == QuerySet and data.query.is_sliced) or type(data) == MP_NodeQuerySet:
```
vabene1111, line 50: can you add a comment here why `type(data) == MP_NodeQuerySet` should not be filtered for space ?

This is an extremely critical security function as not filtering where a filter would be necessary could leak data between spaces so everyone needs to understand why a certain piece of code exists.

cookbook/serializer.py
```diff
@@ -200,20 +202,36 @@ class Meta:
...
 class KeywordSerializer(UniqueFieldsMixin, serializers.ModelSerializer):
     label = serializers.SerializerMethodField('get_label')
+    image = serializers.SerializerMethodField('get_image')
+    numrecipe = serializers.SerializerMethodField('count_recipes')
 
     def get_label(self, obj):
         return str(obj)
 
+    def get_image(self, obj):
+        recipes = obj.recipe_set.all().exclude(image__isnull=True).exclude(image__exact='')
```
vabene1111, line 212: i think we should add a `space=self.request-space` filter here so that in the unlikely Szenario that a keyword is added to a foreign recipe no data is leaked when returning the space image.

cookbook/tests/api/test_api_keyword.py
```diff
@@ -125,24 +190,189 @@ def test_add_duplicate(u1_s1, u1_s2, obj_1):
...
+    r = u1_s1.put(
+        reverse(MOVE_URL, args=[obj_1.id, 9999])
+    )
+    assert r.status_code == 400
+
+    # attempt to move to wrong space
+    r = u1_s1.put(
+        reverse(MOVE_URL, args=[obj_1_1.id, obj_3.id])
+    )
+    assert r.status_code == 400
```
vabene1111, line 264: should this call return 404 ? most space related api calls return 404 when attempting to use an object that does not belong to ones space in order to prevent leaking information about the existente or non existente of certain objects.

cookbook/tests/api/test_api_keyword.py
```diff
@@ -125,24 +190,189 @@ def test_add_duplicate(u1_s1, u1_s2, obj_1):
...
+
+    # run diagnostic to find problems - none should be found
+    with scopes_disabled():
+        assert Keyword.find_problems() == ([], [], [], [], [])
+
+
+# this seems overly long - should it be broken into smaller pieces?
```
vabene1111, line 271: i think this method is fine. Its well structured and each test is commented so that it is clear what it does.

recipes/settings.py
```diff
@@ -12,22 +12,19 @@
...
-
+# from dotenv import load_dotenv
+# load_dotenv()
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
 # Get vars from .env files
 SECRET_KEY = os.getenv('SECRET_KEY') if os.getenv('SECRET_KEY') else 'INSECURE_STANDARD_KEY_SET_IN_ENV'
 
 DEBUG = bool(int(os.getenv('DEBUG', True)))
+DEMO = bool(int(os.getenv('DEMO', False)))
```
vabene1111, line 27: this flag was removed

smilerz:
> > one more note: I moved the manual schemas (so far just recipes and keywords) to schema.py
> 
> these are the ones used for the API endpoint documentation right ?

that's correct -

cookbook/tests/api/test_api_keyword.py
smilerz: 404 is fine - I didn't realize you had a convention in place.

cookbook/migrations/0121_auto_20210518_1638.py
smilerz: for some reason that keep leaking in - every time I run a migration it adds it in.  I usually take it out but must have forgotten one.

cookbook/serializer.py
smilerz: I'll add a comment - but to discuss it here, the MP_NodeQuerySet doesn't have a filter function and would only come from a API call and not a nested serializer, so matches the same criteria that is_sliced was checking for.

smilerz: I looked into this more - I am not using that mixin for Keywords - it didn't work.  Instead I overrode the queryset in the API for a variety of reasons.   I'll simplify the override to include .filter(space=space) as a final step to ensure that it is always captured.

smilerz:
push 080b4d3 should incorporate the changes you requested.

smilerz:
I added a migration to find leading or trailing spaces in Keyword, Food and Unit.  it just prepends or appends a '_' to the name.

I started running into issues as patches (partial updates) seem to automatically trim the results!

cookbook/tests/api/test_api_keyword.py
vabene1111: technically using incrementing IDs is already a bad practice but i prefer them over guids but that means that one should not be able to verify the existence of objects by incrementing ids

vabene1111:
ok everything looks good, just need to merge it now and rewrite the migration line, hope to get that done today or tomorrow

smilerz:
> ok everything looks good, just need to merge it now and rewrite the migration line, hope to get that done today or tomorrow

Double check that the webpacks are minimized.  I may have missed turning minimize back on and building.

PR #820, opened 2021-08-20
Send proper protocol to Django from Nginx

nemec:
In order to run over HTTPS* django requires the X-Forwarded-Proto header sent from the nginx proxy. I didn't add everything needed for TLS, but this header works with HTTP and HTTPS so I figure it might save a few people some time if they ever decide to secure their installation.

\* specifically, I noticed that image URLs are requested over HTTP instead of HTTPS even when the main page is HTTPS until this header is sent


vabene1111:
this definitely makes sense and the header is already included in all non manual setups, so it needs to be in manual as well. Thanks!

PR #846, opened 2021-08-31
Food tree

smilerz:
Several changes:
- convert food to MP_Node tree along with all necessary tests
- create generic split list view to handle mulitple model types 
  - food only model that uses it.  keyword and unit should be easy to implement
- create TreeMixin to manage custom API methods to handle trees for both keyword and food
- create merge Mixin to manage merge functionality as a general use case (looking forward on units)
- create genericAPI to wrap OpenAPI generator.  Single method can be used to call any model/function
  - setup required in model.js to describe the parameters
- create generic modal forms to handle arbitrary number/types of input fields
- refactored GenericMultiSelect to use new genericAPI


smilerz:
@vabene1111 - I updated the github actions to only run on your repository (I think).  Please confirm that they are still functioning.

cookbook/models.py
```diff
@@ -37,16 +37,73 @@ def get_model_name(model):
...
-        if len(q) != 0:
-            return q[0], False
-        else:
+        try:
+            return self.get(name__iexact=kwargs['name'], space=kwargs['space']), False
+        except self.model.DoesNotExist:
             with scopes_disabled():
-                node = self.model.add_root(**kwargs)
-            return node, True
+                return self.model.add_root(**kwargs), True
```
vabene1111, line 47: do we need to disable scopes here as `space` should always be present in `kwargs` and if not we want this function to fail?

vabene1111: other than that i love the generic approach

cookbook/models.py
```diff
@@ -37,16 +37,73 @@ def get_model_name(model):
...
+    def has_children(self):
+        return self.get_num_children() > 0
+
+    def get_num_children(self):
+        return self.get_children().count()
+
+    # use self.objects.get_or_create() instead
+    @classmethod
+    def add_root(self, **kwargs):
+        with scopes_disabled():
```
vabene1111, line 102: same as above

cookbook/models.py
smilerz: The MP_Tree needs to see the entire tree to work correctly when dealing with root objects, if it doesn't it will add a duplicate path which will cause a unique constraint error.  
We can explicitly check for the existence of 'space' and trigger a failure if it is not present.

cookbook/models.py
smilerz: the scopes_disabled might be redundant in one of these spots, left over when I was reorganizing the code.

PR #849, opened 2021-09-01
Allow specifying port through TANDOOR_PORT environment variable

cschmatzler:
While Docker containers have their own IP address namespace and port collisions are impossible to achieve, other container solutions share one.
Podman, for example, runs all containers in a pod under one namespace, which results in every port only being allowed to be assigned once. Having a default port in the image that cannot be changed can cause collisions and incompabilities.

This change allows setting a custom port for Gunicorn through the environment while keeping the default port as fallback to keep backwards compatibility.

### DIFF
--- a/boot.sh
+++ b/boot.sh
@@ -1,6 +1,8 @@
 #!/bin/sh
 source venv/bin/activate
 
+TANDOOR_PORT="${TANDOOR_PORT:-8080}"
+
 echo "Updating database"
 python manage.py migrate
 python manage.py collectstatic_js_reverse
@@ -9,4 +11,4 @@ echo "Done"
 
 chmod -R 755 /opt/recipes/mediafiles
 
-exec gunicorn -b :8080 --access-logfile - --error-logfile - --log-level INFO recipes.wsgi
+exec gunicorn -b :$TANDOOR_PORT --access-logfile - --error-logfile - --log-level INFO recipes.wsgi


cschmatzler:
Will do, out of home at the moment, will get to it tomorrow! Thanks for letting me know I forgot about it.
Anywhere else in the documentation that you would like it added?

Sep 2, 2021 3:10:50 PM smilerz ***@***.***>:

smilerz:
I don't know enough about Docker, but does Dockerfile need modified too in order to accommodate this?

cschmatzler:
The only reference we have in the Dockerfile is `EXPOSE`.  
This does not actually expose anything on its own, just creates a shortcut to that port being exposed by using the `-P` flag on `docker run. I'm personally not a big fan of it because it implicitly exposes network ports, but it's a matter of taste. Since it's a build step in the Dockerfile, we cannot change that to a variable port.

You guys would have to decide whether to remove it altogether (docs already use the explicit `-p` flag everywhere as far as I can tell) or leave it in as a default.

vabene1111:
thanks for the feedback. 

@cschmatzler can the `EXPOSE` really be removed without breaking existing setups? Wouldn't the nginx container loose access to the application container if the used port (8080) is not exposed?

I honestly dont know to much about it but every example dockerfile i saw used this syntax.

ambroisie:
> @cschmatzler can the `EXPOSE` really be removed without breaking existing setups? Wouldn't the nginx container loose access to the application container if the used port (8080) is not exposed?

IIRC, expose is stricly for documentation purposes, it does not modify the image/container in any meaningful way.

vabene1111:
sorry for taking so long, its merged now

PR #860, opened 2021-09-04
Optional nginx template to support subfolder configuration

smilerz:


vabene1111:
did we merge things into this branch that dont belong here ? This is mostly done, its just the template that we need to add to the docs right ? 

i remember i wasn't sure if nginx would like the second config, i guess its safer in the docs ?

smilerz:
The config is all that is missing.  Documentation is probably fine.

vabene1111:
added it to the docs

Issue #854, opened 2021-09-03
Error Deleting Food


smilerz:
### Version
0.17 develop branch

### Bug description
When attempting to delete a Food that has been removed from a recipe:
`django.db.models.deletion.ProtectedError: ("Cannot delete some instances of model 'Food' because they are referenced through protected foreign keys: 'Ingredient.food'."`

Should that foreign key cascade?  what are the implications?

vabene1111:
interesting question. The reason the delete fails is that the food is used in an ingredient. At this point no further check is done in order to verify if the ingredient itself belongs to a recipe. 

Technically ingredients should be deleted when they are removed from a recipe. This is something I wanted to look into for quite a while as this seems to be the default behavior of all django rest framework functions (leaving the entry when a reference is removed) which makes sense but not in our/all cases. So i think this question comes down to how to properly remove references that dont need to exist on there own if separated from their parent (mainly ingredients). Can this be done in the serializer or is a signal the way to go.

smilerz:
> interesting question. The reason the delete fails is that the food is used in an ingredient. At this point no further check is done in order to verify if the ingredient itself belongs to a recipe.
Is an Ingredient ever part of more than 1 recipe?

vabene1111:
no it is not and i dont think this could ever be useful so it is definitely safe to introduce some kind of deletion method on removal of the recipe link

smilerz:
Putting this here for future reference.  post/pre-delete signal is probably the most effective way to do this.

https://stackoverflow.com/questions/12655347/django-cascade-delete-on-reverse-foreign-keys

vabene1111:
never really looked into signals, they sound like a very nice technology but was never sure how performant they are, but the solution looks good to me!

smilerz:
I have a post-save signal implemented as part of FTS to update the search_vector fields in signals.py.

In general, the app is far more read intensive than write, so sacrificing write performance shouldn't be much of a burden.

vabene1111:
#861 fixes this for now but i think i will rewrite to signals at some point reverting the change made in ab316236f97c820a07292196c95b759b2dc93e6a so that foods no longer cascade ingredients but ingredients no longer exist if they are not needed anymore

smilerz:
BTW - the signal method never worked - Django was checking for constraints prior to the pre-delete signal.  I just overrode the destroy method on the Food model.

smilerz:
no longer relevant in v2

PR #861, opened 2021-09-04
post food_tree bug fixing

smilerz:
a couple minor bug fixes.
one notable change - I removed the get_or_create on supermarket_category from the create method of FoodSerializer.
I will add an explicit create functionality on the modal form (vue-multiselect already supports this), but otherwise will assume that the Supermarket Category already exists.  I am not 100% sure that this doesn't have impacts elsewhere.


smilerz:
ab316236f97c820a07292196c95b759b2dc93e6a fixes #854

PR #866, opened 2021-09-07
implemented pagination on Log apis

### LINKED ISSUE #484: API pagination
implement proper pagenation/result size limiting for the API at least for the Log Views which are sliced by default and therefore broken in the detail views

- [ ] re-enable the appropriate tests

smilerz:
fixes #484 

made some guesses on expected behavior, added comments as to choices.  Validate before merging

### DIFF
--- a/cookbook/tests/api/test_api_cook_log.py
+++ b/cookbook/tests/api/test_api_cook_log.py
@@ -33,21 +33,21 @@ def test_list_permission(arg, request):
 
 
 def test_list_space(obj_1, obj_2, u1_s1, u1_s2, space_2):
-    assert len(json.loads(u1_s1.get(reverse(LIST_URL)).content)) == 2
-    assert len(json.loads(u1_s2.get(reverse(LIST_URL)).content)) == 0
+    assert json.loads(u1_s1.get(reverse(LIST_URL)).content)['count'] == 2
+    assert json.loads(u1_s2.get(reverse(LIST_URL)).content)['count'] == 0
 
     obj_1.space = space_2
     obj_1.save()
 
-    assert len(json.loads(u1_s1.get(reverse(LIST_URL)).content)) == 1
-    assert len(json.loads(u1_s2.get(reverse(LIST_URL)).content)) == 0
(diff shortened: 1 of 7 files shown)


cookbook/tests/api/test_api_cook_log.py
```diff
@@ -33,21 +33,21 @@ def test_list_permission(arg, request):
...
+    assert json.loads(u1_s2.get(reverse(LIST_URL)).content)['count'] == 1
 
 
 @pytest.mark.parametrize("arg", [
     ['a_u', 403],
-    ['g1_s1', 404],
+    ['g1_s1', 403],  # changed expected value.  based on list permissions the log is visible, but not editable
     ['u1_s1', 200],
-    ['a1_s1', 404],
+    ['a1_s1', 403],  # changed expected value.  based on list permissions the log is visible, but not editable
```
vabene1111, line 50: interesting, this is definitely an oversight on my part because i do not think any user should be able to see logged cooking events of other users or maybe this could be explicitly added as a sharing feature in the future. I will fix this after merging

cookbook/views/api.py
```diff
@@ -598,35 +598,31 @@ class ViewLogViewSet(viewsets.ModelViewSet):
...
-            return self.queryset
+        # working backwards from the test - this is supposed to be limited to user view logs only??
+        return self.queryset.filter(created_by=self.request.user).filter(space=self.request.space)
 
 
 class CookLogViewSet(viewsets.ModelViewSet):
     queryset = CookLog.objects
     serializer_class = CookLogSerializer
-    permission_classes = [CustomIsOwner]
+    permission_classes = [CustomIsOwner]  # CustomIsShared? since ratings are in the cooklog?
```
vabene1111, line 611: currently all ratings only apply to the user itself. i think we can add a privacy setting that allows to share ratings and cook logs with other users or even completely change the behavior to always be public inside a space but that is a separate topic to look at i think

vabene1111: thats why IsOWner is correct

PR #869, opened 2021-09-08
Move/Merge updated to handle object vs id

smilerz:
Move/Merge modal now work with change to object in drop down.


vabene1111:
ah i missed that, thanks

PR #882, opened 2021-09-11
README: Fix Docker installation URL

MadWalnut:
The Docker installation URL in README seems outdated, so I replaced it with the working one.


vabene1111:
interesting, i remember some time ago this was changed to include the .html and wasn't working without it, now its back to normal 😂 thanks for the fix

PR #907, opened 2021-09-15
Performance

smilerz:
fixes speed when importing large number of recipes.

**note**
turning off node_ordering leaves the facets (nested representation of the food) unsorted.  one possible work around would be to leave sorting on and disable before import, turn it back on and run a command to reorder the nodes
```
Food.node_order_by=[]
-----import happens here----
Food.node_order_by=['name]
with scopes_disabled():
    Food.fix_tree(fix_paths=True)
```
another approach would be to temporarily enable ordering on a schedule and resort the tree


vabene1111:
couldn't we also sort when making the API request ? I font see why existing instances in the DB should be modified to sort, the index PG keeps on the path column should allow to easily sort when querying or am i understanding something wrong ?

smilerz:
> couldn't we also sort when making the API request ? I font see why existing instances in the DB should be modified to sort, the index PG keeps on the path column should allow to easily sort when querying or am i understanding something wrong ?

In any view except one that gets a representation of the tree all at once like this one.
image

The method that builds that tree (annotated_qs) relies on path/depth to derive parent/child relationships with only a single iteration through the list.  To serve it back to the browser sorted you'd have to do N additional sorts for each set of parent/children.

smilerz:
> I'd like to test one additional method to enable a 'bulk' mode on get_or_create.  If you explicitly insert at the end of the tree does a subsequent insert while sorted sort the whole tree, or just itself?

FYI - this didn't pan out.

vabene1111:
ok thanks for the feedback. I guess we will have to see how this works out. Using bulk around the importer is likely a very important and useful thing but proabbaly requires some major refactoring of the whole importer stack (which is due anyway as i made some wrong assumtions at the beginning and have already identified several places for improvement). 

For the example you show: do you think we could just abuse the frontend to sort those entries?

smilerz:
> For the example you show: do you think we could just abuse the frontend to sort those entries?

You can make code do anything :)
In the short term the worst case scenario is we can point people to the Admin interface which allows drag&drop reordering.
image

Longer term we can decide between several options:

1. bulk import
2. client side sorting
3. 'Sort Now' function
4. scheduled re-sorting

Options 1, 3 & 4 would need some testing to make sure they are safe during simultaneous writes and probably wrapped inside a transaction with some heavy duty error checking.

vabene1111:
i think that sounds like a very good plan 👍

PR #912, opened 2021-09-22
moved validation from part of step to part of recipe

smilerz:
prior attempt checked if food is part of ingredient__step but some foods that are not part of a recipe are still part of an orphaned step.  Moved validation to checking if part of recipe.  Also added check to prevent deleting food that has children that are part of recipes.  Modified Unit to allow deletion of orphaned units that are part of Ingredient and/or Step but are not linked to a Recipe.

### DIFF
--- a/cookbook/models.py
+++ b/cookbook/models.py
@@ -43,8 +43,7 @@ class TreeManager(MP_NodeManager):
         try:
             return self.get(name__exact=kwargs['name'], space=kwargs['space']), False
         except self.model.DoesNotExist:
-            with scopes_disabled():
-                return self.model.add_root(**kwargs), True
+            return self.model.add_root(**kwargs), True
 
 
 class TreeModel(MP_Node):
@@ -363,6 +362,12 @@ class Unit(ExportModelOperationsMixin('unit'), models.Model, PermissionModelMixi
     def __str__(self):
         return self.name
 
+    def delete(self):
+        if Recipe.objects.filter(steps__ingredients__unit__id=self.pk).exists():
+            raise ProtectedError(self.name + _(" is part of a recipe and cannot be deleted"), self.ingredient_set.all().exclude(step=None))
+        else:
+            return super().delete()
+
     class Meta:
         constraints = [
             models.UniqueConstraint(fields=['space', 'name'], name='u_unique_name_per_space')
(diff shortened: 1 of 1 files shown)


vabene1111:
Thanks for this fix, just to make sure

- orphaned steps are something that i want to allow in the long term (as i want to allow using steps in multiple recipes independently) so this change should not remove ingredients/foods/units that are part of a step that currently has no recipe ?

smilerz:
> Thanks for this fix, just to make sure
> 
> * orphaned steps are something that i want to allow in the long term (as i want to allow using steps in multiple recipes independently) so this change should not remove ingredients/foods/units that are part of a step that currently has no recipe ?

oh - yes, it does do that.  I had a bunch of food/units that were attached to orphaned steps with no way to manage that - which is what prompted the change.

vabene1111:
hmm what do you think about a management page to clear unused foods/units/keywords and also steps ? 

Could be integrated on the space page next to the statistics or something like that ?

vabene1111:
or we implement one of the standard split list views for steps with the future perspecive of adding a filter that allows to filter for "unused" items 🤔 could be something to be included in the StandardFilterMixin and a model function to find unused entries depending on the model

smilerz:
> hmm what do you think about a management page to clear unused foods/units/keywords and also steps ?
> 
> Could be integrated on the space page next to the statistics or something like that ?

IDK - mass deleting things makes me nervous.  But some way to manage steps/ingredients is necessary if enforcing ref-integrity on food/units

> or we implement one of the standard split list views for steps with the future perspecive of adding a filter that allows to filter for "unused" items 🤔 could be something to be included in the StandardFilterMixin and a model function to find unused entries depending on the model

There are definitely some 'standard' filters I've been thinking about for awhile. So this idea appeals.

> that would go hand in hand with the multi select option we talked about a while ago (i think, maybe i just thought about that feature ..)

I have a bulk edit feature for the model page mostly designed in my head.

vabene1111:
ok so bulk edit and some standardized way of filtering for the model list views it is! Does that make this PR irrelevant or shoudl i still merge it ?

smilerz:
Yes and no - if you don't want to delete food/units from orphaned steps it definitely needs changed.

vabene1111:
we could delete them for now but i fear that we might forget about this once we implement "step independence" and then delete stuf that should not be deleted. 

This also comes down to the question of when we are going to make steps independent, are we going to delete previously orphaned steps or keep them around ...

smilerz:
it would only matter if that step that should be 'saved' isn't part of any recipe.  But putting it on hold until further progress has been made makes sense.

RE: step independence - I can see some instructions being really useful.  I'd like to see embedding multiple pictures as very useful in that scenario (e.g. here are pictures how to butterfly a whole chicken)

vabene1111:
yes, some better way to integrate images both to steps and recipes is definitely needed. I added the "files" system as a quick workaround but it does not actually resovle the fundamental issues. 

I will take a look at this at some point hopefully soon.

PR #920, opened 2021-09-28
Sort tree enhancements

smilerz:
Allows disabling sort_tree in ENV variable (enabled by default)
Admin function to manually sort tree
Admin function to Enable/Disable tree sorting (will revert to ENV value on app restart)

At app startup run fix_tree - will resolve any problems found (includes sorting nodes if switching between sorted/unsorted)

> nodes with characters not found in the alphabet
> nodes when a wrong path length according to steplen
> orphaned nodes
> nodes with the wrong depth value for their path
> wrong number of children

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -128,4 +128,5 @@ REVERSE_PROXY_AUTH=0
 # however, when doing large imports of recipes that will create new objects, can increase total run time by 5-10x
 # Disabling SORT_TREE_BY_NAME (setting value to 0) will store objects unsorted, but will substantially increase speed of imports.
 # Keywords and Food can be manually sorted by name in Admin
+# This value can also be temporarily changed in Admin, it will revert the next time the application is started
 # SORT_TREE_BY_NAME=0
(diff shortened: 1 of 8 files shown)


vabene1111:
ok so i was just looking trough this, i did not know changing if sort is enabled or not was so easy. As far as i remember i only really noticed slowdowns when running imports and the imports are run by a central function catching (i think/hope) all exceptions and gracefully handling those. That means we could disable sorting once the import starts and enable it back after the import finished also executing the `fixTree` function am i correct with that assumption ?

Another question would be if there is any way to disable tree ordering for only one space or if it always has to be enabled/disabled for all spaces (which for the hosted version might be bad..). Maybe this could be achieved by not overriding the class attribute but one of trebeards internal functions 🤔 
```
Food.node_order_by = ['name']
Keyword.node_order_by = ['name']
```

smilerz:
>  As far as i remember i only really noticed slowdowns when running imports and the imports are run by a central function catching (i think/hope) all exceptions and gracefully handling those. That means we could disable sorting once the import starts and enable it back after the import finished also executing the `fixTree` function am i correct with that assumption ?

Mostly.  The one danger is if an in process write clashes with a write that starts after the switch,  but could be caught with a Try/Except 
 
> Another question would be if there is any way to disable tree ordering for only one space or if it always has to be enabled/disabled for all spaces (which for the hosted version might be bad..). Maybe this could be achieved by not overriding the class attribute but one of trebeards internal functions 🤔

The ordering is handled by the 'path' column, which means that it is Model specific.  It do per space you would have to have a per-space Food/Keyword class.  Though maybe a sorted/unsorted model is possible and switches between the two if necessary?

PR #952, opened 2021-10-11
Fix path error

tandy-1000:
Path should be bin/python3.9, errors out on debian otherwise.


vabene1111:
thanks for the fix, have you tested this in the docker image as well and can confirm its working there ?

vabene1111:
ok thanks, i will test if this works in docker as well just to make sure and get this merged asap

vabene1111:
i am sorry, i just noticed that this is in the documentation .. for some reason i tought that this was in some kind of script. Thanks for fixing this 👍

PR #960, opened 2021-10-16
LDAP support

lostlont:
Added LDAP authentication support.

`django-allauth` provides tons of authentication methods but LDAP seems to be not included in them, at least in recent versions.  `django-auth-ldap` can do that in turn, and it could be integrated quite easily.

I know LDAP may sound weird to have in a software made for personal usage and for user bases of a handful of people, but for those who use more than one service like this on one server, even for a smaller use base, it makes sense to handle accounts in a single place instead of handling them separately in all services. I made this change because I have this use case and it may be useful for others too. Tried it in a Docker instance of v0.16.8, then ported it to the latest develop branch and tried it in a local dev environment. I'm open to either doing changes or additional docs for this, or to rejecting the whole idea if it wouldn't fit the original concept.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -120,6 +120,7 @@ REVERSE_PROXY_AUTH=0
 # Django session cookie settings. Can be changed to allow a single django application to authenticate several applications
 # when running under the same database
 # SESSION_COOKIE_DOMAIN=.example.com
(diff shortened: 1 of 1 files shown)


vabene1111:
Thank you very much, i do see the use case for LDAP, i implemented it for some other applications i have as well. 

I will review the implementation and test if adding this introduces a lot of overhead in the final image and get this merged for hopefully the next release.

PR #1012, opened 2021-11-01
prepTime and workTime of CookbookApp not imported correctly

krial057:


vabene1111:
thanks, i must have missed this when copying from the yml format 👍

PR #1016, opened 2021-11-04
Add time unit hint to recipe edit view

StephenBrown2:
When editing the recipe for the first time, I was unsure of the time unit to use. Minutes makes the most sense, but I figure adding it in the label will help others in the future.


vabene1111:
thanks, that makes it more understandable

PR #1020, opened 2021-11-04
Add KubeSail install method

PastuDan:
Thanks for inviting me to add the KubeSail install to the docs!

Also, with PiBox our goal is to support open source devs, so after the Kickstarter is over and we're selling them on our website, we plan on donating a healthy portion of the sale if the traffic originates from the Tandoor docs or site (similar to how the amazon affiliate program works). We're still working out details, but if you have a Stripe account I think this should be pretty straightforward if you're interested. Will let you know once we get our shop set up!

I commented out the `A portion of every...` sentence until we actually get this in place :)

### DIFF
--- a/docs/install/kubesail.md
+++ b/docs/install/kubesail.md
@@ -1,13 +1,17 @@
 !!! info "Community Contributed"
     This guide was contributed by the community and is neither officially supported, nor updated or tested.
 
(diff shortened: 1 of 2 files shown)


vabene1111:
Awesome, i like the simplicity and idea of your product, feel free to contact me once its going, would love to have a chat with you one day.

PR #1038, opened 2021-11-06
Fix Templating documentation

rose-a:
Fix bad variable naming in docs


vabene1111:
interesting, i dont remember changing this ever but thanks for paying attention and fixing this issue :)

rose-a:
Well, I copy-pasted the template from the docs and got a very general error that there is an error with my template... so I went digging 😉

vabene1111:
Haha awesome, on tick is to copy the template reference from the ingredient context menu. In the future I hope to be able to add some kind of autoconpletion to the text editor and maybe even show live previews of the ingredients

PR #1068, opened 2021-11-13
Render separate ingredients flex row for each step 

### LINKED ISSUE #1067: Odd rendering of step names in ingredients list
### Version
Please provide your current version (can be found on the system page since v0.8.4)
Version:  **1.0.1**

### Bug description

Step headers are currently output as lone `<b>` tags within a `<table>` (in contrast to headline elements defined within the ingredient list, which are output as a table cell spanning all columns, as expected).

This results in unexpected word-wrapping (and IMO is also just "not as it should be"). 

grafik

grafik

rose-a:
closes #1067

Renders ingredients for each step in its own flex row with an optional title **before** the table.

Disclaimer: I never woked with vue before, so this might contain some glaring beginner mistakes 😉

### DIFF
--- a/vue/src/apps/RecipeView/RecipeView.vue
+++ b/vue/src/apps/RecipeView/RecipeView.vue
@@ -88,23 +88,21 @@
                 </div>
               </div>
               <br/>
-              <div class="row">
-                <div class="col-md-12">
-                  <table class="table table-sm">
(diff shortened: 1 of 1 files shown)


vabene1111:
thanks for the fix, i will test it tomorrow or so and get it merged asap

PR #1082, opened 2021-11-21
K8s install

TomHutter:
I created a bit more sophisticated setup for the k8s installation:

- The database is set up as a stateful set.
- The database container runs as a low privileged user.
- Database and application use secrets.
- The application also runs as a low privileged user.
- nginx runs as root but forks children with a low privileged user.
- There's an ingress rule to access the application from outside.

Documentation is also updated in `docs/install/kubernetes.md`.


vabene1111:
awesome, thank you 👍 i dont understand anything what you did there but it looks good so i am going to go ahead and merge this in the next few days :)

TomHutter:
In that case I have to thank you for the application in the first place and additionally for your confidence :-) I hope, I can be of help in case someone has issues with it.

PR #1104, opened 2021-12-01
Update synology.me

Nailik:
Added information to fix problems where container could not reach each other because firewall blocked it.
Added information how to setup ssl via reverse proxy.


Kaibu:
Thanks for the update!

PR #1150, opened 2021-12-20
punctuation fixes

auanasgheps:
First commit was done in rush. Fixed a couple of punctuations.


vabene1111:
thanks for the fix :)

PR #1211, opened 2022-01-03
Added Saffron and PDF export format

TiagoRascazzi:
Exporting to Saffron was added.

Exporting to a PDF uses pyppeteer to generate a PDF of the recipe page using print css.
This essentially allows to print multiple recipe to PDF.

Also in the recipe page the css was modified to avoid page break in the ingredient list and each individual steps


vabene1111:
this is very cool, thank you. We could also add a "print this recipe book" function using the pypeteer library. I need to take some time to properly review this but i like it very much, thanks !

cookbook/integration/integration.py
```diff
@@ -65,45 +65,32 @@ def do_export(self, recipes):
...
-
-                    recipe_zip_obj.close()
-                    export_zip_obj.writestr(str(r.pk) + '.zip', recipe_zip_stream.getvalue())
-
-            export_zip_obj.close()
-
-            response = HttpResponse(export_zip_stream.getvalue(), content_type='application/force-download')
-            response['Content-Disposition'] = 'attachment; filename="export.zip"'
-            return response
+        files = self.get_files_from_recipes(recipes, self.request.COOKIES)
```
vabene1111, line 71: i dont understand why the request cookie is passed, it isn't used anywhere or isn't it. I will check once merged and remove it if not used but maybe comment on this if it is not safe to remove the cookie parameter

vabene1111: ah i see its not used for all but for some of the exporters

vabene1111:
ok merging this, will need to test how this impacts performance if a browser is rendering every single recipe and someone (like me) who has hundres of recipes exports all of the at the same time but other than that this looks really cool

vabene1111:
Ok so i just merged and tested the changes, i have the following notes

- for puppeteer to work it needs to download a chromium bundle which is quite large (for some peoples VPS's). I could not really test performance yet (see below) but i suspect it might be meaningful on some servers. For that reason i think i will disable the PDF export by default (for now) and add an environment setting to allow it for server owners to decide if they want to enable it or not. 
- Exporting multiple recipes at the same time does not work due to a async issue. I honestly have not looked much into how djangos async works, i would have thought that if one recipe works exporting multiple works as well but it does not. Can you take a look at that ? If not i will just add to the docs that for now only one is possible (i only tested with manage.py runserver so far, is there any change in the setup required to get async to work properly) ?

TiagoRascazzi:
Ok I ll see If I can find something for exporting multiple

Issue #680, opened 2021-06-09
Shopping list does not merge ingredient if it has no unit


philip-n:
### Version
Version: 0.16.1

### Bug description
If multiple recipes are added to the shopping list and the same ingredient appears on different recipes (e.g. 100g flour per recipe), then the list automatically merges these and displays the sum (300g flour).

**Bug** This does not work for ingredients without unit, e.g. "tomato" or "onion".

Suggestion: If multiple of the same ingredient show up on a shopping list, and they have no unit, also combine them (--> 3 tomato instead of three times tomato).

vabene1111:
i think there was a technical shortcoming that created this behavior but i will definitely see that i fix this when reworking the page, thanks for the issue

Issue #1187, opened 2021-12-31
shopping list v2: finish all with one click


TWART016:
In Shopping list v2: It would be great if it's possible to finish all ingredients of a recipe / categories / Created_by with one click

vabene1111:
could be a context menu on the group header and check depending on what you group by 🤔

smilerz:
> could be a context menu on the group header and check depending on what you group by 🤔

I like that - you can finish a category all at once

smilerz:
Implemented in #1240

Issue #1185, opened 2021-12-31
map ingredients to supermarket


TWART016:
**Is your feature request related to a problem?**
Is it possible to add a Supermarket field in the ingredients entry? I normally buy an ingredient always in the same supermarket.

**Describe the solution you'd like**
When adding into shopping list v2, the ingredient is then automatically preselected (dropdown), but can be changed. The same applies to adding a recipe to the shopping list.

smilerz:
I don’t understand.  Can you try describing this differently?

TWART016:
If I edit a Food e.g. cheese, I want a field with supermarket. There I can select the supermarket.
shopping list:
Any in shopping list v2 group by supermarket would also be great

![image](https://user-images.githubusercontent.com/29805669/147803246-fc79108a-0ce2-4001-a95d-ae6dab23052d.png)

![image](https://user-images.githubusercontent.com/29805669/147803358-3fb30c9e-6031-4272-9b78-e0859d6b94b9.png)


![image](https://user-images.githubusercontent.com/29805669/147803121-51ac8477-e85b-4068-bd4a-1d782a0c5860.png)

![image](https://user-images.githubusercontent.com/29805669/147803282-fe979f2c-345e-455b-91da-105261f7355a.png)

vabene1111:
i dont really see why this should be done because basically everything can be bought in multiple stores but couldn't you just create seperate categories for each supermarket if you really want to do this ? e.g. Aldi - Meat, Aldi - Vegetables, .. ? 

Alternatively you could use (as it is intended by us) generic categories and link them to supermarkets so that you get all vegetables in aldi in correct order

TWART016:
Partly this can work, but I see it as more of a workaround.

With Aldi - Meat, Aldi - Vegetables the categories and the supermarket are mixed. For example, I buy steaks at Aldi, but minced meat somewhere else. Or olive oil only at this supermarket Z.
Categories are for me rather: vegetables, fruit, meat, salad, spices, cheese, dairy products.

Let's say there are 5 supermarkets and at least 7 categories. That would be already 35 entries.

For the Meal Planner it would also be better, because I select there: I would like to have vegetables and meat. There the supermarket is only a hindrance.

However, I may not have understood the concept behind it.

smilerz:
@TWART016 in that scenario you would have two supermarket categories: meat & minced meat.  
[...]

Issue #965, opened 2021-10-19
Shopping list problem when an ingredient is used in more than one recipe


0utl4w:
### Version
v0.17.2

### Bug description
When an item on the shopping list is the result of an igredient that is used in more than one recipe the values of the units are not added, they a displayed next to each other.

At first I thought there is a problem with the decimal seperators because I had numbers with more than one point. But those seperators didn't make sense at all. So I figured out that the first number was the value of the first recipe and the value from the second recipe is simply added right behind.,, I think a screenshot explains it better
![2021-10-19 15_37_39-Einkaufsliste - Brave](https://user-images.githubusercontent.com/80005700/137922543-f0964ffd-52b6-4343-91e1-c7df446d2b67.png)

In my example it should be 6 Knoblauchzehen (it displays 2.004.00) and so on.

vabene1111:
i added some decimal rounding in the last release, i fear this might be the cause of this issue. I think i will fix it for the old version as well so that smilerz isnt to pressured to get the new one out .. or maybe i should leave it broken 🤔 😂

vabene1111:
fixed in next release

PR #1240, opened 2022-01-05
Feature/shopping list v2

smilerz:
various fixes and enhancements


smilerz:
resolves #680, #490, #453, #710 #1187, #1185, #338, #490, #131, #965, #680, #941, #710, #453

PR #1262, opened 2022-01-09
Extension and hopefully simplification of bug reporting process

MaxJa4:
To get more information about bugs and prohibit having to ask one by one for specific information, I extended and redesigned the bug template.
Fell free to change parts or suggest changes.
Please note, that all explanatory parts are hidden as comments in the markdown (which the user will see when creating a bug ticket) so they don't unnecessarily clutter the finished bug report.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,9 +7,75 @@ assignees: ''
 
 ---
 
-### Version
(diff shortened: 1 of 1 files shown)


vabene1111:
this is cool, thanks. I saw on another repo that github improved the issue templates but haven't gotten around to adding one myself. I will need to look at what exactly we need to ask but thanks for laying out the foundation.

Issue #1275, opened 2022-01-10
readable receipt file names in default export


MarcusWolschon:
**Is your feature request related to a problem? Please describe.**
When exporting a single or multiple/all receipts, the individual zip files are currently just named 1.zip, 2,zip, 3.zip,...

**Describe the solution you'd like**
The name of the receipt with all illegal characters removed and a length limit should be used as the file name.
If this causes collisions, either _1.zip, _2.zip could be added as needed or all receipts get their id number as a postfix.

**Additional context**
This also allows storing some exported receipts on a shared Nextcloud or other file storage that multiple users share.
Each with their own receipt applications (not all of them Tandoor Receipts) or using the receipts manually. (e.g. by having an exported PDF in each receipt in addition to the machine readable version and the image file.)

MarcusWolschon:
Shell script to rename a directory of already exported 1.zip 2.zip .... into readable .zip file names that Tandoor will still recognize 


```
for file in *.zip ; do 
   mkdir temp ; 
   cd temp; 
   unzip ../$file ; 
   NEWNAME=$(cat recipe.json | sed -e "s#{\"name\":\"##" | sed -e "s#\".*##" | sed s/[ä]/ae/g | sed s/[ü]/ue/g | sed s/[ö]/oe/g | sed s/[^a-zA-Z0-9_]/_/g) ; 
   cd .. ; 
   rm -r temp ; 
   mv $file $NEWNAME.zip ; 
done

```

MarcusWolschon:
Proposal:

In [cookbook/integration/default.py line 59](https://github.com/TandoorRecipes/recipes/blob/93b868bc698cb096726f992687a72ece35739348/cookbook/integration/default.py#L59) use the sanitized receipt name

MarcusWolschon:
Issued a pull request with a fix:
https://github.com/TandoorRecipes/recipes/pull/1280

vabene1111:
i think i want to add a "slug" field to recipes which should be unqiue to a recipe. This would also allow to link recipes together after an import, that is currently not possible.

i will merge the PR for now as i think this is already an improvement but in the long run i will rewrite this to a "slug" system

PR #1280, opened 2022-01-11
#1275

MarcusWolschon:
#1275 readable receipt file names in default export

### DIFF
--- a/cookbook/integration/default.py
+++ b/cookbook/integration/default.py
@@ -2,6 +2,7 @@ import json
 from io import BytesIO, StringIO
 from re import match
 from zipfile import ZipFile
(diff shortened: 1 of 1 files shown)


vabene1111:
reverted this change as i broke when a database contained multiple recipes with the same name. As discussed on discord this needs a properly thought trough solution with unique slugs for recipes wich then can be re linked after import/export

MarcusWolschon:
multiple recipes with the same name worked just fine in a zip here.
All 3 zip clients could extract then selectively including MS Windows Explorer.
But whatever.

vabene1111:
how do you store two different files with the same name inside one folder/zip ? that simply does not work

PR #1281, opened 2022-01-11
PDF export improvement

TiagoRascazzi:
I was not able to recreate the await issue.
However I was able to find that when exporting large amount of recipes:
	The page would timeout: fixed by adding a loading page similar to import
	Pyppeteer goto() would timeout: fixed by creating the page in the recipe loop

To improve the speed, pyppeteer now wait for a selector instead of networkidle0

The export filename has the date appended to it; in YYYY-MM-DD format for it to be in sorted order in the users file directory

#1211


vabene1111:
this is a rather large PR and also change to the way exports work, i will need to properly review this once i am back home. 

Does this also add a loading page to all other exporters so that all of them profit from this workflow ?

TiagoRascazzi:
Yes the loading page is for all of the exporters

vabene1111:
as this is a little larger and requires some testing i will try to get the next version out and merge this then, hope you are ok with that

vabene1111:
merging this now onto a feature branch and fixing the issues myself, thanks for providing all the foundations to get this working.

TiagoRascazzi:
Ok, great.

vabene1111:
ok only thing left now is the step images/files but those are not supported by the other exporters as well so we will need to tackle that in its own go.

Issue #1093, opened 2021-11-30
Recipe link in plan


schwantuschke:
I would find it nice if there was a link to the recipe directly in the 3-point menu in the plan.
I can't get to the recipe from the entry, or have I overlooked that?

vabene1111:
you can open the entry and click on the recipe on the right side to open it but i think @Kaibu could just add a link to the context menu as well, thats probably a nice feature

schwantuschke:
Hello @vabene1111.
Here is a screenshot of an entry in the plan. Where do I get to the recipe there?
When I click on the recipe, I only get a selection of other recipes
<img width="839" alt="Bildschirmfoto 2021-12-01 um 14 08 02" src="https://user-images.githubusercontent.com/4090141/144240387-9e30bfd5-d45d-49a4-8036-0d9f484c0651.png">
<img width="833" alt="Bildschirmfoto 2021-12-01 um 14 10 57" src="https://user-images.githubusercontent.com/4090141/144240400-6c0ccc81-5d0d-487a-9fb9-a4d66cb5f346.png">
.

vabene1111:
It seems like one of the components is not loading correctly for you, that is very interesting (also i love rote linsen lasange :D)

@kaibu any idea how this can happen ? @schwantuschke can you post how you setup tandoor ? did you use docker or a manual setup and can you see if there are any errors shown in your browser console when visiting the meal plan ?

Kaibu:
Very weird indeed, you should have a Recipe card on the right.
It's still the only dynamically imported vue component in the app, which could be the issue.

Why that is the case - I'll have to investigate

Kaibu:
I'll still add a context menu entry to view the recipe

schwantuschke:
I have an account on app.tandoor.dev and use Firefox and Safari (on macOS and iPad OS).
Maybe this infos from Firefox console help you:
```Error: Promised response from onMessage listener went out of scope 4 background.js:841:170
Laden fehlgeschlagen für das <script> mit der Quelle "https://app.tandoor.dev/static/vue/manifest.7ffb0d32c26b.jsonvue/js/chunk-1b7f9ba9.js". plan:1:1
Fehler beim Laden von 'https://app.tandoor.dev/static/vue/manifest.7ffb0d32c26b.jsonvue/js/chunk-1b7f9ba9.js'. Ein ServiceWorker fing die Anfrage ab und es kam zu einem unwarteten Fehler. service-worker.js:1:12130
[...]

PR #1289, opened 2022-01-12
#1093 Recipe link in plan

MarcusWolschon:
#1093 add a recipe link into meal plan

I'm not experiences with Vue or Javascript development in general.
So please feel free to do any cleanup you see fit.

I coult not find a way to hide the context menu item if no receipt was linked to the meal plan entry.

### DIFF
--- a/vue/src/apps/MealPlanView/MealPlanView.vue
+++ b/vue/src/apps/MealPlanView/MealPlanView.vue
@@ -132,11 +132,11 @@
                 <ContextMenuItem
                     v-if="contextData.originalItem.entry.recipe != null"
                     @click="
-                    $refs.menu.close()
-                    recipe = contextData.originalItem.entry.recipe
-                    if (recipe != null) {
-                        navigateToURL(resolveDjangoUrl('view_recipe', recipe.id))
-                    }
+                        $refs.menu.close()
+                        recipe = contextData.originalItem.entry.recipe
+                        if (recipe != null) {
+                            openReceipt(recipe)
+                        }
                     ">
                     <a class="dropdown-item p-2" href="javascript:void(0)"><i class="fas fa-pizza-slice"></i> {{ $t("Recipe") }}</a>
                 </ContextMenuItem>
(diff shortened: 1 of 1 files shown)


smilerz:
> I coult not find a way to hide the context menu item if no receipt was linked to the meal plan entry.

Does this mean that `v-if` did not work?  or you didn't know the syntax?

MarcusWolschon:
> > I coult not find a way to hide the context menu item if no receipt was linked to the meal plan entry.
> 
> Does this mean that `v-if` did not work? or you didn't know the syntax?

I don't know the syntax and if it works within a context menu element.
This is the first time I've even seen Vue.

smilerz:
> I don't know the syntax and if it works within a context menu element. This is the first time I've even seen Vue.

Understood - I don't know if it will work there or not.  None of the contributors have worked with Vue before, so we are all learning as we go.  The [Vue documentation](https://vuejs.org/v2/api/) is pretty good, StackOverflow has lots of examples as does this project.
In general, you can add `v-if="object.attribute"` to any class and it will check for truthiness.

MarcusWolschon:
@smilerz done and tested

vabene1111:
i would say this does not fix the related issue as the basic issue is that for some reason the dynamically loaded component does not render for some users (while working for others). 
This is of cours a good workaround and generally good feature until @Kaibu gets around to investigate why the recipe card is not rendered.

vabene1111:
the code looks fine but why did you put most of the code into the html tag directly and made a function for the navigate to url ?

I think i would prefer if the code was in a function as well, something like open recipe. This is only a minor thing but i am traveling and cant quickly change this so it would be great if you could do that. Also please use `==!` for the comparison to avoid possible JS bullshit.

MarcusWolschon:
Because in the function it could not resolve the resolveDjangoUrl() function.

MarcusWolschon:
Noted.
Can't promise I'll have time to learn how to do this tomorrow.
The nutritional values import is more important for me.

vabene1111:
i am on vacation for the next few days anyway so no hurry.

MarcusWolschon:
Couldn't find out how to import that function.
Needs a JavaScript developer to clean that up.

vabene1111:
sorry for being so picky in this and some other PRs but if it works like this reviewing, merging and updating a PR like this is simply more work than doing the feature myself. 

you can simply use `this.resolveDjangoUrl("view_recipe", recipe.id)` when the function is imported as a mixin as seen in recipe switche line 96

MarcusWolschon:
Done

vabene1111:
this does not really do what i wanted (move most of the code from the JS @click inline code to the function but i will just implement this myself now to get this unblocked.

Issue #1304, opened 2022-01-13
Improve Issue Forms


vabene1111:
@MaxJa4 has already improvde the issue template but we can also migrate it to the new github issue forms system https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

The old forms basically need to be rewritten in the new templating system

MaxJa4:
That looks like a great and long needed feature of GH itself 👍🏻
I'm tempted to take a look.

smilerz:
as a future idea a template for "can't install tandoor" that links to various things that need checked first: nginx config, debug logs, headers, etc, etc, etc would be good too

MaxJa4:
Like a little FAQ "Before you report, see these FAQs". Sounds good

MaxJa4:
No worries.

Maybe a label like 'FAQ' would be good to label an issue for evaluation in that regard. That way, you don't need to worry about it right away @vabene1111

MaxJa4:
When creating a new issue inside the GitHub App (iOS), there is no issue template. That's probably because the template structure is still in beta and the app doesn't support it yet. Maybe enabling the old ones again helps (if GH automatically prefers the new format but can fall back to the old one; will test this)

MaxJa4:
Update: Having both enabled causes either conflicts (when they have the same name) and the old format overwrites the new one by default.
So the only two options I see are either reenabling the old templates and adding something like a "(old)" behind their names or just leave the GitHub-App without templates for now. @vabene1111

MaxJa4:
One downside of having checkboxes in a template (like in the feature request form): GH thinks there are tasks, so it says "1/3 tasks" in the issue overview... having a simple input field for writing Yes/No/Partly/... would solve that issue or it can also be removed entirely...

vabene1111:
I think I am fine with the app not working. Regarding the task lists I think that is OK to, we will see if it annoys to much we can still change it.

Thanks for all the testing and writing things up, really like the new templates.

PR #1309, opened 2022-01-13
Disabled old issue templates and added new ones with new GitHub issues format

MaxJa4:
Implemented new GitHub issues templates (#1304).

Feel free to adjust to your needs.


vabene1111:
awesome, thank you very much, i love this :+1:

vabene1111:
will merge it like this for now and if we find things we need to change update in the future we will.

PR #1317, opened 2022-01-14
Example for third-party authentication

mheiland:
Providing an example to integrate Keycloak as IAM for Tandoor. Hinting that both SOCIAL* variables are required.

### DIFF
--- a/docs/features/authentication.md
+++ b/docs/features/authentication.md
@@ -32,7 +32,7 @@ as environment files loaded by docker compose don't support multiple lines for a
 
 Take the example configuration from the allauth docs, fill in your settings and then inline the whole object 
 (you can use a service like [www.freeformatter.com](https://www.freeformatter.com/json-formatter.html) for formatting).
-Assign it to the `SOCIALACCOUNT_PROVIDERS` variable.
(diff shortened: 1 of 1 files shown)


Akruidenberg:
Woul love to get an example for OpenID connect with Authentik or Authelia.

mheiland:
> Woul love to get an example for OpenID connect with Authentik or Authelia.

That should be very similar to Keycloak, you need a Client-ID and a Secret Key, everything else is defined by the OpenID protocol or is an implementation detail of the IAM solution. Django Allauth does not contain a provider for those two though. Perhaps you can contribute a Allauth provider. Their implementation appears to be trivial, defining a few API paths and OpenID scopes: https://github.com/pennersr/django-allauth/tree/master/allauth/socialaccount/providers/keycloak

Akruidenberg:
Authentik is not using KEYCLOAK_REALM, so what do i need for this? Also, is KEYCLOAK_URL the same as the OpenID Configuration Issuer? Did some testing, but don't get it working at the moment.

vabene1111:
awesome, thanks for this PR. An example often helps people getting started

PR #1330, opened 2022-01-15
Makes reverse proxy port configurable

S1ructure:
Esspecially if you are running multiple containers on your machine this option might get important


vabene1111:
Thank you for your contribution but I don't really understand it. 

Why should this be handled in the env instead of just in the docker compose where everyone (or me at least) expects the port mappings to be configured?

S1ructure:
Thank you for the fast answer.

TL;DR;
.env file should be the only file which needs adaptions for running the project since it shouldnt be required making changes in infrastructure code
---

Actually the same reason why you specify all other environment variables there too - one place for all environment related configuration.

It makes it just harder for everyone running the project with the need to adapt more files than the .env file.

A docker setup should be configurable using the environment file, not the stack definition itself. Moreless the Open-Close-Principle on a Infrastructure as Code level.

Regards

vabene1111:
i understand your reason but i am not sure about this change since the nginx is technically not part of the application. I will need to think about this.

S1ructure:
agree about that - maybe a better naming would be anyway REVERSE_PROXY_PORT instead of  NGINX_PORT since the application doesnt care about which reverse proxy or if any.

PR #1343, opened 2022-01-17
Allow to specify an actual path using DATABASE_URL

MatthiasLohr:
Currently there is no possibility to move the location of `db.sqlite3` out of `/opt/recipes`. This pull request allows to specify full paths via `DATABASE_URL`, e.g.:

```
sqlite://nobody@nowhere/db/db.sqlite3
```

This will not access `/opt/recipes/db.sqlite` but `/opt/recipes/db/db.sqlite`.

### DIFF
--- a/recipes/settings.py
+++ b/recipes/settings.py
@@ -258,7 +258,7 @@ WSGI_APPLICATION = 'recipes.wsgi.application'
 # Load settings from env files
 if os.getenv('DATABASE_URL'):
     match = re.match(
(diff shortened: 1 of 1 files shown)


vabene1111:
thanks for the fix, i will include this in the next release

MatthiasLohr:
Great, thanks! Any idea when the next release will be released? This small feature blocks me in providing a much simpler way to deploy your great software to a Synology DSM NAS as described currently in the docs.

vabene1111:
i can push a beta update soon, this will likely be rather stable but there is no way to migrate backwards. Otherwise it will be about one or two weeks

MatthiasLohr:
That would be great! The other option would be a patch release with just this change cherry-picked.

vabene1111:
i know that is not what you want to hear but that simply is not justified for such a small tweak, i am sorry.

PR #1351, opened 2022-01-17
Add hint about trailing slashes for subpath setups

MaxJa4:
Add hint about trailing slashes for subpath setups due to recent issue on Discord.


MaxJa4:
GH seems to insist on putting a new line char at the end of the file when using the GH editor... well, shouldn't make a difference.

vabene1111:
thanks 👍

Issue #476, opened 2021-03-04
Import recipes with multiple instruction steps


smilerz:
**Describe the solution you'd like**
Allow URL import of multiple steps once hhursev/recipe-scrapers#283 is complete.

**Additional context**
Likely depends on #177 as well.

vabene1111:
the problem here is that steps are a very loosely defined thing on different websites. Some pages put every super small thing like "add water" "add salt" into separate steps while others use them the same way as i have set them up as more general parts of a recipe.

It would be cool to preserve those steps but i fear that it will be rather hard to make such a system work for all pages.
Nonetheless going to leave this open, definitely an area where the importer can be improved.

smilerz:
> the problem here is that steps are a very loosely defined thing on different websites. Some pages put every super small thing like "add water" "add salt" into separate steps while others use them the same way as i have set them up as more general parts of a recipe.

totally agree with that - I'm curious to see how the recipe_scraper addresses it.  My guess (hope) is that each scraper is adapted to how that site manages steps.

It probably isn't workable without a way to easily collapse steps in case it over generalizes.

vabene1111:
The scraper is really cool and i guess it handles the pages that are supported very well but i kinda hate the feeling that the lack of structure on those pages forces us to have hardcoded scrapers for individual sites 😂

smilerz:
> The scraper is really cool and i guess it handles the pages that are supported very well but i kinda hate the feeling that the lack of structure on those pages forces us to have hardcoded scrapers for individual sites 😂

I've seen enough "standards" that it doesn't even remotely surprise me anymore.

MarcusWolschon:
A way forward would be to already extend the preview-page that is shown after the scraper is done.
Add a label, that in the "steps" textarea everything separated by 2 newline will become a step.
That way the code is ready when the scaper outputs steps later
and the users can already do this now.
In a quick fashion when importing many receipts in a row. 
(e.g. new users moving their receipts over, who should get a good impression of Tandoor)

Limitations: All ingredients will be attached to the first step and steps will import without titles.
This limitation however is not likely to changes anyway.

vabene1111:
i want to rewrite the preview page or even better use the standard recipe editor to do that but thats a larger job. I will try to look at this somewhat soon

birdwing:
> Limitations: All ingredients will be attached to the first step and steps will import without titles. This limitation however is not likely to changes anyway.

[...]

PR #1389, opened 2022-01-21
#476 import recipes with multiple steps + cookidoo importer

MarcusWolschon:
Github doesn't update PRs with commits that are added to the branch after a PR is closed.
So here's a new one.
The Cookidoo importer is probably already pretty useful for many users without the changes to  cookbook/templates/url_import.html  and cookbook/views/data.py 
That's easy to split off.

If you still feel the need to do this together with a complex javascript editor for steps in the import html page
or to combine import and recipe-edit pages forst. then **I suggest a feature branch** instead of develop, where such a revamp can be done.

Working:

- User is instructed that manually added headers are converted into steps if and only if he/she chooses to via a checkbox.
- Only a new scraper for the Cookidoo site (with curated recipes that don't feature excessive use of steps) generates these automatically for the time being.

Improvements:

1. preview shows steps preformated
2. preview and preview-editing shows a text about the use of markdown
3. **added a checkbox during import** to enable the creation of steps similar to "all keywords"
4. text of the header is used as the name of the step
5. localised default step names for English and German
6. Added unit test case for the "cookidoo" scraper,
7. improved the translations
8. added the ability to work with complex receips that have grouped steps. (generating not only level 1 but also level 2,... headers in the instructions)
9. fixed support for cookidoo.at and cookidoo.ch in the cookidoo scraper
grafik


vabene1111:
sorry, as discussed i will close this. Feel free to merge this into your own build

PR #1399, opened 2022-01-22
Correct path to kubernetes files

geisterfurz007:
The docs mention an incorrect path to the kubernetes yaml files. This PR aims to correct that.


vabene1111:
thank you very much :+1:

Issue #1371, opened 2022-01-19
Refactor docker installation instructions


MaxJa4:
Since more and more instructions for different reverse proxies are added, scrolling up and down between the Docker-Compose section and the respective reverse proxy section is prone to error. There have been a lot of setup issues recently which is IMO partially because of a rather confusing layout of the [docker installation instructions](https://docs.tandoor.dev/install/docker/).

I would therefore suggest to refactor all these instructions into a better layout, maybe some extra information and hints about common issues (or references to those in the FAQ).

Additionally, it seems like ```{ % include "./docker/plain/docker-compose.yml" % }``` doesn't work in the docs to show a .yml directly from the repo... did that work in the past?

If you want, you can assign me to this (and add appropriate labels like documentation etc; thanks).

- [x] include troubleshooting step for raspberry pi ([see here](https://github.com/TandoorRecipes/recipes/issues/803#issuecomment-907748685))
- [x] include deleting local folders if any errors in postgres setup occurred

PR #1435, opened 2022-01-28
[Docs] Added example configs and header information

MaxJa4:
Added example configs for plain local nginx and Apache2.
Also added FAQ section for all generic issues when not setting all required headers correctly.
Added section for required headers in the Docker installation docs.

I hope this reduces the amount of "header-issues" which seem to be a constant issue-source.
As always, feel free to adjust.


MaxJa4:
The Apache configs were used/tested by their respective authors as stated in the text.
The nginx config was used by myself before I switched to Caddy (will be added soon).

If you want to be extra sure and test those configs again, feel free to do so.

MaxJa4:
Added additional content from #1371

vabene1111:
thank you for all your works on the docs, thats really saving me a lot of time for other things 👍

PR #1437, opened 2022-01-30
[Docker] Web_recipes waits for db_recipes to be ready

MaxJa4:
Fixes the setup issue almost all Raspberry Pi users have and probably some others with slow or low spec systems.
The web_recipes container waits for db_recipes to be healthy.
db_recipes is healthy, as soon as postgres is completely ready.

Note: ps_isready doesn't work here for some reason.

Tested multiple times with a Raspberry Pi 4.

### DIFF
--- a/docs/install/docker/nginx-proxy/docker-compose.yml
+++ b/docs/install/docker/nginx-proxy/docker-compose.yml
@@ -9,6 +9,11 @@ services:
       - ./.env
     networks:
       - default
+    healthcheck:
+      test: ["CMD-SHELL", "psql -U $$POSTGRES_USER -d $$POSTGRES_DB --list || exit 1"]
+      interval: 5s
+      timeout: 2s
+      retries: 5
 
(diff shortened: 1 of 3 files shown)


vabene1111:
awesome, thanks for fixing this 👍

MaxJa4:
Ran into an issue here. Docker-Compose 1.25.0 (like supplied by `apt install`) is not supporting that specific depends_on statement.
The current version of Docker-Compose is 1.29.2 (which is also the way that the official docker docs suggest). So it should be fine, can cause problems on old not up-to-date systems though... what do you think @vabene1111 ?

MaxJa4:
Tested -> RPI 3 with this version of docker-compose.yml works like a charm :)

vabene1111:
since only people with new installations will face this problems and existing users wont be impacted when updating this is totally fine (the docker compose version)

thanks for making this :)

PR #1439, opened 2022-01-30
Simple font size change

oliviapinson:
Made a small change to the font size so longer months won't affect arrow buttons. I think this helps the meal plan calendar look more aesthetic.
Here's an example: 
(before)
before
(after)
after


vabene1111:
thanks, that looks better 👍

PR #1445, opened 2022-01-31
Add new env parameter to set ldap ca cert file

thumm:
My ldap server uses TLS encryption with self signed certificates. With this PR it's possible to set the ca cert file used in ldap authentication to check the ldap server certificate via an optional environment variable. Details can be found [here](https://www.python-ldap.org/en/python-ldap-3.3.0/reference/ldap.html#ldap.OPT_X_TLS_CACERTFILE).


vabene1111:
awesome, thank you. Will get this merged for the next release

PR #1464, opened 2022-02-01
Fix Make_Header/Make_Ingredient text and reference

StephenBrown2:
When working with my recipes, I found that the text for "Make Header" and "Make Ingredient" still had an underscore in the words, so I fixed that up.
I then also found that the reference for Make_Header was duplicated in some places as Make_header (with a lowercase "h"), so this removes those references and makes all "Make_Header"s have a capital H.


vabene1111:
awesome, thanks for fixing this 👍

PR #1469, opened 2022-02-01
Keep screen awake when viewing recipe

### LINKED ISSUE #1434: Prevent switching display off after OS timeout

If you are cooking a recipe from your Tandoor recipe instance on a mobile device the display switches off after the timeout declared in the setting of your OS. Some apps, e.g. [Chefkoch](https://play.google.com/store/apps/details?id=de.pixelhouse&referrer=utm_source%3Dchefkoch.de%26utm_medium%3Dbanner%26utm_content%3Dmagazin_top%26utm_campaign%3Dchefkoch.generalcamp) keeps the display on.

Keep the display on

_No response_

### Additional context

Maybe i can implement this on my own. I am able to contribute python code, but never worked with django

### Contribute

- [X] Yes
- [X] Partly
- [ ] No

benscobie:
To resolve #1434

This is rather basic, at some point you could add a switch to toggle this functionality on & off.

### DIFF
--- a/vue/src/apps/RecipeView/RecipeView.vue
+++ b/vue/src/apps/RecipeView/RecipeView.vue
@@ -208,6 +208,7 @@ export default {
             servings_cache: {},
             start_time: "",
             share_uid: window.SHARE_UID,
+            wake_lock: null,
         }
     },
     watch: {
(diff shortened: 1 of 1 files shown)


vabene1111:
awesome, thanks for implementing this. I need to review this since i dont know much about this function but i am pretty sure it should work. Since i have my phone never turn off i never even noticed that this is a problem 😂

vabene1111:
ok just checked the vue lifecylcle docs, beforeUnmount seems to be a vue 3 lifecycle hook so i changed that to beforeDEstory, otherwise this seems correct, thanks for the contribution!

vabene1111:
ok it appears that this is indeed available in v2 as well but not in the docs .. ok staying with unmount then ..

ccoenen:
it should be noted that there is limited browser support for `navigator.wakeLock` ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/wakeLock), [caniuse](https://caniuse.com/mdn-api_wakelock)). Neither mobile firefox nor iOS Safari support this. So the only benefit is to Chrome users on android.

PR #1472, opened 2022-02-02
updated .dockerignore

smilerz:
ignore node_modules at every directory depth
ignore .vue and .ts files
ignore mediafiles/staticfiles/sqlitedb
ignore pytest configs/folders
ignore openapi generator

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,4 @@
-node_modules
+**/node_modules
 npm-debug.log
 Dockerfile*
 docker-compose*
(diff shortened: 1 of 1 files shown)


smilerz:
> Very nice, I'm excited to see the docker image size difference with this optimization :)

based on anecdotal evidence it got cut about 50%

vabene1111:
thats quite a lot ... but it makes sense, node modules is gigantic and it wasn't a problem before we moved the node build to the build process but when we did that i forgot to update .dockerignore

thanks :)

MaxJa4:
1.0.5: 181.54 MB / 743.5 MB (compressed / uncompressed)
1.0.6: 101.62 MB / 351.4 MB (compressed / uncompressed)
-> ~44% / ~53% size reduction. Pretty good guess smilerz :) Awesome!

PR #1484, opened 2022-02-04
Fixed JS parsing error on settings page

kristof-mattei:
The final `}` in this piece of code should've been included in the comment. 

image

This PR fixes it.

This fixes a spurious `}` preventing the whole code block from being loaded. `applyPresets` was broken because of this. 

Also formatted the JS to look alike.


vabene1111:
thanks for the fix :)

PR #1490, opened 2022-02-05
Added docker compose file for IPv6 native

murphy83:
This version makes use of a separate user defined network and allows to run tandoor in IPv6 native (if configured correctly, you will not need a reverse proxy like traefik/nginx etc.) Keep in mind to adjust the firewall rules accordingly to not expose services you do not want to be exposes (eg. django-server-container of tandoor).


vabene1111:
awesome, i thank you for contributing 👍

PR #1496, opened 2022-02-06
K8s deployment: Small fix for dupplicate entries.

poggenpower:
@TomHutter tagging you here, because you did the main work on this docs. 
Found some dupplicate declarations in the `40-sts-postgresql.yaml`, which should be fixed with this commit.


vabene1111:
i have zero knowledge about k8, will leave this open for a few days, if nobody objects i will merge this. 

Thanks @poggenpower for fixing this.

TomHutter (approved):
Yep, this can't be right and never could have worked. I was sure, I had the code verified, but apparently not.  Thanks for the fix.

PR #1498, opened 2022-02-06
Moved DB-waiting from docker-level to container-level

MaxJa4:
Undid the depends_on change in the docker-compose file, since it requires a fairly recent docker-compose version (>= 1.27.0) and can therefore cause issues in some cases.

In general, this should be handled by the requester - so the web-container - anyway.

`boot.sh` now retries every 5 seconds when the `python manage.py migrate` command exits with error code `1`.
This error code is thrown, when the DB (container) is not reachable yet (the exact error we're trying to tackle here). Also, when there is a misconfiguration like renamed db-container, where retries won't help but it doesn't hurt either - still worth it imo since it's an easy fix in the boot.sh like this.
Other errors I could provoke exited with error 130 so not all exit with a 1.
Success is obviously 0.

Tested with a Raspberry Pi 4 and on an Ubuntu 20.04 cloud server with reverse proxy.

### DIFF
--- a/boot.sh
+++ b/boot.sh
@@ -3,7 +3,7 @@ source venv/bin/activate
 
 echo "Updating database"
 while python manage.py migrate ; status=$? ; [ $status -eq 1 ]; do
-        echo "Migration failed! Database may not be ready yet, retrying in 5 seconds..."
+        echo "Migration failed (error #${status})! Database may not be ready yet, retrying in 5 seconds..."
         sleep 5
 done
 python manage.py collectstatic_js_reverse
--- a/boot.sh
+++ b/boot.sh
@@ -1,13 +1,33 @@
 #!/bin/sh
 source venv/bin/activate
 
-echo "Updating database"
-while python manage.py migrate ; status=$? ; [ $status -eq 1 ]; do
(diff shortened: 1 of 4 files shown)


vabene1111:
awesome, i think i will change the text to add "or misconfigured" or so but that is already very good. 

It can obviously hang in an infinite loop but with a non reachable DB that that would happen to some degree anyway.

MaxJa4:
That's a good point yes. I didn't want to bloat the log message but on the other hand, in this case, the more information, the better.

Yes, failing and "crashing" or retrying in a modest 5 second rhythm has the same result here in case the error is not caused by a slow booting DB container. Wouldn't cause any harm imo :)

MaxJa4:
An alternative would be to simply exit the boot script after a certain amount of retries, which would not start the gunicorn server.

This would theoretically be the safest way, since no infinite loop is happening for the migration if left unattended and the logs are clearer to analyze since there will be less spam after the container "gave up". Also partially working installations would be prohibited that way. 
What do you think?

vabene1111:
awesome, thats very cool, thank you for all the testing and work.

PR #1510, opened 2022-02-08
Fixes for English spelling and grammar

ZuluWhiskey:
I spotted a few spelling mistakes whilst using the product in English so have gone through and tried to change as many as I could spot - I've tried to leave any American spellings as-is (I'm from the UK), apologies if I've made any mistakes on that side of things.

I was looking to capitalise all acronyms such as `api` to `API` and `postgresql` to `PostgreSQL` but didn't as I wasn't sure if that would be acceptable or not, if it is then I can do another PR in the future.

As an aside to this PR, would the project be open to having English US **and** English GB/UK locales?


vabene1111:
thanks for providing those fixes. Sadly translating the english orignals in the msgid field of the localization file wont work, english strings need to be fixed either in the msgstr or directly in the source code locations as translation wont be applied otherwise. 

I am german and have learned english in school and online so my english is a mix of american/GB and all the source translations (the strings inside the code) are a mix of those two worlds. I guess we could try to differentiate those two versions but i am not sure if that is worth the effort ...

The thing is that english is the source language, i am not sure how to best approach this translation issue. Please give me a few days to think this trough and remind me if i forget it, i will show you how to best add translations then.

vabene1111:
closing this as it would break all translations. If someone has the time to feel free to change directly in the source, at some point we will need to decide if we perform the major undertaking of rewriting all backend translations so that you can edit them via translate.tandoor.dev which would also enable us to add a GB/US variant difference. 

For that we need to decide if we want to move to some "ID" based translation system

PR #1513, opened 2022-02-08
Ping DB instead of trying migration

MaxJa4:
In #1498 I added a retry functionality for the migration.
Since blindly retrying migrations can be dangerous since I'd see it as blindly messing with data, knowing something isn't right, I decided to look for a better solution (and also reduce log spam drastically).

With this, the PostgreSQL server in POSTGRES_HOST (e.g. db_recipes) will be pinged before migration.

The only downside to this is the necessity to add postgresql as additional dependency which increases image size slightly.
If someone knows another way to get access to the pg_isready command, please let me know.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM python:3.9-alpine3.12
 
 #Install all dependencies.
-RUN apk add --no-cache postgresql-libs gettext zlib libjpeg libwebp libxml2-dev libxslt-dev py-cryptography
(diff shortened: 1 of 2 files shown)


MaxJa4:
Installation size of postgresql: 32.11 MB
Installation size of postgresql-client: 2.07 MB
Sizes are uncompressed. So image should be only slightly bigger with the client variant, barely noticeable :)

Tested and worked flawlessly for me.

vabene1111:
awesome, thats really cool, thanks for making this better :)

PR #1516, opened 2022-02-08
Update de.json

hyperbit00:
typo in ressource


turnbeutelvergesser (commented):
The original text was correct, I think.  Zu "dem" Rezept umwandeln, hence "Zu interne_m_ Rezept".

hyperbit00:
> The original text was correct, I think. Zu "dem" Rezept umwandeln, hence "Zu interne_m_ Rezept".

Mhh, If I look at
Https://de.m.wiktionary.org/wiki/Flexion:intern
I See the Second table Dativ: "dem internen"

turnbeutelvergesser:
Yes, but then you'd need to use "Zu_m_ interne_n_ Rezept umwandeln" which does not sound natural, because you don't use the Dativ in that case.

vabene1111:
after accidentally merging please do not make PRs to master! 
also i agree that `m` is correct in this context.

i fixed Ressource manually

PR #1517, opened 2022-02-08
Update django.po

hyperbit00:
anpassung der Übersetzung und korrektur der ausgangsstrings


vabene1111:
manually changing the source strings leads to undesired behavior. When i update the translations every now and then by running the makemessages command it will update all changed source strings on its own. Also if you can please use the tranlate.tandoor.dev translation site, that makes it much easier to prevent merge conflicsts.

vabene1111:
updated by using makemessages to properly fix for all languages

Issue #1531, opened 2022-02-13
Connection Refused after install


reibuehl:
### Issue

I have installed TandoorRecipes with docker-compose. All three containers are up and running but when I try to access it via http://<FQDN of Docker Host>/, I only get a connection refused from the browser.
The onlything strange I could find in the logs is that the nginx container seems to have a weird timestamp. (see logs below)

### Tandoor Version

Can't login, but docker image is vabene1111/recipes:latest from 5 days ago

### OS Version

Debian 10.11

### Setup

Docker / Docker-Compose

### Reverse Proxy

No reverse proxy

### Environment file

```shell
# only set this to true when testing/debugging
# when unset: 1 (true) - dont unset this, just for development
DEBUG=0
SQL_DEBUG=0

# HTTP port to bind to
# TANDOOR_PORT=8080

# hosts the application can run under e.g. recipes.mydomain.com,cooking.mydomain.com,...
ALLOWED_HOSTS=*

# random secret key, use for example `base64 /dev/urandom | head -c50` to generate one
# ---------------------------- REQUIRED -------------------------
SECRET_KEY=xxxx
# ---------------------------------------------------------------

# your default timezone See https://timezonedb.com/time-zones for a list of timezones
TIMEZONE=Europe/Berlin

# add only a database password if you want to run with the default postgres, otherwise change settings accordingly
DB_ENGINE=django.db.backends.postgresql
# DB_OPTIONS= {} # e.g. {"sslmode":"require"} to enable ssl
POSTGRES_HOST=db_recipes
POSTGRES_PORT=5432
POSTGRES_USER=xxxx
# ---------------------------- REQUIRED -------------------------
POSTGRES_PASSWORD=xxxx
# ---------------------------------------------------------------
POSTGRES_DB=djangodb

# database connection string, when used overrides other database settings.
# format might vary depending on backend
# DATABASE_URL = engine://username:password@host:port/dbname

# the default value for the user preference 'fractions' (enable/disable fraction support)
# default: disabled=0
FRACTION_PREF_DEFAULT=0

# the default value for the user preference 'comments' (enable/disable commenting system)
# default comments enabled=1
COMMENT_PREF_DEFAULT=1

# Users can set a amount of time after which the shopping list is refreshed when they are in viewing mode
# This is the minimum interval users can set. Setting this to low will allow users to refresh very frequently which
# might cause high load on the server. (Technically they can obviously refresh as often as they want with their own scripts)
[...]

PR #1535, opened 2022-02-14
Added existence check for Nginx config file and mandatory parameters

MaxJa4:
Checks on Web-Container startup, whether the Nginx configuration file (Recipes.conf) is present.
If it's not present - which would prohibit Nginx from working properly (see #1531) - a warning is shown.

If GUNICORN_MEDIA is not 0 (0 is default) and therefore Nginx is not used, no warning will be shown.

Path to config file is hardcoded, but there is no point in changing the mounting point in the docker-compose file anyway and the warning doesn't prohibit the startup - it's just to make issue analysis for the users and developers easier.

### DIFF
--- a/boot.sh
+++ b/boot.sh
@@ -6,7 +6,7 @@ NGINX_CONF_FILE=/opt/recipes/nginx/conf.d/Recipes.conf
 
 echo "Checking configuration..."
 
-if [ ! -f "$NGINX_CONF_FILE" ] && [ $GUNICORN_MEDIA -ne 1 ]; then
+if [ ! -f "$NGINX_CONF_FILE" ] && [ $GUNICORN_MEDIA -eq 0 ]; then
     echo -e "\n[WARNING]\nNginx configuration file could not be found at the default location!"
     echo -e "Path: ${NGINX_CONF_FILE}\n"
 fi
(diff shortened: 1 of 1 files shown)


MaxJa4:
I would also add a check for the existence of the security key and postgres-pw since the are required...
Since we recently had a discussion about that in Discord, I'm waiting for your OK on that.
Would be a temporary solution ofc, until these checks are done in the UI or the .env is restructured :)

vabene1111:
i think a warning on startup, no matter on what solution we choose to implement for variables/config, is always good if a required setting is not there and helps debugging, so go ahead and thank you very much for all these little but very impactful tweaks and changes you are making 👍

vabene1111:
i will leave this open until you have the variables done so you dont need two PRs

MaxJa4:
Done.

Worst case (no nginx config found + both variables not set) looks like this:
```
Checking configuration...
[WARNING]
Nginx configuration file could not be found at the default location!
Path: /opt/recipes/nginx/conf.d/Recipes.conf
[WARNING]
The environment variable 'SECRET_KEY' is not set but REQUIRED for running Tandoor!
[WARNING]
The environment variable 'POSTGRES_PASSWORD' is not set but REQUIRED for running Tandoor!
Waiting for database to be ready...
Database is ready
Migrating database
[...]
```

vabene1111:
awesome, thank you 👍

PR #1554, opened 2022-02-18
[Docs] Add syntax highlighting for nginx subpath example

MaxJa4:


vabene1111:
awesome, thanks, i totally forgot that

PR #1564, opened 2022-02-21
retain original text when parsing ingredients

smilerz:


vabene1111:
this is to be merged after i am done with the import/export branch correct ?

smilerz:
> this is to be merged after i am done with the import/export branch correct?

it's a pretty small change - so up to you.  the only real change at the front end is a rename of original with original_text and making sure that original_text is always submitted with ingredients.

Issue #1579, opened 2022-02-24
Some translations are not applied (new shopping list, GenericModals for models)


MaxJa4:
### Tandoor Version

1.0.8

### Setup

Docker / Docker-Compose

### Reverse Proxy

Caddy

### Other

Also happens on local dev setup.

### Bug description

- The new shopping list is completely untranslated even though it should be regarding the vue src and the *.json locale files
- Modal for creating new keywords/foods etc (general modal) are not translated but data in *.jsons is there

smilerz:
@MaxJa4 - can you provide a specific example for the GenericModal?

MaxJa4:
/list/keyword/ for example, it's partially being translated there.
/list/food/ is completely untranslated (but strings are there in the jsons).

smilerz:
Just to confirm - when you edit a food the items below are all in English despite the translations `"OnHand"` and `"OnHand_help"` being present and correctly translated in de.json?
![image](https://user-images.githubusercontent.com/1808639/155748586-9e01bd5e-9455-4e39-972b-6f23cde68bac.png)

MaxJa4:
V1.0.8: (everything not translated except header and "cancel" button
![image](https://user-images.githubusercontent.com/74194322/155750771-f15772c3-8aae-432c-be26-38182e7c7731.png)

Develop: (same as above, same with the advanced fields)
![image](https://user-images.githubusercontent.com/74194322/155750861-30324a26-3309-4b8b-8291-a5ff084cf78b.png)

PR #1586, opened 2022-02-25
fix translations in helper libraries models.js and utils.js

smilerz:


smilerz:
fixes #1579

vue/src/apps/ShoppingListView/ShoppingListView.vue
```diff
@@ -934,6 +934,8 @@ export default {
                 this.ui = Object.assign({}, this.ui, this.$cookies.get(SETTINGS_COOKIE_NAME))
             }
         })
+        this.$i18n.locale = window.CUSTOM_LOCALE
+        console.log(window.CUSTOM_LOCALE)
```
MaxJa4, line 938: Will this be in the container logs or just when developing locally?

smilerz: I’m bad about taking out console messages from debugging. That should be removed. But it would only be in the browser console.

MaxJa4: Haha just looked at the code bc I was curious myself :)

MaxJa4: Thanks for fixing the issue

smilerz: No problem. The root cause was the helper library not translating. Not sure if it was compile issue or not having access to the locale.

PR #1591, opened 2022-02-27
Optimized advanced search settings UX

MaxJa4:
Small changes to render some search related controls better, especially in languages with longer words (German here) and correcting the vertical alignment between label and control in all languages.

## Advanced search settings (EN) - _Before | After_
image image
## Advanced search settings (DE) - _Before | After_
image image

## Filter logic gates (EN) - _Before | After_
image image
## Filter logic gates (DE) - _Before | After_
image image


vabene1111:
awesome, i did not have time to do this yet, this is much better, thanks 👍

PR #1611, opened 2022-03-01
missing prepTime assignment

smilerz:


vabene1111:
i think i already fixed this on the importer rewrok branch but thanks for fixing it in develop as well, will leave this open depending on how much progress i make today with the importer

PR #1620, opened 2022-03-03
Update boot.sh

ratheile:
Hi,

I added a small fix to the boot.sh file:
 - Add port to pg_isready to allow non-standard ports.


MaxJa4 (approved):
Good addition.

ratheile:
Thanks for merging!

PR #1623, opened 2022-03-03
Improve recipe import for non schema.org recipes

gloriousDan:
When testing the import with version 13.19.0 of recipe-scraper, I realized that we do not use all information the scraper object may provide.

Especially when a recipe was scraped that didn't support Schema.org, scrape_me provides fields only through the respective scraper methods and not through its `schema.data` attribute.

Added support for
* `scrape.description()`
* `scrape.prep_time()`
* `scrape.cook_time()`
* Keywords from `scrape.cuisine()` and `scrape.category()`
If these properties return an exception the fallback is to use the old behaviour and try to get the data from the schema object directly.

getting the URL from `scrape.canonical_url()` instead of from `scrape.url` also can't screw anything up since in the recipe_scraper library the default implementation of `canonical_url` defaults to returning `scrape.url` anyway.

### DIFF
--- a/cookbook/helper/recipe_url_import.py
+++ b/cookbook/helper/recipe_url_import.py
@@ -28,9 +28,14 @@ def get_from_scraper(scrape, request):
             recipe_json['name'] = ''
 
     try:
-        description = scrape.schema.data.get("description") or ''
+        description = scrape.description()  or None
     except Exception:
-        description = ''
+        description = None
+    if not description:
+        try:
+            description = scrape.schema.data.get("description") or ''
+        except Exception:
+            description = ''
 
     recipe_json['description'] = parse_description(description)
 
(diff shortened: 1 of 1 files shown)


vabene1111:
aweseom, thank you for looking this all up and implementing this. I will likely merge it directly into my branch where i rework the URL importer frontend anyway

gloriousDan:
My pleasure. 
It would be great to have this merged in the next release (or whenever recipe-scraper is updated to 13.19.0) since without the change to `scrape.canonical_url()` recipes can not be imported from kptncook. 
I'm not sure what your timeline on the import/ export refactor is but if it takes longer it would be great to at least merge the `canonical_url()` line earlier

vabene1111:
ok, good to know. i was planning on working on the import export rework today but it will likely have a few weeks of testing ahead of it so maybe i will merge this into develop first and then update my feature branch.

gloriousDan:
That would be great, otherwise I could also live with it not working for a couple weeks.

PR #1640, opened 2022-03-08
Fixed typos and grammatical issues.

CameronJGrant:
Gunicoren should be Gunicorn
Send should be Sent. (Also reworded more clearly)


vabene1111:
thank you very much !

PR #1653, opened 2022-03-17
Fixes the formatting

bradpoulton:
I added too many ! and it did not format properly.


vabene1111:
thanks for the fix, did not even notice :see_no_evil:

bradpoulton:
> thanks for the fix, did not even notice 🙈

Thanks for the awesome app!

PR #1698, opened 2022-04-04
fix links in german locale

HarHarLinks:
idk if these need to be fixed on weblate or what, but I can't log into your weblate with github and I don't care enough to create an account for something that ends up here anyway.

before this change, these links lead to
image


vabene1111:
thanks for the fix 👍

PR #1709, opened 2022-04-09
Docs: Reintroduce restart-always to web_recipes service

gloriousDan:
Following a support request and discussion in discord this PR aims to reintroduce the `restart: always` policy for the tandoor web-service container in the standard docker-compose.yml.

It is unexpected for new admins or users unfamiliar with docker-compose that all containers except for tandoor-web restart after a reboot.

Restart loops were the reason why `restart: always` was removed in https://github.com/TandoorRecipes/recipes/pull/1498
I think it's acceptable behaviour that such restart loops happen when the configuration is faulty and thus the container crashes.

### DIFF
--- a/docs/install/docker/nginx-proxy/docker-compose.yml
+++ b/docs/install/docker/nginx-proxy/docker-compose.yml
@@ -11,6 +11,7 @@ services:
       - default
 
   web_recipes:
+    restart: always
     image: vabene1111/recipes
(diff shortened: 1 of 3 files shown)


vabene1111:
Thanks for the pr. I will need to think about this but I see your point. Also some other services I use tend to behave like this as well.

PR #1710, opened 2022-04-09
Navbar tweaks

hendrikbl:
The navbar logo had a height of `5vh` wich ich relative to the screen size and results in a very big logo on bigger screens like a 1440p screen (see first image). This PR does two things to make the logo look better on multiple screens.

- Use a fixed height, in this case `40px` for every screen size.
- Use the already existing `svg` image instead of the `png` wich can get blurry depending on the scaling.

Additionally I swapped the logo and menu button on mobile screens. Since there is no Navigation Drawer as known from Material design, there is no point in moving the Logo to the right on mobile. It looks more consistent if the logo stays at the same place for all screen sizes.

I'd like to hear what you think about these changes.

### 1440p screen
old:
drawing
new:
drawing

### Mobile screen
old:
drawing
new:
drawing

### DIFF
--- a/cookbook/static/css/app.min.css
+++ b/cookbook/static/css/app.min.css
@@ -1,3 +1,7 @@
+.brand-icon {
+    height: 40px;
+}
+
 .spinner-tandoor {
     animation: rotation 3s infinite linear;
     content: url("../assets/spinner.svg");
--- a/cookbook/templates/base.html
+++ b/cookbook/templates/base.html
@@ -60,17 +60,19 @@
 <body>
 
 <nav class="navbar navbar-expand-lg navbar-dark bg-{% nav_color request %} bg-header" id="id_main_nav"
-     style="{% sticky_nav request %}">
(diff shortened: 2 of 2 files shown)


vabene1111:
Thank you, I like those changes. Maybe we should also support the left hand setting for the nav expander bit this can be done in a separate task.

vabene1111:
@Kaibu you added the navbar, do you think this is a good idea or was there a reason you used relative height ?

Kaibu:
I like this change, I'll add back the left taskbar as an option for left-handed people

vabene1111:
ok so merging this for now, i think we will have to properly clean up/sort the settings at some point anyway, there we can also rename the setting to a global left/right hand mode.

Kaibu:
I also tweaked the recipe switcher to work with this change and added a left-handed mode that works vise-versa
I took the opportunity to rezise the action buttons aswell, they were huge on mobile

Thanks for your contribution!

Before
image
After
image

PR #1714, opened 2022-04-10
Docs: nginx mount volumes read-only

gloriousDan:
Nginx should only read and serve all files from the following volumes and doesn't need write access:
* `staticfiles` 
* `mediafiles` 
* the nginx config
 
I wondered why the read-only flag was only set for the nginx config volume.
I tested setting read-only for staticfiles and mediafiles aswell on my setup and didn't have any issues.


vabene1111:
This makes sense, even tough the security impact is likely very small. Will test this as well and get it merged hopefully soon.

PR #1719, opened 2022-04-12
Remove namespace vom k8s templates

tristanlins:
The namespace prevents installing into a custom namespace. 
Also, "default" is the default namespace and does not need to be specified.


vabene1111:
Thank you, I have no clue about k8s but this sounds reasonable. I will get it merged with tje next release.

PR #1723, opened 2022-04-14
Fix some issues to improve the print output

vekunz:
I fixed some issues with the print output that some elements were not hidden in print mode.

---

In addition to the issues, I fixed there is another issue, I could not fix. In print mode, the border around the portions selector should disappear. Currently, this is not possible, because in tandoor.min.css on line 10444 is this style, and the `!important` makes it impossible, to override this from inside a component (because the CSS selector is very strong):
```css
textarea, input:not([type="submit"]):not([class="multiselect__input"]):not([class="select2-search__field"]):not([class="vue-treeselect__input"]), select  {
    background-color: white !important;
    border-radius: .25rem !important;
    border: 1px solid #ced4da !important;
}
```
Since I do not know what the overall effect of removing the `!important` is, I left it there.

---

This is my first contribution and I hope that I have done it properly.

### DIFF
--- a/cookbook/static/themes/tandoor.min.css
+++ b/cookbook/static/themes/tandoor.min.css
@@ -9548,7 +9548,8 @@ a.text-dark:focus, a.text-dark:hover {
 @media print {
     *, :after, :before {
         text-shadow: none !important;
-        box-shadow: none !important
+        box-shadow: none !important;
+        background-color: transparent !important;
     }
 
     a:not(.btn) {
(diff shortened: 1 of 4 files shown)


vabene1111:
Thank you very much for the contribution and well structured comments.

@Kaibu do you know if the important is required?

Ashish8684:
Nice sir

Kaibu:
Looking into it

vabene1111:
will get this merged for now and open an issue for the other point you mentioned for me to review later

Issue #1631, opened 2022-03-06
[suggestion] Youtube import?


adrianmihalko:
### Describe the solution you'd like

It would be great to have a feature to import recipes from Youtube? 

My idea:

- import video's thumbnail as recipe image 
- import description of the video (many times the ingredients and instructions are listed here)
- add link to the youtube video to description

I follow many BBQ channels on Youtube and I want to add recipes from there. It often happens that I like a recipe on the video, but I don't know where to save it. If we could just save a picture and a link to the video to the recipes, that would be great.

vabene1111:
that makes sense, as youtube is a large platform we could justify writing a special case for the improter for this. i like the idea, thanks 👍

vabene1111:
maybe you can even load the embed url of the video and embed it in markdown, not sure if that is within spec

tomtjes:
You could also try and grab the subtitles and add them to the recipe, just in case the video description doesn't contain the cooking instructions.

parkroland:
Hi all, I think I'm going to give this one a shot.

PR #1732, opened 2022-04-18
Add YouTube Import Functionality

parkroland:
This PR is regarding issue [#1631](https://github.com/TandoorRecipes/recipes/issues/1631) asking for the ability to import YouTube videos. 

I changed the `recipe_from_source` API to utilize the [pytube](https://pytube.io/en/latest/) library to get information from a YouTube URL. I used regular expressions to distinguish YouTube urls from the other import methods and some defensive coding to prevent users from importing invalid/expired YouTube links.

Additionally I added a small unit test to test this functionality.

Here is a link to a [GIF](https://drive.google.com/file/d/1wQKyElrz6Gj6Xs6K57h0V9ea3AsN3kjQ/view?usp=sharing) that displays the user experience that can be expected from this change.

One thing to note is that YouTube video descriptions oftentimes exceeds the current description limit for recipes. I assume that a user would like to edit this themselves so I did not address this common issue.

Let me know if there is anything else I can do to make this better :)

### DIFF
--- a/cookbook/helper/recipe_url_import.py
+++ b/cookbook/helper/recipe_url_import.py
@@ -2,6 +2,7 @@ import random
 import re
 from html import unescape
 from unicodedata import decomposition
+from pytube import YouTube
 
 from django.utils.dateparse import parse_duration
 from django.utils.translation import gettext as _
@@ -168,6 +169,33 @@ def get_from_scraper(scrape, request):
     if scrape.canonical_url():
         recipe_json['url'] = scrape.canonical_url()
         recipe_json['recipeInstructions'] += "\n\n" + _("Imported from") + ": " + scrape.canonical_url()
+    return recipe_json
+
+
+def get_from_youtube_scraper(url):
+    """A YouTube Information Scraper."""
+    default_recipe_json = {
+        'name': '',
+        'description': '',
+        'servings': 1,
+        'prepTime': 0,
+        'cookTime': 0,
+        'image': "",
+        'keywords': [],
+        'recipeIngredient': [],
+        'recipeInstructions': "",
+    }
+
+    recipe_json = default_recipe_json
+    try:
+        video = YouTube(url=url)
+        recipe_json['name'] = video.title
+        recipe_json['image'] = video.thumbnail_url
+        recipe_json['description'] = video.description
+        recipe_json['description'] += "\n\n" + _("Imported from") + ": " + url
+    except Exception:
+        recipe_json = default_recipe_json
+
     return recipe_json
 
 
--- a/cookbook/tests/other/test_url_import.py
+++ b/cookbook/tests/other/test_url_import.py
@@ -5,11 +5,13 @@ import pytest
 
 from django.urls import reverse
 
+from cookbook.helper.recipe_url_import import get_from_youtube_scraper
+
+from cookbook.tests.conftest import validate_recipe
 from ._recipes import (
     ALLRECIPES, AMERICAS_TEST_KITCHEN, CHEF_KOCH, CHEF_KOCH2, COOKPAD,
     COOKS_COUNTRY, DELISH, FOOD_NETWORK, GIALLOZAFFERANO, JOURNAL_DES_FEMMES,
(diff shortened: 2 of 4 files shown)


parkroland:
Not sure how to address the security vulnerability as it is flagging code that was there prior to this change.

vabene1111:
this is a very cool and clean way of implementing the youtube functionality 

following notes
- the security alert can be ignored, GitHub somehow bugged out a bit i will need to review this
- i would put the text not in the description but the instructions, this way you dont have an issue with the length limit
- what do you think about adding the video as a clickable image `[!Video Title` with somthing like 
- the data structure of the importer process changes with the next update but is currently on a feature branch, i will make the required changed when merging, just dont be confused when everythign looks a little different after being merged

parkroland:
Thanks for responding! I agree with your suggestions I'll be sure to fix it to put the description in the instructions and have the video be come clickable.

parkroland:
Quick question: I am trying to add the video as a clickable image. However, it looks like this change would require a brief change on the frontend to add an additional anchor tag with the video url according to [this](https://www.tutorialspoint.com/How-to-use-an-image-as-a-link-in-HTML) article. I am trying to add an optional "youtube_url" parameter to the recipe_from_source api in `api.py` but I am having some trouble as I can't find the appropriate files to make additional changes to get the data to the frontend. Any chance you could point me in the right direction?

vabene1111:
Hi, sorry for not making this clearer in my first post but `[!<Video Title>` just works if you replace the relevant parts in `<>` with the appropriate data. You can use anchor tags as you describe above but that is more work and would require changes to the cleaning/filtering functions.

I tested the above in my instance and it worked fine, if it doesn't in your setup let me know so we can investigate

parkroland:
Do you have any links to any online articles explaining what you suggested? Unfortunately adding that change is not working for me.

My code is pretty straightforward, the line is: `'[]({2})'.format(video.title, video.thumbnail_url, url)`
which produces in my local testing a string like: `[!How to Make Italian Bruschetta - Recipe in description](https://www.youtube.com/watch?v=8yPTW1RTFb4)`

No image then appears on my local setup and I receive a 404 error on Django saying it cannot find the url. Thanks!

vabene1111:
your code is very close but not completely correct compared to the example above, it needs to be 

```python
'[![{0}]({1})]({2})'.format(video.title, video.thumbnail_url, url)
# produces
[![How to Make Italian Bruschetta - Recipe in description](https://i.ytimg.com/vi/8yPTW1RTFb4/sddefault.jpg)](https://www.youtube.com/watch?v=8yPTW1RTFb4)

```

why is this this way? in markdown the syntax for inserting an image is very similar to the one for inserting a link, we are using both basically putting and image inside a link to make the image clickable:

```
![Alt Text}(link)   --> produces an image with alt "Alt Text" and src= "link"
[Link Text](link)  --> produces an anchor tag with text "link text" and href="link"

[ 
    ![Video Title](thumbnail url)
] ( video url) -> produces an anchortag containing an image tag with a.href=video url, img.alt=video title and img.src=thumbnail url 

```
hope this makes it a little clearer

parkroland:
Thanks for the clarification! I really appreciate it. Unfortunately I am still running into this issue.

This is the error I am running into in my local django server and the experience I am running into on the frontend:
```
[21/Apr/2022 01:38:48] "POST /api/recipe-from-source/ HTTP/1.1" 200 1708
Not Found: /data/import/[![How to Make Italian Bruschetta - Recipe in description](https://i.ytimg.com/vi/8yPTW1RTFb4/sddefault.jpg)](https://www.youtube.com/watch
[21/Apr/2022 01:38:48] "GET /data/import/[![How%20to%20Make%20Italian%20Bruschetta%20-%20Recipe%20in%20description](https://i.ytimg.com/vi/8yPTW1RTFb4/sddefault.jpg)](https://www.youtube.com/watch?v=8yPTW1RTFb4) HTTP/1.1" 404 22542
```
Screen Shot 2022-04-20 at 7 40 25 PM

I will be looking into it and keeping you updated :)

vabene1111:
when does this issue appear ? it looks like you are putting the markdown link into the URL ?

PR #1734, opened 2022-04-21
Fix link in German locale

Drumstickx:
This fixes a link in the German locale, similar to #1698. I guess the proper way would be to fix this in Weblate but I have no access there. So feel free to merge this PR or apply the changes in Weblate, whatever is easier for you.


vabene1111:
you can simply create an account on weblate and start over there but this PR works as well so thank you very much, i will get this  merged

Issue #1784, opened 2022-05-07
PlanToEat Import Fails (0 of 0 recipes were imported)


Ruthalas:
### Tandoor Version

1.2.3

### Setup

Unraid

### Reverse Proxy

No reverse proxy

### Bug description

Importing from a PlanToEat export fails with "0 of 0 recipes were imported".

I've tried both the TXT and CSV style exports PlanToEat provides.

A previous issue that may or may not be related is Issue 1031.

vabene1111:
can you provide some sample data either here or via email to info@tandoor.dev so that i can review whats going wrong here ? You need to use the txt format

Ruthalas:
Can do!
I have submitted a sample export to the provided email.
The title of the email is "(Issue#1784) Sample PlanToEat Export Data".

Ruthalas:
Thank you very much for the help.
Just wanted to note that I've discovered two issues with the importer.
It seems to have an issue parsing the tags portion of the export, combining them like so:
![image](https://user-images.githubusercontent.com/27835519/168410907-64903769-4328-430c-b740-a9b6f40a1e33.png)

I have submitted Pull #1804 to resolve that issue.

It also struggles with some ASCII codes, as can be seen here:
![image](https://user-images.githubusercontent.com/27835519/168410956-91903332-17f0-4402-aaf9-5122d8f5c9bd.png)

I don't know how to fix the ASCII issue, so that is not included in the pull request noted above.

PR #1804, opened 2022-05-14
Update plantoeat.py to correct Tag delimiter

Ruthalas:
Changed tag delimiter that the PlanToEat importer uses from "," to "^" to reflect a current export's formatting.

Relates to Issue #1784


vabene1111:
i think we will likely need to split at both `,` and `^` but i will need to look trough the data to see if both are used.

PR #1829, opened 2022-05-29
Some minor visual fixes

astappiev:
Hi all,

First of all, thank you for your work. I'm glad I'm found your project!
Last months I started creating weekly menu and follow them. I have tried a few android applications, but all of them missing some part that I need. And I though maybe I need to create a new one with all functionality I need. But sounds like a terrible idea (in terms of time).

I started using the product today, first I can say the paid subscription is unreasonably expensive. I have tried other applications, and they are cheaper with some more work in some directions. But you have a great benefit - you are open source. And in return to my "free" use, I will suggest some changes which I will do for my needs, but which I consider useful for the community.

Let's start with visual changes, very little one.

### DIFF
--- a/cookbook/templates/base.html
+++ b/cookbook/templates/base.html
@@ -100,19 +100,19 @@
         <ul class="navbar-nav mr-auto">
             <li class="nav-item {% if request.resolver_match.url_name in 'view_search' %}active{% endif %}">
                 <a class="nav-link" href="{% url 'view_search' %}"><i
-                        class="fas fa-book"></i> {% trans 'Recipes' %}</a>
+                        class="fas fa-fw fa-book"></i> {% trans 'Recipes' %}</a>
             </li>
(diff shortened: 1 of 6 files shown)


astappiev:
image
I would also like to have labels for these two submenus. We can hide them on wide screens, what do you think?

Maybe "Manage lists" and "Add Recipe"

vabene1111:
thank you very much for your feedback. The cost of the hosted solution was set based on several factors, including that we do not sell any user data, that it compares to at least a few other application and that, subtracting hosting and cost of doing business, taxes and so on we still at least make a few cents, but maybe we need to adjust them at some point, we will see (thanks for the feedback for now).

I think labels would do those menus good on mobile, i suggest `Tools` and `Add`.

Can you post a screenshot of how the visual changes look, makes it easier for me to verify what changed. Thanks!

vabene1111:
ok i have reviewed your changes and like them, thank you very much. I will look into adding the labels to mobile only nav

PR #1866, opened 2022-06-18
Updates with more packages

nicedevil007:
Updated the manual installation file to get everybody up and running.
There were missing requirements like the nginx installation and so on.

Now a first time linux user should be able to get this up and running.

### DIFF
--- a/docs/install/manual.md
+++ b/docs/install/manual.md
@@ -123,7 +123,7 @@ Things to edit:
 
 Execute `export $(cat /var/www/recipes/.env |grep "^[^#]" | xargs)` to load variables from `/var/www/recipes/.env`
 
-Execute `./bin/python3 manage.py migrate`
+Execute `bin/python3 manage.py migrate`
 
 and revert superuser from postgres:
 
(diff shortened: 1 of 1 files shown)


nicedevil007:
While I was updating the documentation here, I noticed that the installation isn't showing any version number. Dont know if that is a bug or just not developed?

image

And one last question that I wasn't able to figure out:
Is there a possibility to autoconfirm overwriting the static files with the update script? I tryed `-y` and `-yes`

image

vabene1111:
Thank you very much for the pr, that's awesome.

Regarding the Version number this is a knien issue for quite a while which I have not yet been able to fix as the Version number is added during build.

The confirmation can be removed, I am on my phone but it's either in the docker file or the Boot.sh Script with the correct Parameter, I think it's `--no-input`

nicedevil007:
> The confirmation can be removed, I am on my phone but it's either in the docker file or the Boot.sh Script with the correct Parameter, I think it's `--no-input`

That did the trick :) Just want to let you know that I made a PR for an other guy that started with some awesome scripts to make it easy for proxmox users => his scripts starts everything you need to get a LXC up and running with everything you need.

Now I'm able to provide an all in one update script with this `--no-input` argument.

If you are interested => take a look here for the PR => https://github.com/tteck/Proxmox/pull/290
Or even here for his complete lineup of scripts => https://tteck.github.io/Proxmox/

PR #1917, opened 2022-07-07
Bookmarklet fix

smilerz:
this is mostly tested - I don't have any way to test the import for CookbookApp.

Once that has been tested this is safe to put into develop/beta


vabene1111:
i can test that after merging, i have some test data i think, thanks for the PR

Issue #1027, opened 2021-11-05
Copy Me That import


tomtjes:
Import from [Copy Me That](https://www.copymethat.com) accounts would be great!

vabene1111:
i have seen that before but never used it, is there a way to export anything from a database on copy me that ? if so would you willing to provide me with some sample data so i can test the importer better ? if so feel free to email it to me at info@tandoor.dev

noja11:
I'll send you something in a moment

vabene1111:
note for myself: test data stored in my test data folder

integration will be added with one of the next updates

rshide:
Hey @vabene1111, has any work been completed on this integration yet?  I currently use Copy Me That and am wanting to switch to Tandoor.  If you haven't started on this yet I wouldn't mind working on it and submitting a pr.

smilerz:
@rshide - no he's hasn't worked on the app much in the last month.  

But please feel free to take a crack at it and submit a PR!  There is an integration class (cookbook.integration.integration.Integration) that you'll need to use.

vabene1111:
i have sample data provided by someone laying in my inbox waiting for integration, sadly family matters have forced me to take a break in developing, i hope that i will get back soon and then copy me that is on my immediate todo list :)

noja11:
Family always takes precedence! Thanks so much for your efforts and we’ll see you when life permits.



________________________________

vabene1111:
ok i had some spare time and added the copy me that importer thanks to the sample data i got from @noja11 

I pushed it on beta as i cant make another release before christmas but i think beta should be very safe to update to currently (no migrations) so feel free to try it out everyone (remember to switch back to stable afterwards to prevent unwanted updates in the future). 

Build will likely take an our or so, discord will give you a notification once beta is out (i hope).

vabene1111:
i will leave this open until someone trys it out (either beta or release) since i did not have time to do much testing.

rshide:
i just tested it out on the beta branch and it seems to work as I would expect.  I'm not seeing any data quality issues either.  thanks @vabene1111!

vabene1111:
Awesome, that's great. While the HTML format is not great it seems somewhat standardized so I had hoped for good results across different people.

tomtjes:
Thanks for the importer! It made me install Tandoor and try it out immediately. However, I feel like the switch to Tandoor would mean a substantial loss of data. Here are some issues I faced and suggested solutions.

[...]

PR #1937, opened 2022-07-13
Update copymethat.py

tomtjes:
- make use of field for source URL
- preserve "I made this" flag as keyword
- preserve long descriptions in full at bottom of steps (would be nicer if descriptions could be longer)
- preserve ingredient and step headers
- separate notes from steps

This solves part of #1027.


vabene1111:
awesome, thank you for those improvements

PR #1947, opened 2022-07-20
Add S3_CUSTOM_DOMAIN settings, closes #1943

### LINKED ISSUE #1943: Download and serve images from S3
### Tandoor Version

1.3.0

### Setup

Manual Setup

### Reverse Proxy

Others (please state below)

### Other

haproxy -> traefik -> tandoor

### Bug description

When using S3 to store recipe images, Tandoor sets the image url to the S3 url. This is all fine when using a publicly-available S3 provider that uses globally accepted certificates, but when using a self-hosted S3 setup that's only internally accessible and/or uses self-signed/internal ssl certs, images cannot be loaded by public/external viewers. Solution would be to let Tandoor get the image from S3, then serve it from it's own server.

### Relevant logs

_No response_

dmaes:
Add a setting for https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#cloudfront `AWS_S3_CUSTOM_DOMAIN`. This instructs django to use a different domain than the S3 endpoint in image url's.

This fixes #1943 in the sense that setups with a private S3 can setup a public proxy (with limited access rights) to this proxy.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -77,6 +77,7 @@ GUNICORN_MEDIA=0
 # S3_QUERYSTRING_AUTH=1 # default true, set to 0 to serve media from a public bucket without signed urls
 # S3_QUERYSTRING_EXPIRE=3600 # number of seconds querystring are valid for
 # S3_ENDPOINT_URL= # when using a custom endpoint like minio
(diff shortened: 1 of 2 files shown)


vabene1111:
awesome, i had hoped that there would be some easy fix like this. Thanks for implementing it.

PR #1948, opened 2022-07-20
fixed csrf

iamkarlson:
Added in Django 4.0: CsrfViewMiddleware verifies the [Origin header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin), if provided by the browser, against the current host and the [CSRF_TRUSTED_ORIGINS](https://docs.djangoproject.com/en/4.0/ref/settings/#std-setting-CSRF_TRUSTED_ORIGINS) setting. This provides protection against cross-subdomain attacks.

It broke this mechanism for installations behind proxy when origin is forwarded.


vabene1111:
awesome, thnaks. I will quickly read trough the docs and get this merged for the next release hopefully

PR #1952, opened 2022-07-22
fix minor issues in CopyMeThat importer

tomtjes:
improve handling of empty fields and fields that exceed character limits


vabene1111:
awesome, thanks

Issue #1983, opened 2022-08-07
Ingredient parsing for non-latin languages


andyjayne:
### Tandoor Version

1.3.3

### Setup

Docker / Docker-Compose

### Reverse Proxy

Traefik

### Bug description

When importing a recipe from a site in Japanese the ingredient string is not parsed into the correct amount or units. Looking at the ingredient parsing code, it is likely that this is the case for other non-latin languages when the food is found at the start of the ingredient string.

https://github.com/TandoorRecipes/recipes/blob/c7f75fe58f199f92f048bf5bf180869732006832/cookbook/helper/ingredient_parser.py#L224-L225

Example recipe:

https://cookpad.com/recipe/3302420

Example cases:

* `砂糖 50g`
  * **Expected:** Amount: `50`, Unit: `g`, Food: `砂糖`
  * **Actual:** Amount: `0`, Unit: None, Food: `砂糖 50g`
* `卵 4個`
  * **Expected:** Amount: `4`, Unit: `個`, Food: `卵`
  * **Actual:** Amount: `0`, Unit: None, Food: `卵 4個`

PR #1984, opened 2022-08-07
fix: ingredient parsing for non-latin languages

andyjayne:
Before this change the ingredient string for non-latin languages was not being parsed into the correct amount or units when the food is found at the start of the ingredient string.

This was because the regex being used was restricted to latin characters.

With this change the amount and units are correctly parsed from such a string.

Fixes https://github.com/TandoorRecipes/recipes/issues/1983


vabene1111:
thank you very much, i honestly never tought about non latin languages, very cool that this is working now. will merge it asap

Issue #1830, opened 2022-05-30
Change minutes to hours and minutes when at or above 60 minutes 


genebean:
### Is your feature request related to a problem? Please describe.

Reading wait time for recipes that cook a long time isn’t intuitive as most of us think in hours and minutes, not just minutes. For example, a crockpot recipe I added takes 10 hours but shows up as 600 minutes. 

### Describe the solution you'd like

I’d like values at or above 60 minutes to show as hours and minutes.

vabene1111:
fixed

genebean:
Very cool, thanks!

spidersouris:
Could we have the same change for Home/Search/Recipe View pages? It seems this is not applied there.

vabene1111:
I think I need to build a general formatter class to make the formatting of units/amounts/times/servings and what not standardized

PR #1986, opened 2022-08-09
Solves #1830 (Split times over 60 min into hours and minutes)

CameronJGrant:
Adds an option to have time split into minutes and hours. It seems most languages understand "h", "min" notation, so I've stuck with that.

image
image
image
image

### DIFF
--- a/vue/src/apps/RecipeView/RecipeView.vue
+++ b/vue/src/apps/RecipeView/RecipeView.vue
@@ -38,7 +38,7 @@
                         </div>
                         <div class="my-auto mr-1">
                             <span class="text-primary"><b>{{ $t("Preparation") }}</b></span><br/>
-                            {{ recipe.working_time }} {{ $t("min") }}
+                            {{ working_time }}
                         </div>
                     </div>
                 </div>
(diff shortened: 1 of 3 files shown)


CameronJGrant:
I realized I didn't rebase. Reopening with cleaner PR.

vabene1111:
thank you, i really like this change and also the clean and easy to understand implementation. One thing up for debate: do we really need a setting for this ? i think we can just implement this as the default behavior, back when i migrated the features from django to vue I was simply to lazy to implement conversion, i dont see why any one would change this setting and it reduces complexity if we dont add it right ?

feedback appreciated

CameronJGrant:
I agree with you.

My reasoning for implementing it as a setting is that I'm reluctant to force change a feature of the UI that has been around this long. That is, maybe someone, somewhere in the world, actually likes their display being in minutes only. (This comes from personal experience of being forced into UI changes that a dev thought was "better". :P )

_I'm sure I'm over thinking it for something this small_

That said, I'd certainly be open to removing the option and keeping it as hour/minutes. Or perhaps, defaulting to hours/minutes display in models.py and hiding the setting somewhere.

I'll leave it up to you. Let me know and I'll change it.

vabene1111:
Perfectly valid reasoning and I am not sure if it's just the frustration I had recently with the overall complexity or if it's actually the better decision but it would be awesome if you could remove the setting code.

We will have it in the pr forever so if there is a huge demand we can re add it any time.

Thanks a lot for the pr

CameronJGrant:
Sounds good! All set.

PR #2005, opened 2022-08-23
Hide name of step in ingredients card if there are no ingredients

DeadHunter:
Avoids unnecessary listings in the ingredients card if a step has no ingredients.


vabene1111:
Actually this is a feature not a Bug as it allows to "hide" instructions in the ingredient list. When we add the setting to switch to summing the ingredients on top we will probably remove unused headers as well.

But thanks for the well meant contribution.

PR #2013, opened 2022-08-28
Update documentation on external recipes

nough:
apologies that this pull request is messy, I don't fully understand how branches work in github. This pull request should hopefully make the language in the documentation about external recipes a little clearer.

If I have time, I'll add text about how converting from external to internal recipe is a little difficult:
- When you click "convert to internal", you can no longer access the external pdf.
- You can't convert back to external without using the API (you want the retrieve recipe API call: http://tandoor.hostname:8080/docs/api/#tag/api/operation/retrieveRecipe
- Recommended steps would be to copy to new recipe, then convert one of them to internal, then once you're happy that it's converted, delete the external recipe.


vabene1111:
Thanks, this looks good. External recipes are a real mess and I need to improve on them but very little people actually use them so it's kinda hard to justify time spend.


This will help for now, I will get it merged asap.

Issue #1981, opened 2022-08-05
Permissions deneid for view a page 


ignis-draco:
### Tandoor Version

1.3.3

### Setup

Manual Setup

### Reverse Proxy

Nginx Proxy Manager (NPM)

### Bug description

After the update from 1.3.2 to 1.3.3 it is no longer possible to view recipes. 
The error always appears at the bottom right 

"Forbidden
There was an error loading a resource! {"detail": "You do not have the required permissions to viwe this page!"}"

I get these errors both with newly created recipes and with existing ones.  

The gunicorn log says

```
[2022-08-05 11:42:23 +0200] [480] [DEBUG] Current configuration:
  config: ./gunicorn.conf.py
  wsgi_app: None
  bind: ['unix:/var/www/recipes/recipes.sock']
  backlog: 2048
  workers: 1
  worker_class: sync
  threads: 1
  worker_connections: 1000
  max_requests: 0
  max_requests_jitter: 0
  timeout: 30
  graceful_timeout: 30
  keepalive: 2
  limit_request_line: 4094
  limit_request_fields: 100
  limit_request_field_size: 8190
  reload: False
  reload_engine: auto
  reload_extra_files: []
  spew: False
  check_config: False
  print_config: False
  preload_app: False
  sendfile: None
  reuse_port: False
  chdir: /var/www/recipes
  daemon: False
  raw_env: []
  pidfile: None
  worker_tmp_dir: None
  user: 1000
  group: 33
  umask: 0
  initgroups: False
  tmp_upload_dir: None
  secure_scheme_headers: {'X-FORWARDED-PROTOCOL': 'ssl', 'X-FORWARDED-PROTO': 'https', 'X-FORWARDED-SSL': 'on'}
  forwarded_allow_ips: ['127.0.0.1']
  accesslog: None
  disable_redirect_access_to_syslog: False
  access_log_format: %(h)s %(l)s %(u)s %(t)s "%(r)s" %(s)s %(b)s "%(f)s" "%(a)s"
  errorlog: /tmp/gunicorn_err.log
  loglevel: debug
  capture_output: True
  logger_class: gunicorn.glogging.Logger
  logconfig: None
  logconfig_dict: {}
  syslog_addr: udp://localhost:514
  syslog: False
  syslog_prefix: None
  syslog_facility: user
  enable_stdio_inheritance: False
  statsd_host: None
  dogstatsd_tags: 
  statsd_prefix: 
  proc_name: None
[...]

PR #2020, opened 2022-09-01
Update Doku for manual Installation 

ignis-draco:
1) add the "Install project requirements" to the update commands list 
2) add infos from #1981 to doku  
I think it is helpful to mention briefly here that problems can occur. Even if it is not necessary for every update.


vabene1111:
Thanks for the contribution 

Hmm generally speaking it should never actually be required to run the comma more than once.

What do you think rewording this to something like if it's 0 files make maybe run again also See issue link?

PR #2044, opened 2022-09-16
Fix 'markdown' 3.4 version incompatibility 

ambroisie:
See [the release notes](https://github.com/Python-Markdown/markdown/blob/master/docs/change_log/release-3.4.md#previously-deprecated-objects-have-been-removed) about the removal of the deprecated argument.


vabene1111:
thank you very much, didn't even notice this was broken yet 😂

PR #2052, opened 2022-09-21
Update boot.sh

Szeraax:
Set default value to 1 if null or unset for gunicorn workers/threads.


kristof-mattei:
https://github.com/TandoorRecipes/recipes/blob/904358bb003ab581da2f4219f90e55a010bb426c/boot.sh#L68

Isn't there also a `$` missing:

```
exec gunicorn -b :$TANDOOR_PORT -workers :$GUNICORN_WORKERS --threads :GUNICORN_THREADS --access-logfile - --error-logfile - --log-level INFO recipes.wsgi
                                                                      ^^
```

Szeraax:
@kristof-mattei that was an issue present that was fixed in https://github.com/TandoorRecipes/recipes/commit/c7f75fe58f199f92f048bf5bf180869732006832. If you look at the current develop branch, it is not present. However, I don't know what branch is used to build the docker image, so your comment may be right and the docker image is affected. 

I will have to refer to @vabene1111 to determine if the c7f fix is in the docker image or not.

BapRx:
It seems to be ok, from the v1.4.0 docker image:
```bash
❯ docker compose exec web_recipes sh
/opt/recipes # tail -n1 boot.sh 
exec gunicorn -b :$TANDOOR_PORT --workers $GUNICORN_WORKERS --threads $GUNICORN_THREADS --access-logfile - --error-logfile - --log-level INFO recipes.wsgi
/opt/recipes # 
```

Szeraax:
Thank you for verifying that, @BapRx.

vabene1111:
thanks everyone for helping debug this so quickly. The image is build from the master branch, at least the latest and tagged images. I will get your fix merged, i will just change the numbers to have 3 default workers (because everyone has at least one core) https://docs.gunicorn.org/en/stable/design.html#how-many-workers and 2 threads (because there is no recommended default and it sounds good :D)

PR #2065, opened 2022-09-26
Update faq.md

smarth42:
updated user invite location.


vabene1111:
thanks for the clarification

PR #2101, opened 2022-10-03
Allow 1/16th to be a fraction (small salt amounts)

Mikhail5555:


vabene1111:
thanks for the PR. 

I do not use the american notation much but have never seen anything that used 1/16 and even if so, is it really possible to measure such a small quantity? 

I kinda liked 10 to be the smallest fraction, but we can change it, just trying to think if its very useful this way?

PR #2124, opened 2022-10-17
Fixed: Scroll problem in the navbar on desktop

TalhaQuddoos:
The main navbar was not scrollable, it just went under the screen. 
Provided it a height of 400px (please increase if you want) and overflow: auto. 

image


vabene1111:
thats quite some spaces 😂 but makes sense. Two questions: have you tested this on mobile (i know we had the same issue there and it was fixed so just making sure your fix does not conflict with that) and second what do you think about using a relative height as the max height ?

thanks for the contribution and sorry for the late reply

PR #2128, opened 2022-10-21
Fixed: ports are on the wrong service

rdangdev:
the actual webserver is running on port 8080 and on the web_recipes service, as per all the documentation. 80 will not work, at least on the latest patch.

### DIFF
--- a/docs/install/docker/plain/docker-compose.yml
+++ b/docs/install/docker/plain/docker-compose.yml
@@ -13,6 +13,8 @@ services:
     image: vabene1111/recipes
     env_file:
       - ./.env
+    ports:
+      - 80:8080
     volumes:
       - staticfiles:/opt/recipes/staticfiles
       - nginx_config:/opt/recipes/nginx/conf.d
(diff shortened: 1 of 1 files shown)


vabene1111:
have you tested the change? i know it sounds mundane but its easy to get confused with the mappings and I am not 100% certain so would need to test it first.

rdangdev:
all good, change works on my end. one more test would be a good idea.

kleest:
I think this change is wrong. First, it breaks syntax by having a `ports` property without an actual list of port mappings. Second, the `web_recipes` services should not expose ports, since the nginx service is used as a reverse proxy which forwards all requests to the `web_recipes` service internally. I propose reverting the commit, so that `web_recipes` contains no `ports` and `nginx_recipes` has the following `ports` configuration:
```
    ports:
      - 80:80
```

Also see the [comment by @frudolph77](https://github.com/TandoorRecipes/recipes/commit/37bd2017b01458b01bda29c7b919d36c931d47e3#commitcomment-89541603).

vabene1111:
i have reverted the change, i had tested the original version when i first added it and it definitely worked so will leave it as is, every change that might be required to it will be on an individual setup level

PR #2096, opened 2022-10-01
docs: :memo: Add Hebrew readme

YonLiud:
## docs: :memo: Add Hebrew readme

I added the Hebrew translation as part of my journey to translate all projects and applications I use on a daily basis to invite more Hebrew Speakers and Israelis into using those apps.

In case this PR gets accepted I will attempt to translate all of the rest of the markdown files requiring translation!

PR #2143, opened 2022-10-31
translate to brazilian portuguese

paulpessoa:
I really liked the application, I would like to help in its translation and started with the main readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,11 @@
   <br>
 </h1>
 
+#### _Read in other languages_
+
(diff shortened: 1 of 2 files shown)


vabene1111:
Thank you very much for your contribution. As discussed recently in this PR https://github.com/TandoorRecipes/recipes/pull/2096 I do not intend on translating docs/readme/similar files. The reason is that they change somewhat often and changes can be very important. If i am not able to update all versions then we have a problem of out of sync documents which i want to avoid. 

but you are more than welcome to translate the application using translate.tandoor.dev

PR #2168, opened 2022-11-14
Merge pull request #1 from TandoorRecipes/develop

Faisal-byte:
New Scroll Up Button On Every Page


vabene1111:
- files changed that should not be changed
- code looks like it was copied from somewhere and not properly adapated
- comments dont match code
- naming is not useful

PR #2169, opened 2022-11-14
add logs detailing deserialization errors for default recipe import

tlowerison:
Hi there, it took me quite a while to figure out exactly why my imports were failing, especially since I'm unfamiliar with django and it's serialization framework. Adding these logs in, which are only output on deserialization failure, made a tremendous difference in my ability to debug imports and would be useful for all forms of installation (especially those such as docker, kubernetes, etc. where editing the source code isn't really a knob you can turn).

### DIFF
--- a/cookbook/integration/default.py
+++ b/cookbook/integration/default.py
@@ -29,6 +29,9 @@ class Default(Integration):
             recipe = serialized_recipe.save()
             return recipe
 
+        print(json.dumps(data, indent=2))
+        print(json.dumps(serialized_recipe.errors, indent=2))
+
         return None
 
     def get_file_from_recipe(self, recipe):


vabene1111:
Thank you very much for the well meant PR. While i apreciate it I must admit I am not a huge fan of random print statements in code, if there is an error in the function a proper exception should be thrown and exception during import/export are collected as part of the Import/export log objects and returned to the user. 
I am not 100% sure which error you cought using this method so cant really verify but generally this error should then be fixed for upcoming releases so that print statements should not be needed. 

Could you describe further what the issue was you found out in the end? If we leave this in we should wrap it in a condition to only print this kind of data if debugging is enabled.

PR #2179, opened 2022-11-19
Unpin pin from recipe page.

bee395:
This PR prevents a single recipe from being pinned multiple times
It also ads the ability to unpin a pinned recipe from the "RecipeContextMenu"
It also shows a toaster as confirmation whenever a recipe is pinned/unpinned

#### Possible improvements
- Preferably all the functionality related to pinning, should be in a separate utility. Since the "unpin" code is also used in the "RecipeSwitcher"
- If a pinned recipe is opened and is then unpinned via the "RecipeSwitcher" the "RecipeContextMenu" still shows the option to unpin the recipe (since there's no communication between the "RecipeSwitcher" and the "RecipeContextMenu")


vabene1111:
Thank you for the well documented pr. The changes make sense, will review and merge if possible

vabene1111:
Ok code looks good, just one question. Also for the next time just so you can save yourself some time you do not need to add the translation to erve y language, just en. Json is fine.

bee395:
Thanks for the quick review
Will definitely remember to only do the en translation for next time, took indeed a lot more time than expected 😅

vabene1111:
do you think you could remove the language files from your commit by reverting their change ? would save me a lot of hassle manually merging them since people have changed the files since your PR.

bee395:
Sure no problem, I will remove the empty translations

bee395:
I removed all the empty translations (only kept EN and NL)

I also found 2 translation keys that were used in the frontend, but not translated. So I added them in a separate commit

vabene1111:
thank you very much!

PR #2215, opened 2022-12-15
Allow PWA Manifest to work behind CF Access / certain proxy configs

jwalker343:
I am running this app behind Cloudflare Access. Cloudflare Access (and potentially other proxies) ensure that there is a specific authorization cookie on each request going through the proxy. Requests without this cookie get redirected to the login page, or blocked.

It appears that from spec when using a `link` tag _specifically with_ `rel="manifest"` the cors `crossorigin` attribute is set to anonymous/omit and cookies are stripped from the request by the browser. This is documented here: https://github.com/w3c/manifest/issues/535 with links to several other apps that have implemented similar solutions.

This change allows the browser to pass cookies with the request and properly retrieve the manifest.json file, which allows the PWA to be installed as expected.

### DIFF
--- a/cookbook/templates/base.html
+++ b/cookbook/templates/base.html
@@ -19,7 +19,7 @@
     <link rel="mask-icon" href="{% static 'assets/safari-pinned-tab.svg' %}" color="#161616">
     <link rel="apple-touch-icon" href="{% static 'assets/apple-touch-icon.png' %}" sizes="180x180">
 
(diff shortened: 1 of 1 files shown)


vabene1111:
thank you very much for the well documented and reasoned PR. I will need to look at the sources you linked to verifiy this for myself just for security but should not be a problem, will get this merged with the next release if i dont have any further questions.

PR #2233, opened 2022-12-27
Allow an invited user to request the space settings

Knalltuete5000:
Currently an invited user to a space is not able to access the space settings. The request for the space settings was introduced by #1860
The result for an invited user was a `403 Forbidden`.

This fix enables a user to make a `GET` request to the space settings for a space.

### DIFF
--- a/cookbook/views/api.py
+++ b/cookbook/views/api.py
@@ -34,7 +34,7 @@ from oauth2_provider.models import AccessToken
 from recipe_scrapers import scrape_me
 from recipe_scrapers._exceptions import NoSchemaFoundInWildMode
 from requests.exceptions import MissingSchema
-from rest_framework import decorators, status, viewsets
+from rest_framework import decorators, status, viewsets, permissions
 from rest_framework.authtoken.views import ObtainAuthToken
 from rest_framework.decorators import api_view, permission_classes
 from rest_framework.exceptions import APIException, PermissionDenied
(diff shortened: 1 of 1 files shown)


vabene1111:
Hi, thanks for the contribution. 

I see the reasoning for what you did but I am not 100% sure if I like the feeling (for a lack of better description of it). The space settings are the administrative area for the space owner, I would think both users and also contributors to this project would think of this page to only be accessible to the space owner or at least only admins. 

There are currently no dangerous information pieces being displayed but but that could change. Do you really think its necessary for a normal user to be able to view this admin page?

Knalltuete5000:
Ok. Maybe I did not configure it right and I totally agree with you that a user does not need to access the space view. But there is currently this problem. One solution I came up with just now ist to introduce an new endpoint like `/api/space/settings/{id}` to only retrieve a subset, only non sensitive, information of the spacesettings.

What do you think of this solution?

vabene1111:
it would not necessarily need to be an extra endpoint, certain data could be filtered out. The problem I am having is that I do not see a reason to introduce this addional complexity as only the administrators of a space should have to worry about the users, the limits or the other settings available on the page right?

Leyvur:
Hi, I think I'm currently struggling with this issue.

Each page request of `mytandoor.tld/search/` triggers an API call to the endpoint `mytandoor.tld/api/space/1/`. But access to it is blocked for regular users. This leads to a lot of 403 errors which eventually trigger fail2ban at my proxy server.

Since it seems that the search function needs access to this data, I would like to suggest enabling this endpoint for users as well.

vabene1111:
The need for enabling this has been resolved with the next update. It's not out yet but should be in a few days I think.

PR #2267, opened 2023-01-11
Clarify the folder in which yarn needs to be run

besendorf:


vabene1111:
this does not make sense to me, it does not name the folder where yarn needs to be run, is this a mistake?

besendorf:
You're correct somehow a word is missing there.

vabene1111:
ok perfect, wasn't sure what you meant, thanks for the addition.

Issue #1552, opened 2022-02-18
Import Recipes from Cookidoo


codac:
### Is your feature request related to a problem? Please describe.

Add Integration for importing recipes from Cookidoo (https://cookidoo.de/). A payed account is required.
Maybe the technique of https://github.com/auino/cookidump can be reused.

### Describe the solution you'd like

I would like to be able to import recipes from cookidoo directly into Tandoor Recipes using the import functionality of integrations.

### Describe alternatives you've considered

/

vabene1111:
sadly not but i am building some into the application currently. Basically drag the "bookmak me" button on the import page into your bookmarks, open a cookiedoo recipe and click on the bookmark and see if that works. (might only work in chrome)

Tuphal:
2 weeks ago, it worked with cookidoo. 
In the last 6 month I imported my favorites without problem.

Now I get error "403 forbidden".

Either cookidoo was updated or tandoor..

Tuphal:
I have no problems with importing the repices from cookidoo. I do so for the last 6 month with all my favorites:

![grafik](https://user-images.githubusercontent.com/42291781/155673788-862ceca9-d6c1-410c-b6ff-a9024c807828.png)

vabene1111:
@Tuphal using the bookmarklet or URL directly ?

Tuphal:
copying the URL directly.
But I need use the manual selection and editing, because the crawler is not getting the description of the steps, just the ingredience.

cmro:
Maybe not the most elegant solution, but what also works is to copy the source code of the cookidoo page and paste it under import/source. This way also the steps are imported. only cookdoo specific emojis (e.g. rewind) are not imported correctly.

vabene1111:
that works or just using the bookmarklet (which does nothing different than copy the code and move it to the importer). I will test this myself once i am done refactoring the url import page

cmro:
I can confirm that the bookmarklet works (even in Firefox), but you have to make sure that your tandoor instance is running on https, otherwise it does nothing (I guess that's some kind of cross-site-script protection). Also, you have to allow popups for cookidoo.de

Some smaller issues: 

-  As already mentioned, some thermomix specific emojis are not translated correctly (e.g the dough mode becomes , stir:  and so on). I suspect they use symbols that are only available in their own font. This sometimes leads to results that are difficult to read, however, this will probably be difficult to fix.
[...]

PR #2270, opened 2023-01-13
#1552 import recipes from cookidoo

MarcusWolschon:
Implement feature request #1552 by adding a custom importer for cookidoo


vabene1111:
Could you contribute this directly to the recipe scrapers library? This way it would be centrally maintained and tested and everyone would benefit from it, also non tandoor users

MarcusWolschon:
No, the actual scraping is already completely in the Recipe Scrapers Library and gets inherited.
This just overrides 2 behaviors that the library deliberately does in a different way to the stregth and weaknesses of TandoorRecipes

1)
They explicitely converts the recursive list of steps into a string.
I overrides that function to get the flat list Tandoor can handle but other clients using the library can't.
That's the contract of that method in all parsers and I can't adjust all other parsers to add a second method in the interface.
https://github.com/hhursev/recipe-scrapers/blob/main/recipe_scrapers/_schemaorg.py#L225

2)
Also the conversion of custom Unicode characters Cookiedoo uses is an adjustment to the limitations of TandoorRecipes because we can't add the custom font Cookiedoo uses in it's CSS to display the symbols matching the . (At the very least for interlectual property reasons.)
I can't possibly make that choice for apps using the library 

I am contemplating donating other improvements directly to the Scraper library.
Such as code to handle ingredient units that contain a space "geh. Löffel"/"heaped spoon".
They would benefit everyone and are not specific to the abilities and limitations of TandoorRecipes.

MarcusWolschon:
@vabene1111 I found that the Recipe Scrapers Library already offers a method instructions_list() instead of the flat instructions() we used, so I could get rid of a specialized cookiedoo scraper.
I moved the handling of the Thermomix-speific instructions into an existing cleanup method.

vabene1111:
hi, thanks. While (as stated in the other PR) i dislike hardcoded lists I guess thermomix is large enough so that it makes sense in this case. 

I haven't looked at the differentce between instructions and flat_instructions, is there anything that could break in other parser due to this change? Have you tested the change with different sites/the test suite ?

MarcusWolschon:
I tested a number of randomly chosen recipe-sites and they where all imported correctly with steps.
Section headers become their own step, wich is the best we can do without a concept of sections.

vabene1111:
ok thanks. Reviewing this now I still dont understand why the cleanup methods should reside in Tandoor? Everyone using recipe scrapers to parse cookidoo recipes would benefit from the replacement so it would be better there. 
One thing that recipe scrapers currently cant do is localize so maybe the text would need to be handled by emoji's or something similar. 

I also dont know the difference between the two instruction functions yet but could't the old function be fixed in recipe-scrapers directly so that it follows the same structure as all other parsers?

MarcusWolschon:
I did look at custom python markdown extensions to handle this on the display-side but could not find where the instance of "markdown" is created in Tandoor to add a 4th extension to it.
https://python-markdown.github.io/extensions/api/

MarcusWolschon:
I tried to use the existing "automation" function "keyword replace" to do replace the custom UTF-8 characters,
but it doesn't seem to work.
I guess "keyword" here means tags.

I guess I could give the existing automation a string-replacements for steps in addition to the existing tags, ingredients and foods.
Then this part could be moved to a user configuration instead of being hardcoded.
This PR could be reduced to just instructions() -> instructions_list() and there would be no hardcoded replacements in the codebase anymore.
@vabene1111  What do you think? Should I implement that?

vabene1111:
dont have time to write much but

- replacing the charcters the way you do it is fine with me, with emojis i meant actual emojis (standard ones) but not sure if there are proper fitting ones, probably not
- i see that the localization in recipe scrapers is difficult, i dont really like it but its probably best to leave it in tandoor
- i do not want to change instructions() to instructions list as that would change the behavior of all importers. instructions list only splits instructions on linebreaks which the frontend of tandoor is already able to do.


so i guess only two changes necessary
- remove instructions list as the frontend can handle this
- change the default locale to English as this is the default locale everywhere (even tough I understand why you did it since thermomix is primarily german)

MarcusWolschon:
done

MarcusWolschon:
@vabene1111 Just pinging, take your time. Just making sure it's not forgotten.
The changes you wanted have been done 3 weeks ago.

vabene1111:
thanks, i know. Will get this merged when I find It appropriate for the release cycle.

jmvargas:
Hello @vabene1111 , is this going to me accepted soon? it would be a nice feature

vabene1111:
This will be accepted when I accept it. Nobody pays and only few people thank me for working on this project so I do what I can but my time is limited.

PR #2272, opened 2023-01-15
handle 2 more special cases in ingredients

MarcusWolschon:
Add handling for 2 more classes of special cases in ingredients


MarcusWolschon:
Is there a way to get all units ad unit-aliases that contian spaces from the database?

vabene1111:
It would be Unit.objects.filter(space=<current space>).all()

but that would mean that chaching needs to be added as every request to the ingredient parser would otherwise infer a database request against the whole set. I am not sure if this is already implemented to some degree but that is a rather large change so i would say we remove the hardcoded part from this PR, if you want you can play around with the database and if you come up with a solution that improves the results we can discuss it in its own PR, is that ok ?

MarcusWolschon:
done

PR #2273, opened 2023-01-15
Update .env.template

jacobgc:
Fixed broken URLs in .env.template


vabene1111:
oh thank you, never looked at that. need to add proper docs for all env variables but this helps already

jacobgc:
No worries :)

PR #2301, opened 2023-02-03
Fix path traversal in manual installation steps

RoGryza:
[According to gixy](https://github.com/yandex/gixy/blob/master/docs/en/plugins/aliastraversal.md) the example nginx config for manual installation is vulnerable to a path traversal attack


vabene1111:
thank you. This needs to be fixed in all versions of the nginx config in the docs, will do this.

vabene1111:
will add it manually as its easier than merging the PR, thanks!

Issue #170, opened 2020-08-29
Changing Case of units, foods and keywords not possible


stewartadam:
### Version
Please provide your current version (can be found on the system page since v0.8.4)
Version: 0.10.1

### Bug description
During recipe web import, changing the case of an unit (e.g. `Cups` to `cups`) causes page errors to appear on any change (i..e each keypress) in the unit field in an ingredients row. Note that this only occurs if the unit you are changing case for does not already exist in your installation.

1. Import any web recipe. The units on your ingredient row could look like this:
    ![image](https://user-images.githubusercontent.com/77808/91629059-4e782500-e97a-11ea-8c7f-8ff0d28c1a3e.png)
2. Click into one of the parsed ingredients **that does not already exist in your installation** and retype it exactly as is, but lowercase (e.g. `TBSP` to `tbsp`) and hit <enter> to submit the replacement ingredient:
    ![image](https://user-images.githubusercontent.com/77808/91629068-5f289b00-e97a-11ea-9a66-e523303a4b34.png)
3. Note the field now says "Select One":
    
![image](https://user-images.githubusercontent.com/77808/91629070-68196c80-e97a-11ea-8199-17286cce448e.png)

4. Interacting with another field or returning to edit the empty "Select One" field and retype the unit (e.g. `tbsp`) yields pop-up errors on every keypress:
    ![image](https://user-images.githubusercontent.com/77808/91628657-c2b0c980-e976-11ea-9bf9-1c657f713589.png)
5. Use the eraser button to clear the broken unit field
5. Re-enter `tbsp` into the field and hit <Enter> to accept
6. Submit the recipe, note in the `POST /data/import/url` form data that `recipeIngredient[x].unit=null`.

Actual results:
- Unit field gets into a broken state and cannot be fixed without total page refresh and losing progress editing the imported recipe
- Recipe import fails, partially importing the recipe without giving much detail to the end-user

Expected results:
- specifically: unit field can be reset to a working state by pressing erase button
- generally: changing ingredient case is a valid operation that does not produce UI errors
- failure to import recipes during web import provide some human-readable error message to the user
- specifically: import should act as an atopic operation and failure should not produce a semi-completed recipe
- generally: import should be more resilient to localized failures during ingredient row import, perhaps zero/null out that entry and carry on?

stewartadam:
During step 2, one can observe auto-complete requests going through as one types e.g. `GET ?q=tbsp`
```
{"results": [], "pagination": {"more": false}}
[...]

PR #2309, opened 2023-02-10
Feature/edit optional food attribute

bee395:
This PR makes it possible to edit foods, with the unit or name missing, when importing.
This problem was mentioned as part of #170


vabene1111:
thank you very much for the RP, I did not see it until i fixed this myself so closing this, but thank you very much for already adding this.

PR #2326, opened 2023-02-20
Updating/improving kubernetes deployment and its documentation

alexbarcelo:
The documentation for kubernetes is fine, but it used the 1.0.1 release as its base.

I believe that leaving `latest` with some explicit recommendations is good enough in this context.

Also, a minor thing, because nginx is deployed, `GUNICORN_MEDIA=0` seems to be a sane default (and removes the warning on recipes.local/system/ Media Serving section).


vabene1111:
thaks, gunicorn media should be `0` by default but probably does not hurt setting it explicitly.

vabene1111 (approved):

PR #2380, opened 2023-03-21
add: support to scale numbers/amounts in a recipe step

### LINKED ISSUE #1706: Scale arbitrary amounts in the text by the global portions multiplier

#218 added the ability to reference ingredients in the text which automatically can use the same amount from the ingredients list, which is scaled by the global amount of portions multiplier.

Beyond that, I'd like to be able to scale arbitrary numbers by the same multiplier.

This is useful for example when you are going to layer something, but also save some amount for later to garnish.

In my current recipe I added
1. step: grate an amount of cheese
2. step: use half of that cheese
3. step: use the other half minus a garnish amount
4. garnish with grated cheese

Maybe just make the portions variable (or the fraction of the default amount) available to be used in the text with jinja:
`set {{ 100 * portionsFrac }}g of flour aside for later` will become
- `set 100g of flour aside for later` with default portions of the recipe
- `set 500g of flour aside for later` if you set portions to 5 times the default

_No response_

### Additional context

_No response_

RomRider:
Simple POC to fix #1706.

image
image

### DIFF
--- a/cookbook/helper/template_helper.py
+++ b/cookbook/helper/template_helper.py
@@ -15,12 +15,14 @@ class IngredientObject(object):
     unit = ""
     food = ""
     note = ""
+    numeric_amount = 0
 
     def __init__(self, ingredient):
         if ingredient.no_amount:
             self.amount = ""
         else:
             self.amount = f"<scalable-number v-bind:number='{bleach.clean(str(ingredient.amount))}' v-bind:factor='ingredient_factor'></scalable-number>"
+            self.numeric_amount = float(ingredient.amount)
         if ingredient.unit:
             if ingredient.unit.plural_name in (None, ""):
                 self.unit = bleach.clean(str(ingredient.unit))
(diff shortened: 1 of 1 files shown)


vabene1111:
looks good and makes sense, thanks for the PR

RomRider:
Thanks, should I update some documentation to get this merged? Or maybe you have suggestion for another naming instead of `numeric_amount`?

I didn't want to introduce a breaking change and thus didn't reuse `amount`.

vabene1111:
now that you say that, is there a reason you do not access `amount` directly ? because i do not really see why, its not manipulated so it should not hurt reading it?

Some documentation would be awesome in the features section of the docs since the original scaling isn't documented as well, other than that we can try to add a button to the editor although i am likely going to change the editor again in the future so i can also write that down for when I work on that.

RomRider:
Reason I'm not using amount is because you already include the html tags around the number in the value of `.amount`, thus the jinja template parser can't do math with such a string.

I could strip it and re-add it again, but that's ugly.

Will update the documentation, but I'm not sure what you mean about a button?

vabene1111:
understood. 

with a button i mean some kind of way of adding the template tag without going to the docs and finding that it is possible. But just leave it out for now, we can add that once the editor is reworked at some point.

PR #2403, opened 2023-03-29
enable toolbar seperate from debug logging

smilerz:
I like running on debug mode all of the time, but the toolbar adds a very noticeable drag on performance (plus my other users don't need to see it)

This lets you turn on debug and the toolbar seperately.


vabene1111:
makes sense and is fine to for me :)

Issue #2402, opened 2023-03-29
Add input for a missing ingredient count to the "Make Now" search


ignas2526:
### Is your feature request related to a problem? Please describe.

When the "Make Now" search filter is switched on, it shows recipes for which all ingredients are available or one is missing. In some situations, it would be very useful to see recipes for which two or more ingredients are missing.

The motivation behind this request is that the difficulty of obtaining ingredients does not correlate with their count. Obtaining one seasonal ingredient outside of the season can be impossible while obtaining two readily available ones from a supermarket five minutes away is easy.

### Describe the solution you'd like

Next to the "Make Now", add an input field that would accept a number of the ingredients starting from zero. Show recipes with the missing ingredient count that is lower or equal to that number.

### Describe alternatives you've considered

Go to the specific recipe, click "Add to Shopping" and see how many ingredients will be added. With a small number of recipes and a rough knowledge of which recipes require what ingredients it is feasible, but as the number of recipes grows, this will become not feasible.

smilerz:
I looked into doing something similar to this, but the query that it generates is very resource intensive, so is very unlikely to ever be implemented on the front end.

There is a technical workaround (requires knowledge and ability to edit database tables directly) that I can share with you if you ping me on Discord. Not exactly what you are looking for, but it's close.

PR #2404, opened 2023-03-30
Add ability to set maximum missing ingredient count

ignas2526:
Implementation for the feature request https://github.com/TandoorRecipes/recipes/issues/2402

I couldn't get Vue running locally and I'm not a Vue person, so I welcome suggestions on how to make that part better.

I've added logging to my PostgreSQL and ran the explain. The impact of this change on performance is about 0.3% which is negligible.
existing `filter(missingfood=missing)` : `Limit (cost=49004106.03..49004106.16 rows=2 width=1292)`
new `filter(missingfood__lte=missing)` :  `Limit (cost=49127818.45..49127819.13 rows=10 width=1324)`

However, "Make now" is very slow. It makes the query about 22 times slower. When "Make now" is disabled: `Limit (cost=2193183.85..2193183.98 rows=2 width=1292)`.

The only way to solve this performance issue is to precalculate the number of missing ingredients for each recipe. I would suggest adding a materialized view. However, each time ingredient or recipe changes, the code will need to issue a refresh to the view. Maybe later I will try to add at least a regular view to see how impactful it would be. In general, I think that adding views and precalculating things for search is the way to go to improve performance.

Even better would be to have a scheduled job that runs every minute and recalculates the view if the flag is set. The flag would be set if the ingredient or recipe changes. In this scenario, UX would be better as the user could change a couple of ingredients before the refresh happens and the GUI would remain snappy as recalculation will be done in BG.


vabene1111:
hi, thanks for your PR.

to be honest I am not 100% into every little detail of the search because it was contributed by @smilerz and I never got around to look at every detail.

Given that the search is probably the part that causes me the biggest pain in maintaining this project to the point were I was close to just stopping development at all since I had zero motivation at fixing it, thats why I am very hesitant to work on it at all. For the time being I have focused on other things but this problem still lingers around (I have improved performance a lot but its not nearly were I would want it).

Precalculating attributes for the search makes sense in my opinion, I would probably work with database triggers/save events, I also think we need to split the search into two searches, one normal one on the start page and one advanced search. 

Also I think all the settings need to be removed from the UI or at least hidden in some very very deep corner because its all way to complicated for any user without a degree in CS. 

So to sum it up, the search needs a huge overhaul in terms of performance and usability to the point were I think rewriting it is the only option.

P.S.: no offense to @smilerz btw. the search is incredibly powerful and for a technical user very good but simply not working for a "normal" non technical user.

smilerz:
@vabene1111 none taken, UX is hard and the vision in my head outstripped my ability to execute.

smilerz:
@ignas2526 - there are tradeoffs between using the Django ORM and native PostrgreSQL (most importantly in terms of migrations) that need to be carefully considered.  

In Django the triggers are implemented as [signals ](https://docs.djangoproject.com/en/4.1/topics/signals/).  You see conflicting recommendations on whether to use signals or to override save() .  If you happen to find an elegant solution that improves performance without adding management complexity I'm sure everyone would appreciate it!  But I'd recommend trying to stay within the capabilities of the ORM, otherwise future code changes will likely break any enhancements.

vabene1111:
Yes, I agree with all of the above. Signals is actually what I meant with triggers as it's kinda the orm equivalent of a direct database trigger, I just forgot what they were called. 

I think we probably need to start by defining a list of all the features available with the search engine and then draft a solution on how to make each one efficient. 

We might also be able to utilize caching more extensively to improve the search performance.

vabene1111:
That said I think we can merge this pr and move the discussion to a seperate issue.

I am happy for all help I can get here as this is a huge task.one good thing about it is that we can probably solve one problem at a time withlut breaking anything.

smilerz:
I implemented fairly complete test cases for search - so that is a good starting point for breaking down capabilities (and tests to validate nothing got broken!)

PR #2416, opened 2023-04-06
Small fix for german language

dnlmlr:
Very cool project! 

As a native German speaker this instantly stood out to me, so here is a quick fix


vabene1111:
thanks for the fix, must have been someone from austria or so who made that translation

PR #2420, opened 2023-04-11
Fixed typo

noxonad:


vabene1111:
this is a fix for an english type in the romanian translation which does not make much sense

PR #2431, opened 2023-04-22
Pytest fixes

smilerz:
all tests now should pass


smilerz:
> oh boy thats a lot of changes ... this is not only tests but also code, have you tested that this does not change anything?

outside of tests it's all cosmetic or removing unused modules.  VSCODE enforces max-line-length which makes it look like a lot more than it actually is.

vabene1111:
ok, i saw a few small code changes inbetween but if those are all fine i will get this merged.

PR #2432, opened 2023-04-24
add NEVER_UNIT automation

smilerz:


cookbook/helper/ingredient_parser.py
```diff
@@ -50,10 +51,22 @@ def __init__(self, request, cache_mode, ignore_automations=False):
...
                     self.never_unit[a.param_1] = a.param_2
                 caches['default'].set(NEVER_UNIT_CACHE_KEY, self.never_unit, 30)
+
+            TRANSPOSE_WORDS_CACHE_KEY = f'automation_transpose_words_{self.request.space.pk}'
+            if c := caches['default'].get(TRANSPOSE_WORDS_CACHE_KEY, None):
+                self.transpose_words = c
+                caches['default'].touch(TRANSPOSE_WORDS_CACHE_KEY, 30)
+            else:
+                i = 0
+                for a in Automation.objects.filter(space=self.request.space, disabled=False, type=Automation.TRANSPOSE_WORDS).only('param_1', 'param_2').order_by('order').all():
```
vabene1111, line 61: is transpose_words and never_unit mixed up here ?

cookbook/helper/ingredient_parser.py
```diff
@@ -249,15 +262,31 @@ def apply_never_unit_automations(self, tokens):
...
-            return tokens
+            return ingredient
+
+        else:
+            if self.transpose_words:
+                for rule in self.transpose_words:
+                    ingredient = re.sub(rf"\b({rule[0]}) ({rule[1]})\b", r"\2 \1", ingredient)
+
+            else:
+                for rule in Automation.objects.filter(space=self.request.space, type=Automation.TRANSPOSE_WORDS, disabled=False).order_by('order'):
```
vabene1111, line 286: this hits the database for every ingredient parsed and might also execute extremly large amounts of regex

docs/features/automation.md
```diff
@@ -68,6 +68,17 @@ You can also create them manually by setting the following
 These rules are processed whenever you are importing recipes from websites or other apps
 and when using the simple ingredient input (shopping, recipe editor, ...).
 
+## Transpose Words
```
vabene1111, line 71: Is there also documentation for never unit ? or is that the same as transpose words?

cookbook/helper/ingredient_parser.py
smilerz: That would only occur if not cached - I can take that out and have it so the rule never runs if the cache doesn't exist.  Or do you thoughts on another way to limit it?

docs/features/automation.md
smilerz: they are both there...

cookbook/helper/ingredient_parser.py
vabene1111: iteresting, when i first reviewed there was somethign else, now it look correct

vabene1111:
ok so finally got this merged, thanks for your patience. I think one of the next things we need to do with automations is write some more/proper tests because i am slowly loosing oversight but other than that everything seems to be correct.

PR #2434, opened 2023-04-25
ldap auth: allow connecting using StartTLS

ssams:
Currently connctions to LDAP servers which require secure connections initiated via StartTLS are not possible. This change adds the environment option `AUTH_LDAP_START_TLS` and forwards it to the ldap provider (https://django-auth-ldap.readthedocs.io/en/latest/reference.html#auth-ldap-start-tls), so one can optionally enable StartTLS.


vabene1111:
thank you, will add it for the next release

Issue #2460, opened 2023-05-05
Recipesage import


Caesiumhydroxid:
### Tandoor Version

1.4.9

### Setup

Docker / Docker-Compose

### Reverse Proxy

No reverse proxy

### Bug description

The import from recipe-sage does not seem to work correctly for servings and preparation times. 

This is an example of a json object from the export:
```
{
    "@context": "http://schema.org",
    "@type": "Recipe",
    "datePublished": "2023-05-05T10:57:43.937Z",
    "description": "Schoko - Muffins - super schokoladig und schnell zubereitet, ergibt 12 Stück. Über 165 Bewertungen und für ausgezeichnet befunden. Mit ► Portionsrechner ► Kochbuch ► Video-Tipps!",
    "image": [
      "https://chefbook-prod.s3.us-west-2.amazonaws.com/1683284261803"
    ],
    "name": "Schoko-muffins - Schnell & Einfach",
    "prepTime": "PT60M",
    "recipeIngredient": [
      "70 g Butter , weiche",
      "200 g Zucker",
      "1 TL Vanillezucker",
      "1  Ei(er)",
      "210 ml Milch",
      "300 g Mehl",
      "2 TL Backpulver",
      "1 TL Salz",
      "80 g Kakaopulver , ungezuckertes",
      "75 g Schokoflocken oder Raspelschokolade",
      "evtl. Fett für die Form oder",
      "Papierförmchen"
    ],
    "recipeInstructions": [
      {
        "@type": "HowToStep",
        "text": "Die weiche Butter, den Zucker, den Vanillezucker und das Ei verquirlen. Die Milch dazurühren. In einer anderen Schüssel Mehl, Backpulver, Salz, Kakaopulver und Schokoflocken miteinander vermischen. Dann mit der Buttermasse verrühren."
      },
      {
        "@type": "HowToStep",
        "text": "Eine 12er Muffinform einfetten oder mit Förmchen auslegen und den Teig einfüllen. Bei 150°C Umluft ca. 20 min. backen."
      }
    ],
    "recipeYield": "1 Portion(en)",
    "totalTime": "PT15M",
    "recipeCategory": [],
    "creditText": "schnell & einfach",
    "isBasedOn": "https://www.chefkoch.de/rezepte/1403541245245072/Schoko-Muffins.html",
    "comment": [
      {
        "@type": "Comment",
        "name": "Author Notes",
        "text": ""
      }
    ]
  }
```
  
[...]

PR #2461, opened 2023-05-06
Fix recipesage time and yield import

Caesiumhydroxid:
This pull request should fix the error mentioned here: https://github.com/TandoorRecipes/recipes/issues/2460 it is tested on my system and seems to import recipe-sage (ld-json) just fine.


vabene1111:
thanks for the fix, will give it a try and get it merged hopefully for the next release

vabene1111:
i just reviewed the code: do you think we need another library for that? i already have a function to handle iso times i think, I would prefer not to add another libary just for that small function. 

If you dont want to I will try to take a look at implementing this function so that we dont need an additional library

vabene1111:
fixed this using the available functions

Issue #895, opened 2021-09-14
MealPlan Refactor


vabene1111:
- [ ] list of "i want to cook this in the near future" or maybe generally some kind of recommendation system 🤔 
- [ ] search
- [ ] add
- [ ] calendar export
- [ ] edit/detail modal
- [ ] plan type edit modal (also new fields like color, icon) decide if user or space scoped

old plan related issues
- [ ] nutrition (maybe as part of a big health feature set) #336
- [ ] type differentiation (probably fix with user/space unique constraint) #281
- [ ] remove via drag and drop (optional in my option depending on other features) #115

Kaibu:
Recipe search in the plan could be improved and rank higher if planned often/recently, instead of recommended list

phfeustel:
@vabene1111 is it possible now(v.1.0.2) or planned to have the meal plan shared with other users? I like to share the plan for one week with my family members who are logged in as users.

Kaibu:
@phfeustel I added recipe sharing and it's on dev already (https://github.com/TandoorRecipes/recipes/issues/1024). It's been a few very stressful weeks and we have not yet come around to another relesae.

It will be in the next Release for sure tho.

michaelszubartowicz:
Just curious, does your bonus feature include a rolling system where I could cycle through a list of recipes with a selected tag?
I have some knowledge regarding recommender systems. If you need help, hit me up ;)

vabene1111:
yes something like this is planned, we will make sure to get in touch if we hit a roadblock :) thanks!

balanceofcowards:
>  advanced but cool: automatic, tag/book/ingredient/whatever based meal planing

+1 for this. In my view, a randomized "fill the next x weeks with random recipes" would already be a great first step. Anything beyond that would be truly awesome!

PR #2468, opened 2023-05-18
Auto meal plan

AquaticLava:
This is a possible implementation of the auto based meal planning mentioned in issue #895. 

The auto planner creates a meal plan entry for every meal type from a given start day to a end day. 
Each meal type can have the possible random recipes restricted by inputting keywords, only meals that mach all inputted keywords will be used.
The meal entries are generated asynchronously.

This is what the GUI currently looks like,
image

Between the time i started working on this and now, the interface for the meal plan tab has changed, I am not sure where the button for auto meal planning should go, i added one next to the create button. There is currently no way to access it in the mobile size of the interface.

Let me know of any changes needed!


AquaticLava:
Perhaps we could have a api endpoint that returns a random recipe? You would input a list of keywords and would get a random recipe back. This is my first time working with python and Django, so I'll have to do some reading on it. What files would i need to change to add something like that?

Right now if a meal type is empty recipes are selected randomly. It should be a easy change to either not plan when there are no keywords or to add a check box. Do you have a preference?

Also do you think that the buttons on the bottom of the meal planner would need to be reworked? This is how they look currently.
Screenshot_20230519_191946

vabene1111:
I personally think the meal planner ui is far from good. It's ok and does it's job but many parts could be much more user friendly.

If you don't feel like you can do it I can build the Api endpoint, if you want to try your luck you will need a serializer that handles a lots of obkety that contain the keywords for each meal type and if it should be planned.

And then you would need to add an additional route/method to the MealPlanEntryViewSet similar to the image function in recipeviewset that takes this data and does the planning.

AquaticLava:
I'll take my chances at building the endpoint! I'll start by reading the Django docs and reading the the MealPlanEntryViewSet and recipeviewset. If i have any farther questions ill let you know! 

Would it be helpful to change this to a draft PR?

I defiantly understand the difficulty of UI design, i would say its what i am weakest at. What i tend to think looks good is old 90s websites.... At least I'm at the point where if a designer makes a mock up i can build it!

vabene1111:
Awesome, you are exactly at the same place where you I am regarding design 🤣

You can change it to draft but it's OK like this as well. Let me know if you need help.

AquaticLava:
I've been looking at the documentation as well as the source code, i think i have a general idea of how it works. I'm struggling with exactly how to implement my ideas. 

my general plan is to have the client send the server a start date, end date, meal type, keyword list, default servings, and the shared array. I think that this is the minimum amount of information needed to create all the meal plan entries.
The server would then revive the data and turn it into python data types using the AutoMealPlanSerializer. 

From what i have been reading in the source code it looks like all the other sterilizers have a model in the meta class. My understanding is that the model allows the sterilizer to create a object that can then be used in the view sets. In my case the auto meal plan object would create multiple meal plans. 
Do i need to create a class that defines a auto meal plan object that the MealPlanViewSet method that i create uses to do all the planning? if so, where would i put the class?

To create the random recipes i would just create a new queryset using the keywords that were passed from the client.

vabene1111:
i understand that this is confusing, took me a while to understand too. Serializers are used by django rest framework to perform validation on API calls. You can read more about that here https://www.django-rest-framework.org/tutorial/1-serialization/

You do not need a serializer but it might not be the worst idea for this endpoint but serializers can exist wihtout models. In the api.py file there are a few endpoints (below the model ones) that do not use models and just "manually" accept data. Those should be a good starting point for you. 

You bascially need to accept these parameters (which sound correct to me) and then efficently create the entries. 

For that i would recommend loading a set of recipes matching your search criteria (keywords) ordered randomly (order_by('-?') if i remember correctly) and possibly limit it to a certain size ([:<number of days to plan>]) so you dont query to much data. 

After that you create the MealPlanEntry objects and use bulk_create() to insert them all into the meal plan. You could then return those entries but that is extra work so i would just redirect to the meal plan entry view set filtered to your start/end date. 

sorry for the late response, feel free to contact me if you need further help.

AquaticLava:
Im getting closer! I got a object to be written to the database! I didn't realize that the object creation was inherited from the rest api's sterilizer, inheritance can be confusing. 

When i passed the meal type from the frontend to the MealTypeSerializer it did not create a MealType object, just a ordered dictianry. Do you know why that might be? I switched to using the id and pulling it from the database.

For the meal plan objects, does the bulk_create work? in the documentation it says that bulk create does not work with many-to-many relationships, and shared is a many to many field.

I was reading about getting random objects from the database and found this stack overflow thread: https://stackoverflow.com/questions/22816704/django-get-a-random-object. Do you know if our supported database backends are slow when returning a random order?

On the front end do i need a new method in the ApiApiFactory to access the new api endpoint? Also do i need to create a new entry in the url conf? I tried to use the decorators.action like in the image method but the url it generates is only acessible on a instance. What i want to do is make the url be /api/meal-plan/auto-plan, with the decorators.action it makes the url be /api/meal-plan/<pk>/auto-plan.

Thanks for all the help! I'm sure you could have gotten this all done in a few hours haha.

vabene1111:
> When i passed the meal type from the frontend to the MealTypeSerializer it did not create a MealType object, just a ordered dictianry. Do you know why that might be? I switched to using the id and pulling it from the database.

Not sure what exactly you mean. I dont think that its correct to use a model Serializer, i think you need to use a normal serializer

> For the meal plan objects, does the bulk_create work? in the documentation it says that bulk create does not work with many-to-many relationships, and shared is a many to many field.

this should not be a problem from my experience but i am not 100% sure, i would say try it out, if it does not work we can bulk add the relation using the trough manager

> I was reading about getting random objects from the database and found this stack overflow thread: https://stackoverflow.com/questions/22816704/django-get-a-random-object. Do you know if our supported database backends are slow when returning a random order?

you can use random, since its only one call and you take the slice for the amount of recipes needed it should be efficent

> On the front end do i need a new method in the ApiApiFactory to access the new api endpoint? Also do i need to create a new entry in the url conf? I tried to use the decorators.action like in the image method but the url it generates is only acessible on a instance. What i want to do is make the url be /api/meal-plan/auto-plan, with the decorators.action it makes the url be /api/meal-plan/{pk}/auto-plan.

the method is generated automatically by the openapi generator (see docs here https://docs.tandoor.dev/contribute/#api-client). After you create the ViewSet (copy one of the non model ones) you need to add it to the urls (basically just copy that as well :D) after that generate the openapi and the method/endpoint will be there

> Thanks for all the help! I'm sure you could have gotten this all done in a few hours haha.

probably but you wanted to do it and you are honestly trying and not just asking dumb random questions, i always love to support an honest effort. feel free to ask more questions if you want and dont worry if the result isn't perfect, i can fine tune it in the end if necessary.

AquaticLava:
Latest commits implement the backend endpoint. I have left the debug buttons in to help with testing, i will remove them when we are ready to merge but i suspect there will be changes i have to make! 
I think that with adding bulk creation of meal entry bulk deletion and editing will be needed... I think that should go in its own PR though. 

Right now auto adding recipes to the shopping cart does not work when it is set by the check box. sharing also does not currently work.

auto adding is not working because i am missing the addshopping in the request data, I cant figure out how that variable gets into the request.

It looks like sharing doesn't work because the object has not been made in the database yet so it has no id.

vabene1111:
> I think that with adding bulk creation of meal entry bulk deletion and editing will be needed... I think that should go in its own PR though.

yes maybe we should add a field (auto_plan_id) or something so you are able to quickly delete all entries created by an auto plan, but as you say, that can be added with future PR's

> auto adding is not working because i am missing the addshopping in the request data, I cant figure out how that variable gets into the request.
that is something super hacky and something i forget every time, can look it up or just add it when merging. its also not great from a performance aspect as it will run for each recipe so we might want to add a limit of how many meals you can auto plan. 

> It looks like sharing doesn't work because the object has not been made in the database yet so it has no id.
i will see if i spot that when reviewing the changes

generally thank you very much, will try to review your code soon to give you some feedback.

cookbook/views/api.py
```diff
@@ -661,6 +665,57 @@ def get_queryset(self):
...
+
+            for keyword in keywords:
+                recipes = recipes.filter(keywords__name=keyword['name'])
+
+            recipes = recipes.order_by('?')[:days]
+            recipes = list(recipes)
+
+            for i in range(0, days):
+                day = start_date + datetime.timedelta(i)
+                recipe = random.choice(recipes)
```
vabene1111, line 691: if i read this correctly this leaves the chance for a recipe to be choosen twice in the planning period, right?

I think I would just take the index of the day since the recipes are ordered randomly anyway so recipe = recipes[i]

cookbook/views/api.py
```diff
@@ -661,6 +665,57 @@ def get_queryset(self):
...
+                        'created_by': request.user,
+                        'meal_type': meal_type,
+                        'note': '', 'date': day, 'space': request.space}
+
+                m = MealPlan(**args)
+                meal_plans.append(m)
+
+            MealPlan.objects.bulk_create(meal_plans)
+            for m in meal_plans:
+                if request.data.get('addshopping', False) and request.data.get('recipe', None):
```
vabene1111, line 702: i need to look at how the shopping stuff works but its very confusing so i will do it.

vabene1111 (commented):
I really like how you solved this, props for someone learning this its well written and seems to be overall correct. 

I added 2 small comments and i would like to ask you to remove api.ts from the PR so i can merge it easier (i can just regenerate it once its merged), other than that it looks good. 

I will wait for a few days or possibly weeks for 1.5 to become stable before merging because I want this feature in beta for a while to make sure it works well but i cant do that (at least easily) if i am still pushing hotfixes for the last major release, so dont be confused if a few days go by without anything happening.

AquaticLava:
finally got some more time to work on this! The shopping cart and sharing now work! I think it is ready for a final review. If you don't see any problems then I'll remove the api.ts as well as the debug buttons and we can get it merged!

vabene1111:
awesome, thank you. The last few weeks have been very stressfull at work but i will try to get this out asap (probably going to do one more bugfix release for 1.5 first tough and then have this on beta for a while for testing)

vabene1111:
ok so thanks a lot for this PR, finally got around to review it and get it merged. Will do some minor tweaks for performance but overall very happy with it. 

Thanks for the effort put into this and the good work together.

PR #2470, opened 2023-05-19
Enable gunicorn debug log

gloriousDan:
While helping to debug some gunicorn startup issues I realized that the gunicorn log level was not yet configurable. So I modified the .env file and boot.sh script to make the gunicorn loglevel configurable by setting the `GUNICORN_LOG_LEVEL` variable. 

We could maybe also use the DEBUG variable to modify the gunicorn log level as well I'm open to both possibilities.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -3,6 +3,9 @@
 DEBUG=0
 SQL_DEBUG=0
 DEBUG_TOOLBAR=0
+# Gunicorn log level for debugging (default value is "info" when unset)
+# (see https://docs.gunicorn.org/en/stable/settings.html#loglevel for available settings)
+# GUNICORN_LOG_LEVEL="debug"
 
 # HTTP port to bind to
 # TANDOOR_PORT=8080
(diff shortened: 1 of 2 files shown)


vabene1111:
thanks for the PR, have you tested this with the container?

I think having it as an extra variable makes sense but maybe we place it next to the other debug variables in the .env template? 

i really need to wirte a proper docs article about all the environment variables and its considerations ...

gloriousDan:
I tested it and with the default setting it indeed behaves as before. With "debug" additional info is printed by gunicorn. I szill have to test what happens when a invalid value is set and how if the other settings work. I also built the docker image (wirthout node though) under: gloriousdan/recipes-debug 

But I'll do some further testing and move the env variable to the section at the start and then get back to you.

vabene1111:
Thanks, this sounds tested enough to me :)

gloriousDan:
I now tested it with an invalid value and gunicorn starts just fine with the default logging value. I also moved the env variable to the top of the env file and added some more info about it.

PR #2471, opened 2023-05-19
Feature: DISABLE_INTERNAL_AUTH

SpeedyFireCyclone:
This change exposes a new environment variable `DISABLE_INTERNAL_AUTH`, which hides the login form if set to 1, and if a social login provider is found.
If no other provider is found, the behaviour is identical to what it is currently to prevent bad configurations.


vabene1111:
Hi, thanks for the pr. While hiding the form sounds good on first thought it does not really disable password login and might give people a wrong feeling of security. The endpoints in the back end, both for the form and also the Api password login are still available and could be used.

PR #2478, opened 2023-05-29
Fix bottom navigation not hiding in print mode

jwr1:
Currently, the mobile navigation bar shows up at the bottom of each page when printing, as shown below:
Screenshot_20230529_114821


vabene1111:
thanks, i forgot that

Issue #2479, opened 2023-05-31
Add step for venv activation


nabim777:
### Documentation link

https://docs.tandoor.dev/install/manual/

### Affected section

Installation

### Issue description

Since the virtual environment is not activated after creation, we can add code 
source bin/activate.

imharyohm:
Hey, please assign this to me

The-Priya-Saw:
hello @TandoorRecipes, I would like to solve this issue. can u please assign me ?

vabene1111:
feel free to take up this issue :)

PR #2480, opened 2023-05-31
Add step for venv activation  #2479 solved.

imharyohm:
As suggested I had added that command for activation of Virtual environment.


vabene1111:
sorry but this does not make sense, you cannot run install on activate. 

you need to run activate first with `source` and then run pip. 

Please re-open once fixed

PR #2488, opened 2023-06-02
add admin command to delete unattached ingredients and steps

smilerz:
creates two admin actions:
- delete all Ingredients not part of a recipe
- delete all Steps not part of a recipe

### DIFF
--- a/cookbook/admin.py
+++ b/cookbook/admin.py
@@ -10,12 +10,13 @@ from treebeard.forms import movenodeform_factory
 
 from cookbook.managers import DICTIONARY
 
-from .models import (BookmarkletImport, Comment, CookLog, Food, FoodInheritField, ImportLog,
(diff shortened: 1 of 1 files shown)


tomtjes:
@smilerz Could the same be done for foods?

smilerz:
> @smilerz Could the same be done for foods, units and keywords?

theoretically - but I wouldn't implement it.  There are valid scenarios where food and keywords especially, are intentionally unused (root of hierarchy in my personal use case) - allowing mass deletion in those cases would likely lead to unintended outcomes.

tomtjes:
That makes sense. I have a hacky script that deletes unused keywords and foods (and checks that they don't have children in the hierarchy) via the API, and had hoped for a simpler solution to get rid of clutter that inevitably comes with importing web recipes.

smilerz:
Eventually we’ll add filters and multi-edit capabilities to the management page.

PR #2517, opened 2023-06-26
Create Truenas-Portainer

16cdlogan:
Install Tandoor Recipes on TrueNAS Core and Portainer


16cdlogan:
Pardon any formatting issues.

vabene1111:
thank you, i will add this to the docs with the next release. I have only looked over it and it seems to be correct at least the part that I can judge since i do not know much about portainer or true nas

PR #2521, opened 2023-06-26
add note to docker-compose files and update postgres tag

gloriousDan:
These are some small Improvements to the docker-compose files in the docs:

1. Postgres version 11 is the default in the docker-compose files and its support will end in [November](https://www.postgresql.org/support/versioning/). Is there any reason against updating it to the most recent version?
2. Since many support requests in the discord boil down to the fact that the nginx config volume was changed to a bind mount and nginx can't find its config anymore, I added a small disclaimer which should make it clear that it shouldn't be changed without consideration.

### DIFF
--- a/docs/install/docker/ipv6_plain/docker-compose.yml
+++ b/docs/install/docker/ipv6_plain/docker-compose.yml
@@ -2,7 +2,7 @@ version: "2.4"
 services:
   db_recipes:
     restart: always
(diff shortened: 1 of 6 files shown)


vabene1111:
Perfect, wanted to do this for a while but haven't gotten around to it. Will get this merged for the next Version

PR #2531, opened 2023-07-02
Add source URL to recipe export

michael-genson:
This adds the `source_url` field to the recipe export. It's present in other export types (e.g. Paprika) but missing from the default export.


vabene1111:
perfect, thank you

PR #2558, opened 2023-07-24
Optionally hide step ingredients

srwareham:
WIP PR to add ability to show or hide the recipes table for any given step of a recipe. 

Currently working:
* Toggle whether an individual step has the ingredients shown using the edit recipe 3 dots menu

Incomplete:
* User-level default setting for whether or not ingredients should be shown on the step level. Backend database & user settings frontend has been updated, however the Step model hasn't been updated to respect this change
* Refactoring of implementation to combine the two field used for determining whether the view step ingredient table component should be displayed. Target design dependent on solution to above Step model change


srwareham:
Wrong branch, apologies

PR #2591, opened 2023-08-24
Tool/Equipment list

Knalltuete5000:
This is the PR to introduce the tool/equipment list discussed in #2062 

Please let me know if anything is missing/should be renamed or anything else


vabene1111:
hi, thanks for your PR. Sadly I do not think that I can merge this. The reason is that this adds a HUGE amount of complexity for a feature that is likely only going to be used to its full percentage by a very very small amount of users. Please see the note here https://docs.tandoor.dev/contribute/ 

I need to think about this a bit more but I am already struggling to keep tandoor maintained with the other hugely complex features and its already a huge problem for users to actually understand what to do with tandoor, i am not sure this is the way to go, altough I love the Idea of a powerful equipment system. 

Maybe if we can reduce some of the complexity there would be a way to get this merged, but for that I first need to understand what you actually did here.

Issue #2261, opened 2023-01-10
Option to disable Sign-Up / Registration


besendorf:
### Is your feature request related to a problem? Please describe.

With the invite links for spaces, any user can invite other users to the tandoor instance.  This is very problematic for more advanced "enterprise" setups. Especially when one uses LDAP or OIDC a common use case is to disable registration for local users and only allow registration through the Authentication provider to get a single point of truth for the user database and also enforce certain authentication aspects through the provider like 2FA or a password policy.

### Describe the solution you'd like

A configuration option e.g. a ENV variable to disable local sign-ups. Ideally there should also be a more straight forward way to invite already users to a space. I would suggest just entering the username in a field with autocompletion. 

### Describe alternatives you've considered

The only alternative right now is blocking the invite endpoint in the reverse proxy

vabene1111:
to be honest supporting enterprise grade deployments with LDAP is not supported and possible not covered by the license of tandoor. 

Feel free to get in touch with me if you have any enterprise specific needs and we can surely talk about how they could be achieved but those features are not the focus of my free/open source development as these kinds of features don't provide value to the majority of users. 

You can reach me at info@tandoor.dev

besendorf:
Thanks for the response. I understand that this is not the main focus of the project. However I think you missunderstood my use of "enterprise" I don't won't to start a business with a tandoor server but rather my friends and I host several services in a more professional environment where we already use keycloak for authentication.

Also semi-open registrations not ideal from a security point of view. When an adversary gains access to an invite link they could easily create an unlimited number of accounts.

As I said I understand that it is probably not a use case you had in mind I want to ask if that would be something I could help out with and maybe submit a PR. How would you design such a option?

vabene1111:
ok understood.

But what is it that you have with invite links, only a user who has administrator priveledges and owns a space can create invite links, so if no one has these kinds of permission in your space they cannot invite anyone?

or do you see it as a problem that people could create their own space and invite users into that? Maybe we should add a setting to allow or disable autonomous space creation 🤔 

We could also add a setting to allow or forbid the creation of invite links or maybe on a per space level 🤔 

Not sure what the best way from a design perspective is here, to really solve this properly one would need a much more in depth permission model for users. There are several "short term" solutions but I dont want to add to much complexity with for a band aid, what do you think?

[...]

PR #2593, opened 2023-08-28
allow signup for social accounts when the provider is set up

BrainWart:
When a social provider is set up, the social account signup view is enabled. This seems to cover issue #2261, but it needs further testing.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -113,7 +113,8 @@ REMOTE_USER_AUTH=0
 # SPACE_DEFAULT_MAX_FILES=0 # Maximum file storage for space in MB. 0 for unlimited, -1 to disable file upload.
 # SPACE_DEFAULT_ALLOW_SHARING=1 # Allow users to share recipes with public links
 
-# allow people to create accounts on your application instance (without an invite link)
+# allow people to create local accounts on your application instance (without an invite link)
+# social accounts will always be able to sign up
 # when unset: 0 (false)
 # ENABLE_SIGNUP=0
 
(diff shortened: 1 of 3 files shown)


vabene1111:
thanks for your PR. this changes the logic for existing setups but I think this is a rather expected behavior so we can probably safely change it. 

To get this merged we need to 
- verify if there are any tests testing this behavior, maybe add some if not
- explain the behavior in .env.template
- explain the behavior in the docs if there is any applicable section

if you want you could do that, if not I will need to do it when I have time and merge it then,

BrainWart:
Thank you for creating this project!


> To get this merged we need to 
> - verify if there are any tests testing this behavior, maybe add some if not

I am not sure if I will be able to handle this, but I am happy to give it a shot. I am not really much of a Python developer.

> - explain the behavior in .env.template

Will do

> - explain the behavior in the docs if there is any applicable section

I can do this as well.

Thank again!

BrainWart:
@vabene1111,
I did not see any relevant tests or any easy way to ensure the social sign up buttons appear under the expected conditions. If possible, could you provide some guidance for testing for this change?

vabene1111:
hi @BrainWart thanks for making the changes. I am not entirely sure how to do the test, I would probably test the function itself not the view/html, but dont worry about that if this is not your main thing, I can look at it at the time of merging (or decide I am to lazy 😂)

BrainWart:
Sounds good. I appreciate the work you're doing with this project. Thank you

Issue #2514, opened 2023-06-24
Importing a community curated list leads to an error 


panoskpv:
### Tandoor Version

Latest

### Setup

Docker / Docker-Compose

### Reverse Proxy

No reverse proxy

### Bug description

When I try to import a community list in greek (el) I get the following error 'There was an error creating a resource!' and nothing is added. When I try to import english version of this list I get the same error, foods, categories etc. are added but foods are added without nutrion data.

vabene1111:
can you please post some logs from your container so that i can see whats breaking for you?

panoskpv:
After your comment i checked the logs and i got dublicate slug error. I removed all added items and for those i wanted to keep i just deleted the open slug field in admin settings. Now i get the below:

{'selected_version': 'el', 'selected_datatypes': ['food', 'unit', 'category', 'property', 'store', 'conversion'], 'update_existing': True, 'use_metric': True}

Internal Server Error: /api-import-open-data/

Traceback (most recent call last):

  File "/opt/recipes/venv/lib/python3.10/site-packages/django/db/backends/base/base.py", line 306, in _commit

    return self.connection.commit()

psycopg2.errors.ForeignKeyViolation: insert or update on table "cookbook_foodproperty" violates foreign key constraint "cookbook_food_properties_food_id_4c533fef_fk_cookbook_food_id"

DETAIL:  Key (food_id)=(211) is not present in table "cookbook_food".

The above exception was the direct cause of the following exception:

Traceback (most recent call last):

  File "/opt/recipes/venv/lib/python3.10/site-packages/django/core/handlers/exception.py", line 56, in inner

    response = get_response(request)

  File "/opt/recipes/venv/lib/python3.10/site-packages/django/core/handlers/base.py", line 197, in _get_response

    response = wrapped_callback(request, *callback_args, **callback_kwargs)

  File "/opt/recipes/venv/lib/python3.10/site-packages/django/views/decorators/csrf.py", line 55, in wrapped_view

    return view_func(*args, **kwargs)

  File "/opt/recipes/venv/lib/python3.10/site-packages/django/views/generic/base.py", line 103, in view

    return self.dispatch(request, *args, **kwargs)

  File "/opt/recipes/venv/lib/python3.10/site-packages/rest_framework/views.py", line 509, in dispatch

    response = self.handle_exception(exc)

  File "/opt/recipes/venv/lib/python3.10/site-packages/rest_framework/views.py", line 469, in handle_exception

[...]

PR #2595, opened 2023-08-28
#2514 Importing a community curated list leads to an error

djstini:
This adds error Handling for missing food-keys in the open_data_importer slug_id_cache and consequently solves the Bug that caused imports of OpenData lists to fail.

Related Issue is #2514


vabene1111:
thank you very much for the PR, i will merge this as a temporary fix but need to find a proper solution (why is the key missing) in the long run.

djstini:
had some time to dig deeper. The issue happens during then checking if a "Food" already exists, the "IF" in L.130.

Here we validate if the "name" or "plural_name" from the open data file is included in a previously assembled "existing_objects_flat" list.

This works okay, as long as all the names and plural_names are unique.

In the Translations this is lost.

For example:
| key | name - GER |
| food-cilantro-raw | Koriander |
| food-spices-coriander-seed | Koriander |

Since this exists as a duplicate, the second occurrence will not be added to the insert_list, therefore never being created, therefore not existing when trying to add the properties ( at the place where i added the try catch. )

In German the Error is only for the 3 Foods, i will submit a change to the open data List fixing that soon.

I'd recommend leaving in the try-catch, as this is a problem that can occur in the future, since the lists are constantly subject to change.

PR #2624, opened 2023-09-08
Update and add settings

WoosterInitiative:
I've been fighting with CORS and CSRF (amongst other things) while trying to get an API call with ajax to work. As such, I found that `CORS_ORIGIN_ALLOW_ALL` was fixed as `True`. I feel that it should be configurable by the user, so I added an `os.getenv` call for it with a default to `True` for no breaking changes. I also updated it per https://github.com/adamchainz/django-cors-headers to `CORS_ALLOW_ALL_ORIGINS`.

I added that and `CSRF_TRUSTED_ORIGINS` to the template .env file for better discoverability, though they're both commented out.

I didn't do a discussion on this first, but I can start one if that would be preferred.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -13,6 +13,14 @@ DEBUG_TOOLBAR=0
 # hosts the application can run under e.g. recipes.mydomain.com,cooking.mydomain.com,...
 ALLOWED_HOSTS=*
 
+# Cross Site Request Forgery protection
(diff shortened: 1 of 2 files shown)


vabene1111:
Thanks, no worries about a discussion, this does not add any more complexity so it's fine. Will review it and likely get it merged.


Thanks for the pr

PR #2630, opened 2023-09-13
Imports cleanup

smilerz:
code cleanup, no functionality changes

- removed unused imports
- removed variables assigned but never used
- removed commented out code that was replaced with something newer


vabene1111 (changes requested):

smilerz:
Honestly, I'm not at all certain what this file does. 

I'm going to do another change that deprecates dal so I can look at this init (and there's one more) to see if the current version of django uses them at all.

PR #2632, opened 2023-09-14
updates to multiple models uniqueness capabilities

smilerz:
make 'name' unique in space for MealType
make MealType, Unit, Supermarket, Supermarket Category, PropoertyType case insenstive for get_or_create
convert unit conversion create serializer to get_or_create behavior 
enable create duplicate tests for unitconversion and mealtype api


vabene1111:
Thanks, the code looks good to me. One question/consideration:

I know there is a really old issue somewhere discussing if the user would expect the API to change the case sensitity. I think back then I was against it but today, thinking about it, i am leaning towards your Approach. Some feedback would be welcome here.

smilerz:
Worth asking on discord.  But I think most users expect Liter and liter to be equivalent. Same with chicken thighs and Chicken thighs and Chicken Thighs. 

Having these end up as multiple items leads to more unexpected behavior than assuming users intend for them to be different.

Issue #455, opened 2021-02-24
Ingredient Notes Rework


hotdog218:
**Is your feature request related to a problem? Please describe.**
It's kind of a pain when cooking to hover over the comment icon to see the note. 

**Describe the solution you'd like**
It would be cool to have notes that are under x amount of characters just display text. For example, if you just want to specify "softened" for butter, then it would just show up as "butter, _softened_". But if you want to say something longer, like "Make sure the bananas are very ripe", then the current hover system would be used.

**Describe alternatives you've considered**
Text wrapping could also be a viable solution. Maybe a user setting that will toggle between hovering and wrapping.

**Additional context**
Mostly a QoL feature, but for me, the minimal amount of interaction I can make with the computer while cooking, the better.

Example:
![3f2b49b541d166d17e2f24751382448e](https://user-images.githubusercontent.com/24951087/108925631-3cf69980-75fa-11eb-8483-1b30061d7798.png)

vabene1111:
this is a cool idea, i have a big mix of long and short notes as well .. it coudl also be a setting i guess 🤔 definitely something i want to do

Szeraax:
I'd suggest only implementing this or #661.

vabene1111:
i will look at both of those options when i get to this, thanks for the link

Vishrut19:
@smilerz Do you need any help or assistance ? I can help you out if you are interested

vabene1111:
Feel free to give it a try, it's not super complicated but the solution needs to be well thought trough.

Vishrut19:
Hi @vabene1111 I will try to solve it and also I have joined the discord server.Hope you and other people will help me and support. Since this is my first open source contribution please help me out if I need any help.

vabene1111:
Sure, feel free to ask. Also please share a rough draft of what you want to do before implementing anything.

xmayukx:
Hello, @hotdog218 I can work on this issue, could you please assign it to me?

hotdog218:
> Hello, @hotdog218 I can work on this issue, could you please assign it to me?

I don't believe I have the ability to assign anything, plus I think @pedro-camargo11 is already handling this- thanks by the way!

smilerz:
> Hello, @hotdog218 I can work on this issue, could you please assign it to me?

No need to have it assigned to you.  If you have thoughts on how to address you can describe your solution here. Or submit a PR and it’ll be reviewed for merge.

Royal-Dragon:
[...]

PR #2643, opened 2023-09-27
Display short ingredient notes without hover

nidhi-badiger:
If ingredient notes are longer than 15 characters, continue to display the notes with the current hover system.
Else, display the ingredient note as regular text.


vabene1111:
thanks for the PR. #455

would love to get some feedback from the community here, do you think this should be merged as or do we need a setting for this?

hotdog218:
I'd be ok without the setting, but I'm biased, as I'm the guy that made the feature request.

PR #2647, opened 2023-09-30
Added support for keeping SECRET_KEY and POSTGRES_PASSWORD in a file

JohnTheNerd:
This commit adds two optional environment variables:

- SECRET_KEY_FILE
- POSTGRES_PASSWORD_FILE

This change allows mounting secret data when running this in Docker Swarm, instead of having to hard-code it in our docker-compose file or provide it alongside all other environment variables.


vabene1111:
Thanks for the pr, I see this being useful to some people who for whatever reason cannot use the environment.

Will get this merged for a nice long beta period for testing

PR #2672, opened 2023-10-07
Make import list layout responsive

harry48225:
The page showing the list of apps you can import recipes from could work better on small screens. Here's a gif of the current behaviour:
current behaviour

I have added some breakpoints to the grid of apps so that they are easier to read on smaller screens.
new behaviour

Now the names of the apps can be easily read on small screens. However when all the apps are in such a long list it might be confusing for the user to have to scroll down all the way to the bottom to import the file. I think also that maybe when pressing an app the file chooser should open immediately and the bottom section of the page should be removed? Or maybe the bottom part opens as a modal? Despite this potential confusion I think the layout change is still preferable to the current behaviour as right now this page is almost unusable on a phone.

### DIFF
--- a/vue/src/apps/ImportView/ImportView.vue
+++ b/vue/src/apps/ImportView/ImportView.vue
@@ -267,10 +267,10 @@
                         <b-tab v-bind:title="$t('App')">
                             <b-container>
                                 <h4>{{ $t('Select_App_To_Import') }}:</h4>
-                                <b-row class="mt-4">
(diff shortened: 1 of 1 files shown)


vabene1111:
hi, thanks for the PR, the change is definitely better than before but I agree with your conclusion that it is not very intuitive on mobile .

As with some other pages i think we need to change to mobile version in the long run to use some kind of modal/selection promt. 

will merge this for now

Issue #2676, opened 2023-10-08
error when using "de" in "kuratierte Liste von Lebensmitteln, Einheiten und mehr importieren"


BstAA:
### Tandoor Version

commit d42d784aeb0595bbc0ed47eb8bbc27ba61b0787d Author: vabene1111  Date:   Tue Aug 29 13:09:38 2023 +0200

### Setup

Docker / Docker-Compose

### Reverse Proxy

Traefik

### Bug description

When the above mentioned import works o.k. with "en-US" selection, but gives an error 

`Beim Erstellen einer Ressource ist ein Fehler aufgetreten!`

when selecting "de" i.e. German values

sthuber90:
I observed the same thing with the current latest docker image. I'm using docker compose as well and nginx as proxy

djstini:
It's an older Problem... i wrote a fix for it which allows the ingredients to be created.. the problem is that the properties cant be assigned properly now.

The reason for the problem was that the mapping works by name, which makes sense,, if you already have an ingredient called "Sauerkraut" you wouldnt want to overwrite it.

Now in the Germab translation a few of the ingredients that have different names in english use the same translation in German.

For Examples check the pull request from back then: #2595 

Related issue: #2514

I have time today so imma check out what the problem is now.

BstAA:
> I have time today so imma check out what the problem is now.

THX! I am so new to the project that I can't be a real help

djstini:
No Problem, it seems i overlooked in #2595 that if the list was previously not imported the error would still occur.

PR #2682, opened 2023-10-13
open-data-import further duplicate handling

djstini:
**Related to**
#2514
#2595 
#2676 
https://github.com/TandoorRecipes/open-tandoor-data/pull/11

The Problems Source is, the same as in #2595. The German translation of the Import File has duplicate entries for the different Foods.

The Fix i added in #2595 works and is still required for updating already imported lists. But i overlooked, that if the list was never previously imported the import would also fail trying to insert the duplicates.

This aims to fix that by introducing another check "if the food with name is already in the insert_list"

### DIFF
--- a/cookbook/helper/open_data_importer.py
+++ b/cookbook/helper/open_data_importer.py
@@ -117,18 +117,24 @@ class OpenDataImporter:
         self._update_slug_cache(PropertyType, 'property')
 
         insert_list = []
+        insert_list_flat = []
         update_list = []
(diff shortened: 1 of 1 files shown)


vabene1111:
perfect, thanks for the fix. 

I am thinking about changing how the importer works, Maybe I need to load everything that is in the database, do the processing and then dump back into the DB only the ones that changed. But that will be a lot of refactoring ...

PR #2685, opened 2023-10-18
Increase specificity of the delete confirmation dialog

harry48225:
When deleting steps and ingredients I was sometimes unsure which delete button I'd actually pressed. I've tweaked the strings to now tell the user what they're about to delete.

e.g. 
- Steps
  - without a name
    `Are you sure you want to delete this Step?` |-> `Are you sure that you want to delete Step "2"`
  - with a name
    `Are you sure you want to delete this Step?` |-> `Are you sure that you want to delete Step "Prepare vegetables"`

- Ingredients
  - without a name
    `Are you sure you want to delete this Ingredient?` (unchanged)
  - with a name
    `Are you sure you want to delete this Ingredient?` |-> `Are you sure that you want to delete "cloves"`


vabene1111:
thanks for the PR, that makes things more clear

PR #2691, opened 2023-10-21
Update authentication.md

gorrilla10101:
Changed provider list url because existing one doesn't work anymore.


vabene1111:
thank you 👍

PR #2693, opened 2023-10-22
Update to import tags on mealie and chowdown recipes

blowk:
This PR can import Mealie tags so they are imported too.


blowk:
Let me know if I need to change it.

blowk:
Add update to import extra information on chowdown recipes

vabene1111:
thanks for your PR, I will review it and get it in for either the next or the release after that. Sorry for the long wait, had some other stuff to take care of for a few weeks.

PR #2722, opened 2023-11-06
Make gunicorn and nginx listen to IPv6

swnf:
## Problem

I'm using tandoor in a docker compose setup. I noticed that nginx sometimes returns a "Bad Gateway" error and logs something like `upstream server temporarily disabled while connecting to upstream`.

Although I'm not entirely sure, I think this might be caused by my IPv4/IPv6 dual stack setup (I've enabled ipv6 on the network created by docker compose). I assume the DNS request from the nginx container sometimes resolves to the IPv6 address of the gunicorn container. The gunicorn container does not listen to IPv6 connections. Therefore, the connection is refused and nginx returns a gateway error.

## Solution

This PR binds gunicorn to `[::]:8080` instead of `:8080`. This should listen on [both IPv4 and IPv6](https://github.com/benoitc/gunicorn/issues/1138#issuecomment-158358666) addresses (as long as `bindv6only` is disabled, which is disabled by default).

I additionally enabled IPv6 for nginx itself.

### DIFF
--- a/boot.sh
+++ b/boot.sh
@@ -76,4 +76,4 @@ echo "Done"
 
 chmod -R 755 /opt/recipes/mediafiles
 
-exec gunicorn -b :$TANDOOR_PORT --workers $GUNICORN_WORKERS --threads $GUNICORN_THREADS --access-logfile - --error-logfile - --log-level $GUNICORN_LOG_LEVEL recipes.wsgi
+exec gunicorn -b "[::]:$TANDOOR_PORT" --workers $GUNICORN_WORKERS --threads $GUNICORN_THREADS --access-logfile - --error-logfile - --log-level $GUNICORN_LOG_LEVEL recipes.wsgi
--- a/nginx/conf.d/Recipes.conf
+++ b/nginx/conf.d/Recipes.conf
@@ -1,5 +1,6 @@
 server {
   listen 80;
+  listen [::]:80 ipv6only=on;
   server_name localhost;
 
   client_max_body_size 128M;


vabene1111:
interesting, I thought we had done this a while ago, thanks for the PR, will check on it and get it merged.

bositman:
This seems to break gunicorn on systems with ipv6 disabled:
( 
```
cat /sys/module/ipv6/parameters/disable
1
```)
```
[2024-02-17 21:44:50 +0000] [1] [INFO] Starting gunicorn 20.1.0
[2024-02-17 21:44:50 +0000] [1] [ERROR] Retrying in 1 second.
[2024-02-17 21:44:51 +0000] [1] [ERROR] Retrying in 1 second.
[2024-02-17 21:44:52 +0000] [1] [ERROR] Retrying in 1 second.
[2024-02-17 21:44:53 +0000] [1] [ERROR] Retrying in 1 second.
[2024-02-17 21:44:54 +0000] [1] [ERROR] Retrying in 1 second.
[2024-02-17 21:44:55 +0000] [1] [ERROR] Can't connect to ('::', 8080)
```

Any way to fix this?

**Comment by vabene1111** [comment]
can you test if you wrap the call to gunicorn in this script to check if ipv6 is enabled and put a second command without the ipv6 bind in the else block if that works for you?

```
#!/bin/bash

ipv6_disable=$(cat /sys/module/ipv6/parameters/disable)

# Check if IPv6 is enabled
if [ "$ipv6_disable" -eq 0 ]; then
    exec gunicorn -b "[::]:$TANDOOR_PORT" --workers $GUNICORN_WORKERS --threads $GUNICORN_THREADS --access-logfile - --error-logfile - --log-level $GUNICORN_LOG_LEVEL recipes.wsgi
else
    exec gunicorn -b ":$TANDOOR_PORT" --workers $GUNICORN_WORKERS --threads $GUNICORN_THREADS --access-logfile - --error-logfile - --log-level $GUNICORN_LOG_LEVEL recipes.wsgi
fi
```

funkybunch:
@vabene1111 I was experiencing this issue as well - I moved Tandoor from a box that had IPv6 to a box that didn't and ran into this same problem.

I applied your suggested fix to `./boot.sh` inside the Docker container and it fixed the problem for me.  I can submit a PR if you don't already have one ready.

vabene1111:
aebc4a45ffc22d26192b0e7109d95a4f5de4a676

PR #2728, opened 2023-11-10
Improve import error messages

### LINKED ISSUE #2596: "Bad Request" and "Error Fetching Resource"
### Tandoor Version

1.5.5

### Import URL

https://foodskitchen.com/optavia-lean-and-green-cheeseburger-pie-recipe/

### When did the issue happen?

I used the upper right "+" icon, hit import recipe. Then pasted the URL in. Before hitting anything else, I see the error message.

### Response / message shown

```shell
Failure: There was an error fetching a resource!

In the logs: tandoor-web_recipes-1    | Bad Request: /api/recipe-from-source/
tandoor-web_recipes-1    | 172.29.0.4 - - [28/Aug/2023:20:28:12 -0400] "POST /api/recipe-from-source/ HTTP/1.0" 400 53 "http://<removed>:xx/data/import/url" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/116.0.0.0 Safari/537.36"
```

jrester:
This improves the error reporting when importing recipes via a URL. Currently, it will just throw an error without any further information. To fix this, I decided to just propagate the error returned by the backend via the standard toast. The limitation of this is, that there is no localization for the error message from the backend. If you see the need for this and an easy way to add it, please let me know!

Fixes #2596

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -29,4 +29,3 @@ vue/babel.config*
 vue/package.json
 vue/tsconfig.json
 vue/src/utils/openapi
-venv/
--- a/.dockerignore
+++ b/.dockerignore
@@ -29,3 +29,4 @@ vue/babel.config*
 vue/package.json
 vue/tsconfig.json
 vue/src/utils/openapi
+venv/
--- a/cookbook/views/api.py
+++ b/cookbook/views/api.py
@@ -1454,7 +1454,7 @@ def import_files(request):
     """
     limit, msg = above_space_limit(request.space)
     if limit:
-        return Response({'error': msg}, status=status.HTTP_400_BAD_REQUEST)
+        return Response({'error': True, 'msg': _('File is above space limit')}, status=status.HTTP_400_BAD_REQUEST)
 
     form = ImportForm(request.POST, request.FILES)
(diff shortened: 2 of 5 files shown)


.dockerignore
```diff
@@ -29,3 +29,4 @@ vue/babel.config*
 vue/package.json
 vue/tsconfig.json
 vue/src/utils/openapi
+venv/
```
jrester, line 32: Just a small fix for the docker build. In the `.gitignore` the venv is already excluded, but it also has to be done for docker, otherwise the build will fail, when a venv is used.

vabene1111: not sure, it has not been ignored for a long time and everything worked, I need to look into that but we should seperate the change from changing the importer error message as these two things have nothing to do with each other.

vabene1111:
thanks for the PR, will look into it.

.dockerignore
jrester: Yes, you are right. I removed this and will create a separate issue to track this.

vabene1111:
thanks, I like what you did there.

PR #2758, opened 2023-11-28
Updated documentation

smilerz:


.env.template
```diff
@@ -28,7 +28,7 @@ SECRET_KEY_FILE=
 # ---------------------------------------------------------------
 
 # your default timezone See https://timezonedb.com/time-zones for a list of timezones
-TIMEZONE=Europe/Berlin
+TZ=Europe/Berlin
```
vabene1111, line 31: is there a reason for changing this? thousands of people will have to change their environment ...

vabene1111:
one comment, other than that this will help a lot I think for all the people struggling with the database

.env.template
vabene1111: ok so it can be taken from the default environment, that makes sense. should we in this case also comment out the TZ line in the default .env as the system should then load the hosts computer default time zone or do you think that many people might have their data center in a different time zone?

also added a comment below

recipes/settings.py
```diff
@@ -450,7 +450,11 @@
 
 LANGUAGE_CODE = 'en'
 
-TIME_ZONE = os.getenv('TIMEZONE') if os.getenv('TIMEZONE') else 'Europe/Berlin'
+if os.getenv('TIMEZONE') is not None:
+    print('DEPRECATION WARNING: Environment var "TIMEZONE" is deprecated. Please use "TZ" instead.')
+    TIME_ZONE = bool(int(os.getenv('REVERSE_PROXY_AUTH', False)))
```
vabene1111, line 455: shouldn't this be `TIME_ZONE = bool(int(os.getenv('REVERSE_PROXY_AUTH', False)))` ?

I think `REVERSE_PROXY_AUTH` might have slipped in here

smilerz: good catch - ouch!

.env.template
smilerz: the alpine container still needs the TZ env variable passed.

PR #2771, opened 2023-12-02
Fix parsing DATABASE_URL with port number

tourn:
A recent change in the DATABASE_URL regex ignores the PORT and DB part when a port number is provided:
image

With this change it parses properly again:
image

### DIFF
--- a/recipes/settings.py
+++ b/recipes/settings.py
@@ -350,7 +350,7 @@ WSGI_APPLICATION = 'recipes.wsgi.application'
 # Load settings from env files
 if os.getenv('DATABASE_URL'):
     match = re.match(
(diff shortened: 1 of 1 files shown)


vabene1111:
Thanks, no e should probably move that to a dedicated function and have some unit tests for the most common combinations

@smilerz can you post the strings you wanted to support with the change?

smilerz:
> Thanks, no e should probably move that to a dedicated function and have some unit tests for the most common combinations
> 
> @smilerz can you post the strings you wanted to support with the change?

Making username/password optional to support settings sqlite via URL

vabene1111:
sorry, I use connection string extremely rarely, how would the string look then with the file path and without username/password ?

smilerz:
> sorry, I use connection string extremely rarely, how would the string look then with the file path and without username/password ?

sqlite://localhost/data/db.sqlite3

PR #2785, opened 2023-12-06
Add instructions to add the PWA on Firefox for Android

axeleroy:
Hello,

I saw that instructions to install the PWA on Firefox for Android were missing, so I added them to the doc. I also fixed Chrome's heading level  (it was `h3` instead of `h4` like Safari)


vabene1111:
thank you :)

PR #2801, opened 2023-12-12
Reducing fonts.googleapis.com dependency

### LINKED ISSUE #2784: Remove dependency on fonts.googleapis.com

The app is currently depending on fonts.googleapis.com. This is caused by some of the stylesheets of the themes:
```
cookbook/static/themes/darkly.min.css
cookbook/static/themes/flatly.min.css
cookbook/static/themes/superhero.min.css
```

I think removing this dependency and providing a vendored version of the font makes more sense. The two reasons for this are:
- increased privacy
- still being able to access the page on your local network without any issues in case of an internet outage

I was affected by the latter one and had to block fonts.googleapis.com in my browser so that the site would continue to load.

An alternative would be to just refrain from using themes that rely on fonts.googleapis.com. But I think that's rather a workaround than an actual solution.

### Additional context

Related: #75

connormaas-cmu:
fixes #2784

additional note: Jest configuration added to top level


vabene1111:
thanks for the PR and potentially fixing the issue but you checked in over 4.000 files including all the node modules. I can't review and definitly can't merge this. 

If you want to help fix this please find a way to do so without checking in all node modules and changing the least amount of files possible.

PR #2808, opened 2023-12-14
add ability to filter meal plans based on type

smilerz:
I thought there was a feature request for this but I can't find it.


vabene1111:
i kinda wanted to do that when i changed the frontend but I lacked UI skills to make it look nice so I never did the endpoint for it 😂 

do you think we should make this a list parameter to allow filtering for multiple types?

smilerz:
Yes. Probably.  I'll take a look at it this weekend.

vabene1111:
i think there are other endpoints that take lists with get parameters, feel free to copy logic from there

smilerz:
lol you think I can do anything other than copy code :-)

smilerz:
This adds multiple meal_types as part of the API filter.  Should be good to merge now.

Issue #2338, opened 2023-02-27
Reorder books


ptrotary:
### Is your feature request related to a problem? Please describe.

Currently the books section lists the books in the order I create them.

### Describe the solution you'd like

I would like to be able to reorder them, manually/alphabetically/etc.

m7modSy:
## Response to Feature Implementation

Hello @ptrotary ,

 I'm excited to let you know that I've implemented the ability to reorder books in the books section. Users can now manually reorder books according to their preference.

### Changes Made:
- Now you can sort the books in 4 different ways.
1. oldest to newest.
2. newest to oldest.
3. alphabetical order.
4. sort manually:
Here you can re-sort the books as you wish.
The changes are saved even after you leave the website.
you can swap 2 directly adjacent books or you can specify the location where the book should be saved. drag-and-drop functionality was also an idea. but this is inefficient for those who use a phone or table.

You can view the changes in the https://github.com/TandoorRecipes/recipes/pull/2810. I hope this addresses your request. Please feel free to test it out, and let me know if you have any feedback or if there's anything else you'd like to see improved.


Best regards,
@m7modSy

m7modSy:
Little Change to the Solution:
The manual sorting mechanism has been updated to a drag-and-drop functionality.

PR #2810, opened 2023-12-15
Reorder books #2338

m7modSy:
Implemented the ability to sort books in four different ways:

1. **Oldest to Newest:** Sorts books based on their creation date.
2. **Newest to Oldest:** Sorts books in reverse order of creation date.
3. **Alphabetical Order:** Sorts books alphabetically.
4. **Manual Sorting:** Users can now manually reorder books. Changes are saved persistently, even after leaving the website.

Manual Sorting Features:

- Swap: Users can swap the positions of two directly adjacent books.
- Specify Location: Users can specify the desired location for a book.
- Drag-and-Drop: Initially considered, but not implemented to ensure efficiency, especially for users on mobile devices.

**Note:** This pull request addresses [issue #2338](https://github.com/TandoorRecipes/recipes/issues/2338).


smilerz:
- ordering should be handled within the API instead of on the front end
- manually ordering should be added to the Book model and be persistent

vabene1111:
Hi, thanks for your PR. As smilerz said there is some room for improvement, espcecially since this adds a lot of code to the frontend. 

We usually do ordering in tandoor by adding an `order` field to the model (see for example MealType). 

For automatic ordering this is not needed as you could just add a query param to the API, for manual ordering that field could be edited by the user (not very user friendly but thats also the case for properties) but automatic ordering and manual ordering could conflict so that needs to be taken into account.

vabene1111 (changes requested):
see comments above, please also make sur to share a draft of what you want to do before writing code so that we can save you time by giving feedback directly (see https://docs.tandoor.dev/contribute/)

m7modSy:
Hallo, 
Thank you guys @smilerz @vabene1111  for your Feedback on my PR :)

For automatic ordering i think its pretty straight forward.  i added 2 options for ordering through the api client
1. order_field: Specify the field to order by
2. Specify 'asc' for ascending order or 'desc' for descending order.

manual ordering:
I am very confused here:
I have seen how you handle the field order. My thoughts are that I make an update request with every swap, but is that really efficient or even possible?

smilerz:
> manual ordering: I am very confused here: I have seen how you handle the field order. My thoughts are that I make an update request with every swap, but is that really efficient or even possible?

ascending and descending should be part of the API/Serializer, not the model.
The behavior on the API is handled through a combination of Model, API and Serializer.  (model.py, api.py and serializer.py)

Take a look at SupermarketCategory for an example on how this has been handled.

m7modSy:
Hi @smilerz and @vabene1111,
you can take a look now at the code :)
you can now change the manual order by dragging and dropping, exactly like the SupermarketCategory

cookbook/views/api.py
```diff
@@ -651,15 +651,23 @@ def destroy(self, *args, **kwargs):
             content = {'error': True, 'msg': e.args[0]}
             return Response(content, status=status.HTTP_403_FORBIDDEN)
 
-
+import json
```
vabene1111, line 654: please have imports on the top of the file

cookbook/migrations/0206_recipebook_order.py
vabene1111, line 1: can you delete the migration from the PR please, I will create it after merging so that the line is correct.

vabene1111 (commented):
ok finally got around to reviewing this, just 2 minor changes and we are good to go

PR #2821, opened 2023-12-20
Fix various plural inconsistencies

smilerz:
added use plural setting to Manage Space page
updated FoodEditor to respect setting
deprecated Manage Space form


smilerz:
It took me hours to figure out why I could edit plural on Food but not on Unit.

vabene1111:
i am confused, I thought we removed the space plural setting 🤔  need to look into this

smilerz:
The setting was still
In the DB though the default is false. 

The model list page also still
Had the logic to remove plural from the unit modal (but not the food modal)

If you'd like to remove the setting I can update the PR to completely remove it.

vabene1111:
i will take a look when merging this, removing is easy enough if I decide its right, thanks anyway for the fix!

vabene1111:
thanks for the PR, I decided to go a different route and just eliminate that setting completely as it was not implemented in most places anyway and plurals are now kind of a core functionality so the fields to set them will always show in the editor and if people dont want to use them they can just ignore/not fill those fields.

Thanks for bringing up the inconsistencies though, I thought they were all resolved back when we first added this.

PR #2823, opened 2023-12-20
Admin updates

smilerz:
updates to various admin pages

### DIFF
--- a/cookbook/admin.py
+++ b/cookbook/admin.py
@@ -198,6 +198,7 @@ def rebuild_index(modeladmin, request, queryset):
 class RecipeAdmin(admin.ModelAdmin):
     list_display = ('name', 'internal', 'created_by', 'storage', 'space')
     search_fields = ('name', 'created_by__username')
+    ordering = ('name', 'created_by__username', )
     list_filter = ('internal',)
     date_hierarchy = 'created_at'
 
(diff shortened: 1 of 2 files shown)


cookbook/admin.py
```diff
@@ -163,10 +168,18 @@ def delete_unattached_steps(modeladmin, request, queryset):
 
 
 class StepAdmin(admin.ModelAdmin):
-    list_display = ('name', 'order',)
-    search_fields = ('name',)
+    list_display = ('recipe_and_name', 'order', 'space')
+    ordering = ('recipe__name', 'name', 'space', )
+    search_fields = ('name', 'recipe__name')
     actions = [delete_unattached_steps]
```
vabene1111, line 174: usually admin actions affect only selected entries, with this we will delete everything from every space, do you think that is sufficently clear to the user?

I am leaning towards yes as playing in admin is always only for advanced users but wanted to check

vabene1111 (commented):
just one question/comment

smilerz:
That was my thought process as well.  I can investigate to see if there is a way to throw a confirmation.

vabene1111:
I think its fine, just wanted to double check, might change the text to "Delete ALL .." or something but other than that its fine, will get this merged (if I dont forget) for the next update.

vabene1111 (approved):

PR #2842, opened 2023-12-29
Coverage Reports

smilerz:
Added coverage to pytest
Generated reports and pushed to gh-pages
generated badges for coverage and tests
updated README with badges and links to reports.

Note:  [Changes may be necessary on production](https://github.com/stefanzweifel/git-auto-commit-action/tree/v5/?tab=readme-ov-file#push-to-protected-branches), I don't have a way to test against protected branches.


.github/workflows/ci.yml
```diff
@@ -38,3 +38,16 @@ jobs:
...
+    - name: Generate badges
+      if: github.ref == 'refs/heads/master'
+      run : |
+        genbadge coverage -i docs/reports/coverage/coverage.xml -o docs/reports/coverage/coverage-badge.svg
+        genbadge tests -i  docs/reports/tests/pytest.xml -o docs/reports/tests/pytest-badge.svg
+    - name: Commit reports and badge
+      if: github.ref == 'refs/heads/master'
+      uses: stefanzweifel/git-auto-commit-action@v5
+      with:
+        commit_message: Pytest and Coverage Reports
```
vabene1111, line 50: this creates a master commit when it runs ?! wouldn't that mean that there is lots of commit spam on master?

vabene1111 (commented):
added two comments, is there no way to just have the coverage report in the logs of github actions ?

for me this seems like an awful lot of things that can break for a percentage number in the readme ... but if there is no other way its fine

.github/workflows/ci.yml
smilerz: The coverage reports would only run during CI on master, so would limit commits to just releases.

.github/workflows/docs.yml
```diff
@@ -1,17 +1,20 @@
...
+  workflow_run:
+    workflows: ["Continuous Integration"]
+    branches: [master]
+    types:
+      - completed
 
 jobs:
   deploy:
-    if: github.repository_owner == 'TandoorRecipes'
+    if: github.repository_owner == 'TandoorRecipes' && ${{ github.event.workflow_run.conclusion == 'success' }}
```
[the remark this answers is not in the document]
smilerz: Yes, but this should be updated to any status, not just success.  It just needs to wait for the action to complete.

.github/workflows/ci.yml
vabene1111: thanks for the clarification .. to be completely honest I am really not sure what to think about this, I don't super like it this way ... 

How useful are the coverage results, is it actually possible to find spots where test coverage is bad?
also who will write the tests we know are missing, if I look at the github issues and also the things I just want to do I dont really think I will ever have the time to actually make anything out of these results while I will likely have to fix something every 1-2 years when it breaks for some unknown reason.

smilerz: The coverage reports are fully annotated so are helpful.  There is an example [here](https://smilerz.github.io/recipes/reports/coverage/)

That being said, there are other options. Setup coverage so it's only ever produced locally.  There is an online service that will host coverage reports and provide a badge (shields.io).

vabene1111: thank you for the input, I will have to think about this but this code is pretty detatched so should not break for a while when working on other stuff.

PR #2846, opened 2023-12-30
Bug 2838 - addes support for Hebrew RTL and support for RTL/LRT recipes

arnonm:


vabene1111:
thanks for creating this but I cannot accept this PR:

- please read https://docs.tandoor.dev/contribute/#contributing-code 
- I rejected having a language field on the recipe in several comments 
- support for proper LTR/RTL should be considered in a general matter regarding all modules, themes, etc. 
- all language codes that use RTL should be configured and overridable

arnonm:
Thanks,

I personally think there is no way around support LTR and RTL without some
indication of either the direction or the language of the actual recipe.
As for the general matter, I agree - it is going to be a number of PR. The
frontend will need adjusting/tweaking, as will the importers, etc.
Also, the direction criteria will need to be increased to include all the
languages that support it, he/ar as a minimum.
I can handle it on a side branch until it is complete... How do you prefer?

vabene1111:
So what I usually prefer is a technical draft first so that we can discuss it and gather feedback as often either I or some community member has input on a topic. Then everything can be implemented based on what's agreed upon.

That saves people from doing work that I reject for one reason or the other in the end and also saves countless feedback loops.

So if you want to help here, which I would appreciate, please start with an issue, probably since I don't know to much about RTL besides what the name says (maybe that's all already), start with a short explanation and what is currently missing and then what exactly needs to be changed and how in short bullet points.

arnonm:
Will do

PR #2859, opened 2024-01-03
Touchable fix

harry48225:
Increases the touch/hover target size for the ingredients comments button. (I think there was already a mitigation for this but it didn't quite work)

The new target is about the width of the comment icon either side, before it was just the comment icon itself that was the touch/hover target

Also tweaks the vertical alignment of the icons in the row.

Before:
image

After:
image


vabene1111:
thank you, looks good!

PR #2872, opened 2024-01-11
Improved the understandability of the postgres upgrade steps

FaySmash:
I improved the understandability some parts in the psql examples for someone not familiar with the psql syntax.


vabene1111:
thank you for the improvement

PR #2874, opened 2024-01-11
Home assistant connector

Mikhail5555:
Feature showcase video (high quality)

ListView:
image

Edit Screen:
image

If you want me to remove the ci optimizations from this PR. Please let me know

--- Ignore this configuration for connectors.md
image
image
image


vabene1111:
Thank you, as discuessed on discord I will review as soon as possible but it might take some time since this is a rather large change. 

Just from looking at the images: You included homassistant in the UI, do you think that makes sense if the connectors should be usable for all kinds of integrations? Is it maybe preferable to name everything independently of the service.

cookbook/signals.py
```diff
@@ -161,3 +163,9 @@ def clear_unit_cache(sender, instance=None, created=False, **kwargs):
 def clear_property_type_cache(sender, instance=None, created=False, **kwargs):
     if instance:
         caches['default'].delete(CacheHelper(instance.space).PROPERTY_TYPE_CACHE_KEY)
+
+
+if ENABLE_EXTERNAL_CONNECTORS:
+    handler = ConnectorManager()
+    post_save.connect(handler, dispatch_uid="connector_manager")
```
vabene1111, line 170: does this add the post_save to all models? should we limit this to only signal on save events of ShoppingListEntry objects?

cookbook/views/api.py
```diff
@@ -864,12 +884,14 @@ def paginate_queryset(self, queryset, request, view=None):
...
-        return Response(OrderedDict([
-            ('count', self.page.paginator.count),
-            ('next', self.get_next_link()),
-            ('previous', self.get_previous_link()),
-            ('results', data),
-        ]))
+        return Response(
+            OrderedDict(
+                [
+                    ('count', self.page.paginator.count),
```
vabene1111, line 890: I would have preffered if the file wasn't reformatted everywhere, maybe you can discard a few of these changes so its a bit easier to see whats actually changed.

cookbook/views/new.py
```diff
@@ -71,6 +70,31 @@ def get_context_data(self, **kwargs):
...
 
+class ConnectorConfigCreate(GroupRequiredMixin, CreateView):
+    groups_required = ['admin']
+    template_name = "generic/new_template.html"
+    model = ConnectorConfig
+    form_class = ConnectorConfigForm
+    success_url = reverse_lazy('list_connector_config')
+
+    def form_valid(self, form):
+        if self.request.space.demo or settings.HOSTED:
```
vabene1111, line 81: i guess now that you added that extra flag to disable this feature, lets tie it to that and say "not available in demo" for demo and "not enabled on this instance" so that I can enable it eventually

recipes/settings.py
```diff
@@ -555,4 +555,6 @@
 ACCOUNT_EMAIL_SUBJECT_PREFIX = os.getenv(
     'ACCOUNT_EMAIL_SUBJECT_PREFIX', '[Tandoor Recipes] ')  # allauth sender prefix
 
+ENABLE_EXTERNAL_CONNECTORS = bool(int(os.getenv('ENABLE_EXTERNAL_CONNECTORS', False)))
```
vabene1111, line 558: we have new docs for the configuration parameters, can you add this setting to the list there. Also I think we can enable it by default, it looks good so far.

vabene1111 (changes requested):
Ok so finally had the time to review this. Overall I think everything looks good, I just added a few minor comments here and there. 

One big thing for me though is that I do not understand the connector_manager.py (or generally the whole stack of how the signal is processed) ways of working very well or at least that whole async queue thing is something I have not used before. Do you think you could add, maybe directly in the file, a few lines of comments of how this thing works and how the queue is supposed to work (signal will trigger items to be added to the queue, the queue will run them in parallel or after another?!).
Maybe also a few links to the django docs if applicable or whatever you think might make it easier for me in 3 years to understand whats wrong when something breaks 😂

.github/workflows/ci.yml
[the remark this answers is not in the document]
vabene1111: ok I see, yes please put that in a seperate PR because if something break we wont know were it originated from otherwise.

.gitignore
```diff
@@ -54,7 +54,7 @@ docs/_build/
 target/
 
 \.idea/dataSources/
-
+.idea
```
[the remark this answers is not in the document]
vabene1111: interesting, there are a few idea files that should be checked in (just makes it easier for me across devices) so please remove it. The question is if those files should be ignored, but I think that's for another day.

cookbook/connectors/homeassistant.py
```diff
@@ -0,0 +1,85 @@
...
+                todo_domain = await self._client.async_get_domain('todo')
+                self._domains_cache['todo'] = todo_domain
+
+            logging.debug(f"pushing {item} to {self._config.name}")
+            await todo_domain.add_item(entity_id=self._config.todo_entity, item=item)
+        except HomeassistantAPIError as err:
+            self._logger.warning(f"[HomeAssistant {self._config.name}] Received an exception from the api: {err=}, {type(err)=}")
+
+    async def on_shopping_list_entry_updated(self, space: Space, shopping_list_entry: ShoppingListEntry) -> None:
+        if not self._config.on_shopping_list_entry_updated_enabled:
```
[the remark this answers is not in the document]
vabene1111: ok, I understand. The shopping rework will support changing the quantities of manually added entries afterwards but we dont need to support everything in the connector.

cookbook/signals.py
vabene1111: I understand that it is more error prone but I am unsure about performance. Tandoor useses a ton of models so many many many changes are triggered on a lot of operations. 

Signals are far from free in regards of performance so their use should be limited as good as possible so I suggest looking into if a more limited signal could be added.

vabene1111:
ok thanks for the changes. I added some more comments, after those have been resolved I would indeed suggest we quickly hop on discord and talk about the queue system so I am sure that I understand it. I will message you on discord once I have time.

vabene1111:
ok so final review is trough, I am happy with the state. Two final things

1. please rebase the PR so I can merge it
2. add the new environment variables to the environment docs

after that I will get it merged for the next release

vabene1111 (approved):

pippo73:
Hello @vabene1111 , I've discovered this great feature, but it seems not working with my systems. :-(
Where should I have to open the issue? here or on HA. How can I test where is the problem? Could you help me ... anyone else is welcome too :-)

vabene1111:
haha I had some of those as well in the past, one pretty recently but I do not remember exactly what caused it. There are indeed a few functions in django that behave differently when running DEBUG=1 and also when testing with the dev server. I see you already got it sorted out, very nice!

Issue #1257, opened 2022-01-08
Run processes inside container as non-root user


graduatio:
**Is your feature request related to a problem? Please describe.**
The processes inside the recipes container currently run as user root. This is potentially problematic for security reasons.

**Describe the solution you'd like**
Whenever possible the processes should run as non-root user.

graduatio:
I tried to run all containers (nginx, postgres and recipes) as a non-root user. The nginx and postgres docker documentations describe how the containers can run as an arbitrary non-privileged user. I followed the steps below:

For the nginx container:

- added `user: "${UID}:${GID}"` in `docker-compose.yaml` to the `nginx_recipes` section
- added to the `volumes` section in `docker-compose.yaml`:
	`- ./nginx.conf:/etc/nginx/nginx.conf`
- created a `nginx.conf` file in the current directory with following redefined directives:
```
pid        /tmp/nginx.pid;
http {
    client_body_temp_path /tmp/client_temp;
    proxy_temp_path       /tmp/proxy_temp_path;
    fastcgi_temp_path     /tmp/fastcgi_temp;
    uwsgi_temp_path       /tmp/uwsgi_temp;
    scgi_temp_path        /tmp/scgi_temp;
}
```
For the postgres container:
- changed to a Debian Variant image (e.g. postgres:latest)
- added `user: "${UID}:${GID}"` in `docker-compose.yaml` to the `db_recipes` section

For the recipe container:
- added `user: "${UID}:${GID}"` in `docker-compose.yaml` to the `web_recipes` section

Before I started the containers:
- added `UID=` and `GID=` with the desired values to the `.env` file
- created a `mediafiles` and a `postgresql` directory and used `chown` to change ownership to the user which should run the containers

With this setup the database and the nginx server seem to work properly. Tandoor can be accessed by opening the ip-address. But the docker-compose logs show permission errors of the recipes container which cannot change files in `/opt/recipes` when executing the commands in `boot.sh`.
This error occures because `opt/recipes` was created by user root. Changing the owner of the files inside the container manually resolves the problem temporarily.
Maybe this issue can be resolved by setting a default user and changing the owner of `opt/recipes` in the Dockerfile.

vabene1111:
thanks, i will take a look at this at some point when the more pressing features are taken care of

Akruidenberg:
Any progress so far. I would like to run the recipe container rootless. The trick of @graduatio is working, only manually .

graduatio:
[...]

PR #2878, opened 2024-01-13
make images use non-root users

DanielHabenicht:
fixes https://github.com/TandoorRecipes/recipes/issues/1257

also optimized image size of the "recipes" image from 365 MB to 202 MB

Open Questions: 
- [ ] Where do you want the image published to? So I can modify the workflow file. (In my mind maybe only to ghcr.io? This way any fork can also test the workflows) 
- [ ] How should the static files be handled? Now that there is an extra image we could also include them directly in the NGINX image. 

TODOs:
- [ ] Delete root docker-compose.yml

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -29,3 +29,5 @@ vue/babel.config*
 vue/package.json
 vue/tsconfig.json
 vue/src/utils/openapi
+postgresql
+mediafiles
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,15 @@
-FROM python:3.10-alpine3.18
+FROM python:3.10-alpine3.18 as base
 
-#Install all dependencies.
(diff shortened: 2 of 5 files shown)


vabene1111:
thanks for the PR. To be honest I am very hesitant to merge this is this is a major change to the build/release process and from experience something will break, so we have to do some VERY VERY VERY thorough testing. Regarding your questions:

images should be published to docker hub and github, as they are currently being published to these platforms.

staticfiles: Why do we need a non standard nginx image, I always tried to avoid that? I am not yet sure if I prefer staticfiles being inside the container, will need to think about that. 

Have you tested that upgrading an installation from the old system to this version works without anything breaking? if yes please describe exactly how you tested (versions, OS, procedure, ...)

Are you 100% sure that chaning existing data to a new user will not break existing setups/cause permission issues?

Do existing users have to modify their compose when they get this update?

Issue #2576, opened 2023-08-11
Hide iOS Navbar when the site is run as an App in the homescreen


husseinfahmy:
### Is your feature request related to a problem? Please describe.

Love the project, and I have my instance saved as an app on my iPhone homescreen.

An inconvenience I've noticed is that the navigation bar at the bottom of the screen gets in the way of the app and is intrusive to the experience.

### Describe the solution you'd like

There are 2 quick solutions to this that come to mind

1) pad the bottom of the screen so that the menu items aren't obscured (recipes, meal plan, +, shopping list, books) - This is probably simple(?)
2) Hide the nav bar altogether - this seems to be a [1 line solution](https://stackoverflow.com/questions/6011223/how-to-completely-hide-the-navigation-bar-in-iphone-html5), plus potentially another couple of lines to detect whether the client device is an iOS device if this line doesn't play well with other platforms (web, android, etc)

I don't have the dev env set up on my computer, but if you don't have time for this, I can probably just pull a docker image with the dependencies i need and work on it there.

vabene1111:
thanks, this needs to be improved/changed. will look at it at some point and if anyone is interested feel free to open a PR, this should be a rather easy issue.

vabene1111:
ok so I just added the header to the base file proposed on the stackoverflow post. I dont have any apple devices so maybe some apple user could give feedback if the fix was successfull. It can be tested on develop but you cant downgrade since there are migrations involved.

sthuber90:
I've tried testing the fix on the _develop_ branch with docker compose with a fresh install, meaning no local files and no docker volumes. I was able to reach the setup page but after setting up the superuser I only get "Server Error (500)".

<details><summary>docker-compose.yml</summary>
<p>

  ``` yaml
  version: "3"
  services:
    db_recipes:
      restart: always
      image: postgres:15-alpine
      volumes:
        - ./postgresql:/var/lib/postgresql/data
      env_file:
        - ./.env
  
    web_recipes:
      restart: always
      image: vabene1111/recipes:develop
      env_file:
        - ./.env
      volumes:
        - staticfiles:/opt/recipes/staticfiles
        # Do not make this a bind mount, see https://docs.tandoor.dev/install/docker/#volumes-vs-bind-mounts
        - nginx_config:/opt/recipes/nginx/conf.d
        - ./mediafiles:/opt/recipes/mediafiles
      depends_on:
        - db_recipes
  
    nginx_recipes:
[...]

PR #2895, opened 2024-01-20
Fix iOS view

tomtjes:
This should solve #2576.

Here are a couple of before ([1.5.10](https://github.com/TandoorRecipes/recipes/releases/tag/1.5.10)) and after screenshots. The first set is from the PWA, the second set from a Safari browser tab.

Image 2

Image 3

Note: The recipe context menu (where you select "edit", among other things) disappeared on my dev instance. No idea why. But it must be unrelated to my changes, as I see the same behavior with the current develop branch.


vabene1111:
thanks for the fix, interesting that your context menu is gone, for me it is still there on my devices 🤔

vabene1111:
that error is related to the frontend named URL system, are you on a manual install?

this command might be what you need `bin/python3 manage.py collectstatic_js_reverse`

tomtjes:
That solved the missing context menu, thanks!

Issue #2520, opened 2023-06-26
UI glitches: search result hover, quick action button, iPhone home bar


tomtjes:
### Tandoor Version

1.5.2

### Setup

Docker / Docker-Compose

### Reverse Proxy

SWAG

### Bug description

I noticed some minor UI glitches and wanted to collect them here. All of this is seen on Safari on MacOS and iPhone.

## Recipe Description Hover 
![tandoor](https://github.com/TandoorRecipes/recipes/assets/7606307/a75e6275-2c37-4c5d-b7a0-2337385cd7de)
The description doesn't fit in the image space and spills into the space around it.

By the way: Since a few versions ago, I really miss seeing the recipe ratings on the search page. I think there's enough room in the image for another pill that displays the rating.
I would also appreciate an option to display all the keywords in the search results, since I'd like to see the recipe author, for instance.

## Quick Action Button 
https://github.com/TandoorRecipes/recipes/assets/7606307/2a1dc135-6974-47e6-9734-db7e800d9fcc
The quick action button doesn't move with the rest of the page when pulling down or scrolling to the top.

Also, as the above video shows, the home bar at the bottom of the iPhone interface interferes with the navigation icon texts.

vabene1111:
thanks for the report. 

regarding the recipe card: i always wanted to make it configurable, so at some point you might be able to choose what you like to see :)

vabene1111:
solved for now, will likely be further improved when migrating to the new frontend

PR #2898, opened 2024-01-21
Fix description overflow

tomtjes:
Partially solves #2520 by adding scroll to long descriptions:

scroll

Bonus: match the border color of the quick action button aka recipe switcher on mobile with the hamburger menu button's border.

IMG_9189


vabene1111:
thank you for the fixes and very well documented PR :)

PR #2899, opened 2024-01-21
fix tiny typo in README

jmbhughes:
I noticed a very minor typo ("then" instead of "than") in the README, so I fixed it.


vabene1111:
thank you :)

PR #2914, opened 2024-01-29
Add Caching and Test export to CI workflow

Mikhail5555:
With Caching:
image

Previous:
image

Added test results/summary:
image


vabene1111:
thank you, looks cool and good thing to be a bit faster. Since its only CI its also relatively safe to merge.

PR #2925, opened 2024-02-02
Improve menu: remove create & import recipe dropdown

ignas2526:
Remove the create & import recipe dropdown. Instead, place them directly into the menu. I think it makes for a better UX as you don't have to do an extra click. And there's enough space for both.

Screenshot 2024-02-02 092840

Screenshot 2024-02-02 092828


vabene1111:
thank you for the PR but I prefer the smaller variant, the nav is already pretty cluttered. It needs improvements but those will be a major rework.

Issue #2220, opened 2022-12-19
Recipe View Rework


vabene1111:
- [ ] some kind of compact mode #2175
- [ ] #1955

tomtjes:
Not sure if this project is still relevant with Tandoor 2.0, but I wanted to bump it for #1955. Until recently, it was possible to achieve it with custom CSS, hiding the bubbles and making the print view notes visible. That's no longer an option, so a toggle would be appreciated. 

I'd advocate for it being a space or user setting rather than a toggle in the recipe view, because I think it depends on how someone decides to use the notes fields on their instance, i.e. do they typically only contain _good to know_ information, or are they mostly prep instructions that you frequently rely on while cooking.

vabene1111:
makes sense. I tought about adding it to the steps overview when working on that or do you think it should be visible everywhere?

tomtjes:
Do you mean, the option to show the notes instead of the bubbles should only affect the steps overview? I don't think so. It would be nice to have it affect the steps too.
If you mean, should the toggle be in the steps overview, that would probably make sense, if at the same time, a default can be set per space.
(Sorry, it took me so long to get back to you. Also on other things I wanted to get my hands on. Life happened.)

vabene1111:
does for me all the time as well, no worries. I know this is a hot topic, will look into it again.

PR #2938, opened 2024-02-07
Add ability to open recipes with specified number of servings from meal plan

patmagauran:
When creating a meal plan, you can specify the number of servings to make which influences the amount of ingredients added to your shopping list. Then, when clicking the recipe from the plan it currently defaults to the recipe's default number of servings. Instead, this PR makes it so it defaults to the number in the meal plan. This only takes effect when navigating from the meal plan and should not interfere or change behavior elsewhere. 

I have also added a toggle in the meal plan settings to enable/disable this functionality. 

The loading of the correct number of servings is done by introducing an additional url format for recipes of /recipe/id-servings. If the -servings is omitted, it loads as normal. If it is included, the recipe will load with the specified number of servings. 

I am open to any changes in the handling of this as well. I have limited knowledge of django and vue so if there is a better way to propagate the number of servings, I will happily implement it as such.

sample URL: http://127.0.0.1:8000/view/recipe/1-3 -- Opens recipe 1 with 3 servings

The modified Settings page:
image

### DIFF
--- a/cookbook/templates/recipe_view.html
+++ b/cookbook/templates/recipe_view.html
@@ -79,6 +79,7 @@
 
         window.CUSTOM_LOCALE = '{{ request.LANGUAGE_CODE }}'
         window.RECIPE_ID = {{recipe.pk}};
+        window.RECIPE_SERVINGS = '{{ servings }}'
         window.SHARE_UID = '{{ share }}';
         window.USER_PREF = {
             'use_fractions': {% if request.user.userpreference.use_fractions %} true {% else %} false {% endif %},
--- a/cookbook/urls.py
+++ b/cookbook/urls.py
@@ -101,6 +101,8 @@ urlpatterns = [
     path('export-file/<int:pk>/', import_export.export_file, name='view_export_file'),
 
     path('view/recipe/<int:pk>', views.recipe_view, name='view_recipe'),
+    path('view/recipe/<int:pk>-<int:servings>', views.recipe_view, name='view_recipe'),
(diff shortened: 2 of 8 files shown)


vabene1111:
Hi, thanks for the PR. I like the idea but I think there is a much easier way to do this. Please feel free to next time share a concept before investing all the work so I can tell you beforehand. 

What I suggest is adding a normal named GET parameter to the recipe view /view/recipe/1/?servings=5 and then load this number in the mounted or after the API retrieve in the recipe view. I would also say that in this case you don't need a setting for the feature and the change is only something like 3 lines of code :)

patmagauran:
I went with the way I did it because I did not see any places you used named parameters and didn't want to break the pattern. It is the more standard way to handle it. The setting doesn't disable the url configuration either, it only disables the meal plan link change. Probably not necessary but I figured there may be some people who prefer that. Although a better way IMO would be to show plainly both the number of servings a recipe is currently being viewed with and the number that it defaults to with a quick option to reset. This may be important for baking recipes where different amounts mean different cooking times, etc.

I am open to changing it to the way you think best, no worries on the unnecessary work on my part

vabene1111:
perfect, I see your reasoning. 
I would say
- move to named parameter (I want to migrate to that pattern for those kind of things)
- remove the setting

I also agree to your point that the "original" servings should be shown in the recipe view, but thats a different matter for #2220

patmagauran:
I have updated it accordingly

vabene1111:
perfect, thanks for the changes, this looks good to me.

vabene1111 (approved):

PR #2948, opened 2024-02-13
Fixed wrong keycloak docs

maxemann96:
Keycloak was deprecated and removed in recent allauth version. This pr fixes the now outdated docs.


vabene1111 (approved):
thank you for fixing this

PR #2958, opened 2024-02-19
Add initial documentation on the use of properties.

wligtenberg:
As there was a specific issue regarding cost calculation. (And I was specifically looking for it)
I think it would be good to document the use of properties with an example, both using nutritional values and cost.
This should make the feature easier to discover.

### DIFF
--- a/docs/features/properties.md
+++ b/docs/features/properties.md
@@ -1,0 +1,16 @@
+!!! warning
+Properties are currently in a beta stage. They work pretty stable but if I encounter any
+issues while working on them, I might change how they work breaking existing properties.
+I will try to avoid this and am pretty confident it won't happen.
+
+`Foods` can have different `properties` associated with them. These `properties` can be defined by selecting the `Properties` option from the briefcase menu.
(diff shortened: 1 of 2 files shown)


vabene1111:
thanks for the PR! 
Its well meant but to be honest reviewing, writing this comment and merging this is more work for me than writing those few lines myself. Are you interested in also documenting things like recipe properties, the property editor, the meaning of the different fields and maybe also the relation to unit conversions?

If so I would leave this PR open for you to add more stuff and review it once you are finished to not have to read it a hundred times (because again than it would be faster for me to do it myself).

wligtenberg:
I wasn't planning on doing that. Since for me I wanted to ensure that people would find the feature, since I know it is a feature that people are looking for.
However, I do actually want to help out. So yes, I will expand the documentation to include those features as well. I was also not sure how to contribute screenshots. As I do think that adding some screenshots for that part of the documentation would help. I see that images are now linked to github user images, I can't seem to find them in the repo, but I might be missing something.

So please leave open, and I will add the requested bits of documentation.

vabene1111:
thank you very much and sorry for the inconvenience, I just need to somehow manage this all.

PR #2970, opened 2024-02-22
allow passing PK to nested serializer instead of object

smilerz:


cookbook/serializer.py
```diff
@@ -27,18 +28,43 @@
...
+                             )
 from cookbook.templatetags.custom_tags import markdown
 from recipes.settings import AWS_ENABLED, MEDIA_URL
 
 
+class WritableNestedModelSerializer(WNMS):
+
+    def to_internal_value(self, data):
+
+        def get_object(field, model, serializer, value):
```
vabene1111, line 44: is this just the code from wirtablenested adapted to support the new feature?

As this is a rather complex piece of code could you add some comments roughly what the different parts do ?

smilerz: I'll annotate the code.

vabene1111:
Ok so thanks for the comments, that makes it clearer. I really like the idea, I always wanted to be able to do this but I am a bit afraid about the performance implication if this feature is actually used (which it currently isn't but likely will be since thats why we do it). 

From what I see this will run a query for every PK passed to it, which in the case of lists can be a lot. I see several ways of proceeding in order of difficulty

1. leave it as is and just don't use the feature for very large sets. Will still have performance implications overall as it adds load to the server in many small requests but should be fine
2. change the method to first generate a list of PK's that need data, query it and then merge it onto the data object ( will still be one query extra but only one not potentially hundreds or thousands)
3. instead of changing the data change the validation/parsing logic. From what I see `to_internal_value` in serializer default only does validation and parses as a dict. Fields detected as PK (field class/int) could dynamically be changed on the serializer to use `fieldname_id=pk` instead of `fieldname=obj`. This would mean 0 queries but the return object would also only contain the ID (which I think is fine because if you used the ID in the frist place to create/update the object you don't expect an object back, it might even be easier this way). 

let me know what you think

smilerz:
> 3. instead of changing the data change the validation/parsing logic. From what I see `to_internal_value` in serializer default only does validation and parses as a dict. Fields detected as PK (field class/int) could dynamically be changed on the serializer to use `fieldname_id=pk` instead of `fieldname=obj`. This would mean 0 queries but the return object would also only contain the ID (which I think is fine because if you used the ID in the frist place to create/update the object you don't expect an object back, it might even be easier this way).

This option pushes the problem down to `validate_data` where validation will fail as required fields haven't been provided.  Can you think of another method around that?  Do you happen to know if the standard nested serializer allows for PK only?  Might be an option to intelligently switch between methods.

Option 2 should be doable without much trouble.

vabene1111:
I am not sure. I had hoped that we could modify validate data to not do any validation for the nested model (basically as if it were just an ForeignKeyField wihtout any nested serialization attatched). But I haven't looked into this much to be honest. Maybe we can change the field type in the to_internal_value method or override validate to change it there so it behaves like the nested serializer wasn't there?

Might also be something to discuss on the https://github.com/beda-software/drf-writable-nested repo directly as this would be a function likely relevant to many users but it seems pretty quite over there so probably something we do ourselves.

smilerz:
> I am not sure. I had hoped that we could modify validate data to not do any validation for the nested model (basically as if it were just an ForeignKeyField wihtout any nested serialization attatched). But I haven't looked into this much to be honest. Maybe we can change the field type in the to_internal_value method or override validate to change it there so it behaves like the nested serializer wasn't there?
> 
> Might also be something to discuss on the https://github.com/beda-software/drf-writable-nested repo directly as this would be a function likely relevant to many users but it seems pretty quite over there so probably something we do ourselves.

There is an open issue on the repo requesting this feature - I'm following it.  But it's been open for 2 years so I'm not hopeful for quick resolution :)

Let me dig into it a bit more and see what's possible.

vabene1111:
awesome, if you want to discuss it together let me know, writable nested is a very important part of this application, I am a bit unhappy seeing it not being actively maintained so every understanding about it is good I think.

smilerz:
> awesome, if you want to discuss it together let me know, writable nested is a very important part of this application, I am a bit unhappy seeing it not being actively maintained so every understanding about it is good I think.

Option 3 - overriding to_internal_value to dynamically set serializer - doesn't seem to be possible given the way that the nestedserializer is written - I've attempted about 10 different alternatives to work around it, but would required rewriting the whole mixin which seems like a bad idea.

I've resubmitted with basically Option 2 - getting the list of PKs as a single read and merging it into the initial data.

vabene1111:
ok thank you, this still means `n²` queries in the worst case szenario for something that "just" makes it a bit easier to write the frontend/api code. Still I would really like this feature as well so I guess I am going to merge it. 

I will probably manually copy it over because of all the reformatting I will not be able to merge it into vue3 afterwards with all the schema changes. Maybe I will also directly apply it to vue3 and not push it on stable so we have a bit more testing.

vabene1111 (approved):

smilerz:
Why is it n^2?  The list is retrieved with a single read?

vabene1111:
to_internal_value is called once per object of the list or at least once for the list (but i think once per object) and then there is a for loop over the fields so the query is executed once per object for each of the object fields that might be affected. Since writing is usually only done to detail views and since there are ususally not many FK fields that is likely not a problem.

vabene1111:
closing as I think this already got merged with the other PR

PR #2977, opened 2024-02-25
Allow tokens for getting ical

c0mputerguru:
I host an instance of tandoor on my local server, and I'm looking to expose the ical to google calendar via an nginx rewrite rule.  I was planning on having the bearer token as a part of the rewrite rule and tried it, but seems that the specific API endpoint requires a user.  I've made changes that allow it to be accessed with a token as well.


vabene1111:
thank you, that makes sense. Do you feel comfortable adding some tests in the same manner as the other tests making sure this permission rule works?

they are all in the tests_api folder if you want some references. 

I also like your other PR #2978, if you want you can put them together and add tests for both of them at the same time.

c0mputerguru:
I'm not too familiar with how these tests work, but I'll happily give it a
try.  I'll see about combining the PRs and writing up tests, but may need
to bug you for help if I can't easily get it working.

vabene1111:
no worries, feel free to reach out if you have any questions, if you can't get it to work or don't have the time just tell me and I will look into it after merging.

PR #2987, opened 2024-02-29
Add devcontainer support

c0mputerguru:
So I started down the path of wanting to write test for #2978, but didn't want to deal with venvs and messing around with installing random dependencies on my machine.  I've used devcontainers in the past to simplify creation of a development environment, and figured this would be great to simplify the development environment for tandoor.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,26 @@
+FROM python:3.10-alpine3.18
+
+#Install all dependencies.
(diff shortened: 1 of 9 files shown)


vabene1111:
very interesting, I did not know this existed. So from what I understand this is a setup once and forget thing because if I add new dependencies in package.json or requirements it will always automatically pull them trough the install process right ?

Also just out of curiosity, is this working with GitHub Codespaces? If so we could add a small hint that this works too, what do you think?

vabene1111:
thanks for the feedback and information, going to merge this then.

PR #3016, opened 2024-03-01
Coverage

smilerz:
Add coverage reports to pytest
Remove development libraries from docker build


vabene1111:
ok looks good to me, I guess it makes it so the docs don't build when CI fails but that is fine.

Issue #3036, opened 2024-03-14
Auto onhand not working


smilerz:
### Tandoor Version

1.5.14

### Setup

Docker / Docker-Compose

### Reverse Proxy

No reverse proxy

### Bug description

When enabled, auto onhand should mark a food as 'onhand' when it is completed in the shopping list.

PR #3048, opened 2024-03-19
Shoppinglist fix

### LINKED ISSUE #3044: Shopping List Incomplete
### Tandoor Version

1.5.14

### Setup

Docker / Docker-Compose

### Reverse Proxy

No reverse proxy

### Other

_No response_

### Bug description

When grouped by category not all items are displayed.  Group by recipe and group by created_by show all items

### Relevant logs

_No response_

smilerz:
fixes #3044 and #3036

### DIFF
--- a/cookbook/views/api.py
+++ b/cookbook/views/api.py
@@ -1168,12 +1168,24 @@ class ShoppingListEntryViewSet(viewsets.ModelViewSet):
     @decorators.action(detail=False, methods=['POST'], serializer_class=ShoppingListEntryBulkSerializer, permission_classes=[CustomIsUser])
     def bulk(self, request):
         serializer = self.serializer_class(data=request.data)
+
         if serializer.is_valid():
-            ShoppingListEntry.objects.filter(Q(created_by=self.request.user)
-                                             | Q(created_by__in=list(self.request.user.get_shopping_share()))).filter(space=request.space, id__in=serializer.validated_data['ids']
-                                                                                                                      ).update(checked=serializer.validated_data['checked'],
-                                                                                                                               updated_at=timezone.now(),
-                                                                                                                               )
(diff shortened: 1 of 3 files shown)


cookbook/views/api.py
```diff
@@ -1168,12 +1168,24 @@ def get_queryset(self):
...
-                                                                                                                               updated_at=timezone.now(),
-                                                                                                                               )
+            print(serializer.validated_data)
+            bulk_entries = ShoppingListEntry.objects.filter(
+                Q(created_by=self.request.user) | Q(created_by__in=list(self.request.user.get_shopping_share()))
+                ).filter(space=request.space, id__in=serializer.validated_data['ids'])
+            bulk_entries.update(checked=(checked := serializer.validated_data['checked']), updated_at=timezone.now(), )
+
+            # update the onhand for food if shopping_add_onhand is True
+            if request.user.userpreference.shopping_add_onhand:
```
vabene1111, line 1180: I really don't like this from a performance perspective, looping over an unknown amount of food entries and executing queries each time, but I will let it pass for now to restore the functionality.

smilerz: > I really don't like this from a performance perspective, looping over an unknown amount of food entries and executing queries each time, but I will let it pass for now to restore the functionality.

AFAIK there isn't any other way to update a ManyToMany relationship.  I did spend a lot of time looking for a solution though and best I could find is to introduce queueing.

vabene1111: you can actually bulk create relations using the trough model (I did that for open data) but I am generally not a friend of the whole system of user on hand, shared, stock, whatever, there are so many slow and inefficent queries in the shopping system, there needs to be a general overhaul at some point.

PR #3055, opened 2024-03-21
Improve docker run speeds

tooboredtocode:
Alters the docker image and boot script so that static files are collected ahead of time. This massively speeds up the start up as now this step is simply copying the already collected files.


boot.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/sh
-source venv/bin/activate
+
+# conditionally activate virtualenv, since the docker container does not need it
```
vabene1111, line 3: why do you feel like that venv is not needed. Its not really slowing anything down / costing anything and there are at least some reasons to use them besides the fact that it has proven to work very well over many years ? 

https://vsupalov.com/virtualenv-in-docker/#:~:text=Docker%20is%20great%20at%20packaging,Docker%20image%20without%20using%20virtualenv.

Dockerfile
```diff
@@ -23,18 +25,24 @@ RUN \
...
 
 #Copy project and execute it.
 COPY . ./
 
+# collect the static files
+RUN python manage.py collectstatic_js_reverse
+RUN python manage.py collectstatic --noinput
+# copy the collected static files to a different location, so they can be moved into a potentially mounted volume
+RUN rm -rf /opt/recipes/staticfiles/*
+RUN mv /opt/recipes/staticfiles /opt/recipes/staticfiles-collect
```
vabene1111, line 42: doesn't this delete the staticfiles first and then tries to move them, the folder should be empty at that point

boot.sh
tooboredtocode: I personally feel like it's an anti pattern that unnecessarily increases the container size, given that docker is already a tool for dependency management. This change could be removed if you believe it is better to do so though.

Dockerfile
tooboredtocode: Yes that was a wrong fix for a problem I encountered. This should've been in the boot.sh file. I have fixed this though.

boot.sh
```diff
@@ -67,12 +71,20 @@ echo "Migrating database"
 
 python manage.py migrate
 
-echo "Generating static files"
+if [[ "${DOCKER}" == "true" ]]; then
+    echo "Collecting static files"
```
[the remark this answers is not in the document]
tooboredtocode: Yeah, these messages were rather impromptu, I have added better ones in a new commit

boot.sh
vabene1111: ok thanks for your feedback on this, I will focus a bit on vue3 first and do some further research at some point. If anyone from the community wants to state their opinion feel free.

tooboredtocode: then lets split the venv removal into another PR, so this PR can be pushed

tooboredtocode:
@vabene1111 the removal of venv was reverted, i will open another PR for this change though once this is merged though, where this can be discussed more thoroughly

vabene1111 (approved):

tooboredtocode:
Hi, will this ever be merged?

vabene1111:
approving something usually means that I will merge it but since there are literally endless things to do I will do everything whenever I feel like its right to do it. So please have some patience, since this is a non critical change. 

Apart from that thank you very much for your contribution.

vabene1111:
reverted this PR as it lead to errors on certain setups. https://github.com/TandoorRecipes/recipes/issues/3099 

can merge it again if plausible reason and solution is found. Otherwise we will just live with a little slower startup for now, can look into build/setup optimizations again for vue3 version.

PR #3063, opened 2024-03-24
Update views.py

alexratman:
Tiny change, appears to have been because of a change between versions, but was stopping access to Settings


alexratman:
Just spotted this is a duplicate and someone's already got a PR in for it

PR #3068, opened 2024-03-27
Feature/vue3

smilerz:
rebased to develop
added permissions to docs/api - requires at least guest
added swagger api view
added authentication method to openapi schema
added logo to docs/api
fixed tree and merge schemas
deleted custom schema generation
pk only on writeable nestable serializer #2970


vabene1111:
no sure why, I did make any commits but this cannot be merged, can you look into that? Also it includes the #2970 PR as well just so we know for the changelog.

Issue #3091, opened 2024-04-01
Values in properties breakdown for a single recipe are not rounded


zodac:
### Tandoor Version

1.5.15

### Setup

Docker / Docker-Compose

### Reverse Proxy

Apache2

### Bug description

I have added nutritional properties to my ingredients, and have a summary at the end of my recipes. However, these calculates values are not rounded (despite my settings for 2 decimal places), and in some cases have floating point issues resulting in very large decimals:
![image](https://github.com/TandoorRecipes/recipes/assets/2502611/f6be3672-cf45-47a8-b28b-d10566fc82a6)

Perhaps some explicit rounding can be performed on this section for better presentation?

Manon-ds:
Hey! I'm going to work on this

PR #3095, opened 2024-04-05
fix: rounding units

Manon-ds:
fix for issue #3091 
Function roundDecimals was not being called in PropertyViewComponent b-modal element so food properties were not being rounded to the specified number of decimals. 

Here is a side by side comparison of the bug before and after the fix, using the same numbers.

Screenshot 2024-04-05 at 15 59 46

Screenshot 2024-04-05 at 15 58 57


vabene1111:
thanks for the PR :)

PR #3102, opened 2024-04-09
Yapf prettier

smilerz:
setup configurations for flake8, isort, yapf and prettier
created documentation for VSCode and PyCharm


vabene1111:
this is awesome, thanks for the great documentation improvement and the formatting system.  One question: Do we also have to put prettier into vue3 package.json so its setup there from the beginning on?

PR #3109, opened 2024-04-12
Compatible with recipe_scrapers v15

smilerz:


vabene1111:
thnaks for the update. What do you think about using an rc version of recipe scrapers? It looks pretty stable and I dont wan't to leave this PR here laying around until it becomes hard to merge on the other hand scrapers is a very important part of tandoor 🤔

smilerz:
It passes all of our tests. I think it's probably fine.

PR #3114, opened 2024-04-18
Feature/vue3

smilerz:


vabene1111:
i am confused, should I merge this? As far as I see it creates an openapi generator in the root folder which doesn't really make any sense right?

smilerz:
No don't merge.  I'll cancel the PR

PR #3121, opened 2024-04-21
Fix #3060 Next Cloud handle special ingredients ## 

### LINKED ISSUE #3060: Nextcloud Importer handle special ingredients ##
### Tandoor Version

1.5.14

### Setup

Docker / Docker-Compose

### Reverse Proxy

No reverse proxy

### Other

_No response_

### Bug description

In Nextcloud the User can add „Ingredients“ with leading „## „. Then will this rendert as subheader, like in Markdown. 
This „ingredients“ should be removed or something better. Splitting into two steps.

### Relevant logs

_No response_

ellezhu1:
Fix #3060

Before and After:

nextCloudBefore1
nextCloudAfter

### DIFF
--- a/cookbook/integration/nextcloud_cookbook.py
+++ b/cookbook/integration/nextcloud_cookbook.py
@@ -65,7 +65,10 @@ class NextcloudCookbook(Integration):
             if ingredients_added == False:
                 for ingredient in recipe_json['recipeIngredient']:
                     ingredients_added = True
-                    if ingredient.find('##') == -1:
+                    if ingredient.startswith('##'):
+                        subheader = ingredient.replace('##', '', 1)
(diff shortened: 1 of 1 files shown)


vabene1111:
thanks for the RP, two ideas:

1. should we use `.startwith()` to make sure lines that just contain a double ## somehwere are not ignored
2. what do you think of instead of ignoring the lines adding them as a header in tandoor (by setting `is_header` to true on the ingredient and setting the text (without the ##) into the `note` field (leaving food/unit None and amount 0)

ellezhu1:
1. I didn't think it was very likely; however, you make a good point. I have gone ahead and changed it to use `.startswith()` 
2. I did consider making them into subheaders but I couldn't figure out how to do it. Your tip about the `note` field was very helpful. I have now implemented the lines as headers.

Thanks for the feedback!

vabene1111:
perfect, thank you. Will get both PR's merged for the next release.

PR #3130, opened 2024-04-29
rename property apis

smilerz:
`api/food-property` --> `api/property`
`api/food-property-type` --> `api/property-type`


vabene1111:
can you discard the client from the PR, I will regenerate after merging, thats easier than resolving the conflicts manually

PR #3132, opened 2024-04-30
removing pre-cached staticfiles from Docker & boot.sh

smilerz:
`collectstatic` and `collectstatic_js_reverse` both have dependencies on `.env` variables.  They can't be built independent of the environment they are going to run in.


vabene1111:
good catch, thanks 👍

PR #3147, opened 2024-05-04
Remove obsolete "version: 3" to eliminate warning message

bhagerty:
The "version: 3" entry is obsolete and throws a warning: 'version' is obsolete. AFAIK, there's no reason to keep it. There's a lengthy discussion of this issue in a different repo: https://github.com/jasonacox/Powerwall-Dashboard/issues/453

They got rid of "version" in that project. It seems like it should be removed here too.

### DIFF
--- a/docs/install/docker/plain/docker-compose.yml
+++ b/docs/install/docker/plain/docker-compose.yml
@@ -1,4 +1,3 @@
-version: "3"
 services:
   db_recipes:
     restart: always


vabene1111:
ok, makes sense, thank you. can you remove the version from all the different compose examples ?

bhagerty:
@vabene1111 - I updated the remaining `compose.yml` files. My PRs for the changes are maybe not perfect, but I think you can accept them. See the comments with the other PRs.

vabene1111:
perfect, thank you very much :)

PR #3220, opened 2024-07-17
show ingredient notes always

jmawet:
Based on feature request: Always Show Ingredient Notes #3129


vabene1111:
hi, thanks for your PR. I undersand that some people might prefer ingredient notes to be always shown but some people (including myself) prefer them not to be shown all the time. This means a setting needs to be created for this feature. 

As I am currently working on the new Vue 3 Version of tandoor this will not happen in the current version so I am closing this PR for now. Once the new version is ready we can look into this again or maybe I implement it while doing the new version because it has been asked for a few times. 

Still thanks for your effort.

PR #3260, opened 2024-08-14
Update unit_conversion_helper.py

yuekaim:
added us cup as a volume measurement


vabene1111:
thanks, I also added it as an option for units to select

Issue #3270, opened 2024-08-23
zh_Hans Simplified Chinese language support


ttimasdf:
### Is your feature request related to a problem? Please describe.

zh_Hans language is currently in a half-way-done state. I would love this application with Chinese interface, though the search and index may still need improvement for the language, but the translated UI could be a start 😄 

### Describe the solution you'd like

I went through most of the untranslated strings and revamped some of the strings that are *too* machine translated.

The current state on weblate: [Tandoor - Chinese (Simplified) @ Tandoor Translate](https://translate.tandoor.dev/projects/tandoor/-/zh_Hans/)

### Additional context

related issue: #2627

smilerz:
Any translations you add will automatically be included in the next release.

vabene1111:
actually the language currently does not show up I am afraid. The problem is that django needs `zh-Hans` as a langauge code while the file by weblate is named `zh_Hans`. I already implemented langauges for the new frontend which will just scan trough the files and use these. This is probably fixable for the old version as well but I dont want to spend time on the old version so I will leave it to the new one, just another motivation to get done with it sooner :)

Thanks for helping with translations btw, appreciate it :)

ttimasdf:
In addition to the translated strings that the Weblate bot will commit, I believe some other adjustments are necessary for it to work. (I may be mistaken, but the current code base just doesn't work for me...)

On the backend, 'zh-hans' could be included in `settings.py` to provide the option in the system settings. However, on the frontend, specifically the Vue side, the language setting does not take effect.

https://github.com/TandoorRecipes/recipes/blob/beb1823a5704003cdbce26219e06188c371976bf/recipes/settings.py#L477-L495

ttimasdf:
> actually the language currently does not show up I am afraid

just saw the comment above, and this's also the problem I see when I try to hack my translation in to our existed docker image. and currently we also use zh_CN for Chinese which is also unsupported by django (`\cookbook\locale\zh_CN\LC_MESSAGES\django.po`)

ttimasdf:
> I already implemented langauges for the new frontend which will just scan trough the files and use these.

and there's a nuance between `zh-Hans` and `zh_Hans` prevents it from loading even though it found the file. which I think is already fixed like you've mentioned?

![image](https://github.com/user-attachments/assets/4d56436c-0007-468f-bc3a-be3101899b50)

vabene1111:
[...]

PR #3279, opened 2024-08-30
Chinese language support

ttimasdf:
Building on the previous point: https://github.com/TandoorRecipes/recipes/issues/3270#issuecomment-2308151353,

I managed to get zh_Hans working in vue 2 (current branch) by symlinking `vue/src/locales/zh.json` to `vue/src/locales/zh_Hans.json` at https://github.com/ttimasdf/tandoor-recipes/commit/63d59ccd58624f8f494149db058867f001ae5a5c . Although it works for me, it prevents Traditional Chinese translations in frontend from being applied.

This change is rather opinionated. since:

1. I, as a native speaker of Simplified Chinese, can read Traditional Chinese seamlessly and I assume native zh_hant speakers may or may not be the same correspondingly.
2. But Traditional Chinese translation also breaks at the moment 😗, (i.e. do not appear as an option in language selection list.)

So, since making simplified Chinese the default of these two languages *may* be a viable workaround before any backport is available, I would respect the opinions of traditional Chinese users and of maintainers.


ttimasdf:
For anyone interested in a Chinese build, please check out the CI builds at `custom` branch in my fork.

需要中文的用户，可以测试一下我分支里面打包好的容器👇

```
docker pull ghcr.io/ttimasdf/tandoor-recipes:custom
```

vabene1111:
thank you very much for the work you have put into this. To be honst your second commend kinda solved the issue for me because I do not plan on making any noteworthy changes to tandoor until Vue 3. This means people can safely use your fork for a few more month (hopefully, work has been a bit slow on my part) and then just switch to my branch of tandoor again as soon as the new version is out (or as soon as they want to). 

Do you think this is a good solution?

vabene1111:
closed due to inactivity, will be fixed in Vue 3 version

PR #3280, opened 2024-08-30
import of gourmet files

vitoller:
* in gourmet export as html and zip folder
* 'menus' are not imported

Note: it appears that the native format '.grmt' does not include the unit for ingredients

### DIFF
--- a/docs/features/import_export.md
+++ b/docs/features/import_export.md
@@ -1,9 +1,9 @@
-This application features a very versatile import and export feature in order 
+This application features a very versatile import and export feature in order
 to offer the best experience possible and allow you to freely choose where your data goes.
 
 !!! WARNING "WIP"
-    The Module is relatively new. There is a known issue with [Timeouts](https://github.com/vabene1111/recipes/issues/417) on large exports.
-    A fix is being developed and will likely be released with the next version.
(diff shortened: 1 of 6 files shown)


docs/features/import_export.md
```diff
@@ -1,9 +1,9 @@
-This application features a very versatile import and export feature in order 
+This application features a very versatile import and export feature in order
 to offer the best experience possible and allow you to freely choose where your data goes.
 
 !!! WARNING "WIP"
-    The Module is relatively new. There is a known issue with [Timeouts](https://github.com/vabene1111/recipes/issues/417) on large exports.
-    A fix is being developed and will likely be released with the next version.
+The Module is relatively new. There is a known issue with [Timeouts](https://github.com/vabene1111/recipes/issues/417) on large exports.
```
vabene1111, line 5: please discard these changes and all others that strip spaces in !!!! warning/.. blocks as it will break the formatting

vabene1111:
thank you very much for the PR, I have no idea why I haven't seen it for such a long time but it looks very good.

I have added a small comment, also I think the integration needs to be added to "integrations.js" for it to show up in the frontend

vabene1111:
I have decided to just fix this myself because I let you wait so long, sorry again, will be out with the next release

PR #3305, opened 2024-09-15
Make HomeAssistant description field optional

ehowe:
Some HomeAssistant Todo entities do not support the `description` field. This includes the default `Shopping List` entity HA creates and one I use through a custom HA plugin: https://github.com/watkins-matt/home-assistant-google-keep-sync.

This PR adds a configuration to the connector configuration to make this field optional. By default, it is `true` to maintain backward compatibility with existing connectors, but the user can disable it if their Todo entity does not support it.


vabene1111:
thanks for the PR. I will review the code and get it merged for the next update. @mikhail5555 if you want you can take a look too :)

vabene1111:
ok thanks @ehowe for the PR and @mikhail5555 for your feedback, to me everything looks correct as well, so getting this merged for the next update.

PR #3323, opened 2024-09-28
Update custom_tags.py

blowk:
Change the protocol to //domain.tld instead of (http/https).
This way if the website is loaded over https, the images are loaded in with javascript.

### DIFF
--- a/cookbook/templatetags/custom_tags.py
+++ b/cookbook/templatetags/custom_tags.py
@@ -196,7 +196,7 @@ def bookmarklet(request):
 @register.simple_tag
 def base_path(request, path_type):
     if path_type == 'base':
-        return request._current_scheme_host + request.META.get('HTTP_X_SCRIPT_NAME', '')
+        return '//' + request._current_scheme_host.split("://")[-1] + request.META.get('HTTP_X_SCRIPT_NAME', '')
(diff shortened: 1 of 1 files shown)


vabene1111:
sorry but I do not understand what this is supposed to improve? Please also provide links to a source explaining that omitting the protocol and just setting // is a good/standard practicce

blowk:
Sometimes there are problems when loading the website over https.
I know it's better to add the protocol, but Django doesn't always handle the right request from the header (in combination with proxies) and serve the javascript worker with http instead of https.

Omitting the protocol let the browser select the protocol depending on the request. 
Without protocol and when the website is going over https, the browser will load the javascript over https so the images are loaded correctly. Otherwise the browser will not serve them, due mixed content if you load the website for example with Cloudflare.

vabene1111:
Hi, thanks for your reasoning. I have thought about this and disagree with the change, sorry. It is possible to configure tandoor with correct headers and protocols in any setup I can think of, so there is no need to let the browser decide any protocol. If the site is loaded over https it should also load all its contents over https. To be honest it should be completly not supported to use anything besides https even in a lokal network envirnment. 

Still thank you for your suggestion, if you need this feature feel free to create a fork for your customization.

Issue #911, opened 2021-09-21
Bring!  integration? 


dekiesel:
**Is your feature request related to a problem? Please describe.**
Hi, I love recipes and I was hoping to be able to use the week plan with my favorite shopping app, Bring! 
**Describe the solution you'd like**

Once I have created a plan for the week I'd like to be able to click a button and then have all the ingredients in my Bring!  list, which I share with my wife. 

I think this could be possible, Bring describes how to do this here: https://sites.google.com/getbring.com/bring-import-dev-guide/web-to-app-integration

**Describe alternatives you've considered**

I have not considered any.

smilerz:
I'll add this to the list for the shopping list rework.

vabene1111:
this is definitely a good idea, generally there is already a shopping list export which some TODO/Shopping list apps can take (for example you could copy it to google keep) but having proper integrations for several popular providers might be something cool.

What i can personally also recommend is try out the shopping features directly in tandoor as well, sharing, live syncing, sorting, merging of same ingredients and everything i can imagine being needed is already implemented and @smilerz is even working on improving this with a big refactoring soon.

dekiesel:
@smilerz

I think manually would be best because I've had the situation a few times that I needed to add something, so it'd be great to be able to do that. If it isn't too much work, of course.

dekiesel:
Ideally, yes. 

Idk if the api allows for this, but maybe a "delete" button is enough.

So I add something to bring, then realize I forgot something and click on delete (in tandoori). I make my changes to the shopping list and click "add" again.

comfreak89:
A Bring! integration also would be awesome for me - for me, that would be THE killing feature to use Tandoor right now.

Would it not maybe better to integrate the "send to Bring!" button at recipe level like at chefkoch.de at the bottom of every recipe?

You could then choose to send it to the Tandoor internal shopping list or to the Bring! list. That would solve the issue with the maybe "leftovers" at the Tandoor shopping list. 

Example: https://www.chefkoch.de/rezepte/2766911428603391/Indisches-Butter-Chicken-aus-dem-Ofen.html

vabene1111:
I think that is a good idea. Also @smilerz oeople appear to be more attached to external shopping list services than I thought, maybe we can build some kind plug in or integration system like on the importers so that we can support different kinds of external services, not just bring

ulimanaio:
[...]

PR #3326, opened 2024-09-30
Update connectors.md

supaeasy:
Fixed errors in description:
- add / after api
- omit 'todo.' when adding HA-Entity

### DIFF
--- a/docs/features/connectors.md
+++ b/docs/features/connectors.md
@@ -29,6 +29,7 @@ Step 1:
 2. Get/create a todo list entry you want to sync too.
 ![Todos Viewer](https://github.com/TandoorRecipes/recipes/assets/7824786/506c4c34-3d40-48ae-a80c-e50374c5c618)
 3. Create a connector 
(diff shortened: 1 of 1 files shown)


vabene1111:
thank you for the update!

vabene1111:
@mikhail5555 can you look into if this should be reverted ?

supaeasy:
Sorry for the hassle!

vabene1111:
thank you for the feedback, going to revert this.

supaeasy:
Thank you for your offer @mikhail5555 I will contact you once I am back from vacation, some 10ish days or so. Maybe you could look into my question [here ](https://github.com/TandoorRecipes/recipes/issues/911#issuecomment-2387229381) about deleting or checking items in the API in the meantime. I am still getting all of my imported recipes into structure but would like to go "productive" with one solution or the other - though I think my way would fit my needs better since it includes all the details I want for bring list.

PR #3327, opened 2024-09-30
Update manual.md with required python version

IkkiOcean:
The code written in https://github.com/TandoorRecipes/recipes/blob/develop/cookbook/connectors/connector_manager.py makes use of features introduced in Python 3.10, such as the structural pattern matching (match and case) and the union types with the pipe operator ( | ) which are not compatible with Python 3.9. That's why the minimum required python version for the manual installation should be python version 3.10.


vabene1111:
thanks for the update!

PR #3339, opened 2024-10-01
Allow custom ACCOUNT_DEFAULT_HTTP_PROTOCOL

tpansino:
Adds a new setting for customizing the Allauth HTTP protocol used for generating links for things like password reset emails.

[Allauth config docs](https://docs.allauth.org/en/dev/account/configuration.html)


vabene1111:
thanks for the PR. I just looked at some confirmation mesages I have and they always use HTTPS. What kind of setup do you have because from what I understand allauth should automaticall detect the protocol the server is using and use that protocol as well for the links.

tpansino:
First - thanks for the tool, and for teaching me about `allauth`. I was a Django dev for a few years, and I love learning about plugins like that.

I can't remember exactly why I needed this back when I set up Tandoor years ago. My only memory is that it had to do with password reset emails and running Tandoor behind a reverse proxy, but I just tried sending an email without that setting and it seems to be working fine. It's possible I had something else misconfigured that was telling Tandoor to not use HTTPS because of the reverse proxy.

I'll close this as not needed after all. Thanks for prompting me to confirm it no longer seems to be an issue.

vabene1111:
perfect, thanks for your feedback :)

PR #3361, opened 2024-10-12
Fix Paprika Import Images when not downloading from URL.

tlambertz:
Previously, the code only tried to use the cropped image from paprika when downloading was attempted and **actively** failed. But if a user takes an image himself, the url field will be empty, thus downloading won't be attempted and won't fail. 

This resulted in no image being imported at all.

This change makes it so that the download is tried first, but the `photo_data` is always tried if no image exists after, for whatever reason.


vabene1111:
awesome, thank you!

PR #3387, opened 2024-11-02
Make postgres data durable across reboots

MarkSRobinson:
By storing it to a real volume, the pg_data will persist across reboots.

### DIFF
--- a/docs/install/truenas_portainer.md
+++ b/docs/install/truenas_portainer.md
@@ -83,7 +83,7 @@ services:
     restart: always
     image: postgres:16-alpine
     volumes:
-      - ./postgresql:/var/lib/postgresql/data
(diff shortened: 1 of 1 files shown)


vabene1111:
Hi, thanks for the pr. The default setup uses bind mounds which are just as persistent as volumes. Since I prefer the bind mount syntax this is the way the default examples are given.

MarkSRobinson:
Are you sure about that? I setup the tooling as described, but when restarting it would wipe the DB. When using these settings, it would preserve the data.

vabene1111:
I have been running countless containers with the bind mount syntax for years and have never los data, so I am sure. Still your choice is absolutely valid so feel free to go with it.

Issue #2857, opened 2024-01-03
Detect duplicate imports from same URL


syphernl:
### Is your feature request related to a problem? Please describe.

You can re-import the same URL again, which results in a duplicate recipe to be created.

### Describe the solution you'd like

If a duplicate URL is encountered can we show a message "This recipe already exists, what do you want to do?" and give them the option to "Import again" (which essentially duplicates the recipe, useful if there are local changes) or "Update" to overwrite the existing recipe with the one from the given URL.

vabene1111:
I see this being quite useful, thanks for the idea

a simple check upon importing against the source_url field should do the trick along with an additional parameter returned by the importer API endpoint

leoniedonskoj:
Hi! I'm a first time contributor. Could I work on this issue?

smilerz:
> Hi! I'm a first time contributor. Could I work on this issue?

Sure, please propose the high level approach before writing any code.

hiirrxnn:
Hey , I'm new . Can i take up this issue?

leoniedonskoj:
Hi. Sorry for asking, but why was my comment with smilerz's answer deleted?

smilerz:
> Hi. Sorry for asking, but why was my comment with smilerz's answer deleted?

An identical comment was added shortly later - I assumed they were both spam (we get a lot of comment spam requesting to contribute)

leoniedonskoj:
Ah okay, that makes sense. I actually want to work on this issue, it's for a project in my bachelor's degree. I'll just need some time looking into the code and figuring out the high level approach you requested.

smilerz:
> Ah okay, that makes sense. I actually want to work on this issue, it's for a project in my bachelor's degree. I'll just need some time looking into the code and figuring out the high level approach you requested.

Looking forward to it!

ellezhu1:
I see that this is still open, and I'd love to take a look. Where are the relevant files located?

lavanyamehn:
I can continue looking at this issue since I see the previous PR has been closed.

lavanyamehn:
@vabene1111 I just wanted to clarify my approach before submitting a PR for this issue. I am planning on making the following change:

- In the API for importing recipes with URL, I will check if the user already has a recipe with that URL. If yes, then I will allow the user to choose between continuing to import a duplicate version of the recipe and editing the existing recipe. The UI for both importing and editing will be rendered like for any other imported recipe.

Please let me know if I missed anything, thanks!

[...]

PR #3412, opened 2024-11-26
Display a warning message when duplicate recipe is imported

lavanyamehn:
This PR addresses #2857

### DIFF
--- a/cookbook/views/api.py
+++ b/cookbook/views/api.py
@@ -1455,13 +1455,21 @@ class RecipeUrlImportView(APIView):
 
             url = serializer.validated_data.get('url', None)
             data = unquote(serializer.validated_data.get('data', None))
+
+            duplicate = False
+            if url:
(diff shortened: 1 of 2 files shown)


cookbook/views/api.py
```diff
@@ -1455,13 +1455,21 @@ def post(self, request, *args, **kwargs):
 
             url = serializer.validated_data.get('url', None)
             data = unquote(serializer.validated_data.get('data', None))
+
+            duplicate = False
+            if url:
+                # Check for existing recipes with provided url
+                existing_recipe = Recipe.objects.filter(source_url=url).first()
```
ambroisie, line 1462: Could use `exists` instead, to avoid fetching the (unused) object.

vabene1111: is changed/reworked for the new version anyway but yes.

Issue #3313, opened 2024-09-20
Update required Python version


NiklasMM:
### Documentation link

https://docs.tandoor.dev/install/manual/

### Affected section

Installation

### Issue description

The documentation for a manual install states the minimum required Python version is 3.9, which is no longer accurate. For example there is code in https://github.com/TandoorRecipes/recipes/blob/develop/cookbook/connectors/connector_manager.py using "match case" syntax which was introduced in 3.10. 

Docker installs seem to use 3.12 and I could not find any other mention of a minimum version in the codebase. So I'm not sure what the minimum required version is.

It would be cool to update the documentation and maybe add the minimum version to the pyproject.toml

vabene1111:
ah sorry, I think required is 3.10 as you found with the match case but I would just put 3.12 (with maybe a reference to the docker file) so people always use the tested latest version.

mc-knife:
if you want to upload images, then you need 3.12, according to [this comment](https://github.com/YunoHost-Apps/tandoor_ynh/issues/89#issuecomment-2481750603).

MostafaElsehy:
Hi, I was looking into this issue and checked the manual installation docs.
It seems the warning has already been updated to state "Make sure to use at least Python 3.12 or higher".
It looks like this was resolved by #3428. Should this issue be closed now?

PR #3428, opened 2024-12-02
Add note about Python 3.12 dependency to manual.md

hawthorc:
3.12 is necessary for image upload, according to the conversation in #3313 .


vabene1111:
hi, thanks for your contribution. This is more clear than before, thanks.

PR #3434, opened 2024-12-09
Add docs for recipe JSON zip structure and conversion process

mygrexit:
So, I couldn’t find any clear docs on how the JSON zip import/export works, so I reverse-engineered (just kidding) it from the files and used ChatGPT/Claude to fit the format. Ended up writing this section on:

- How the zip structure works (export_YYYY-MM-DD.zip > 6.zip > recipe.json + image.jpeg).
- Steps for turning grandma’s old recipes into JSON using OCR and AI tools.

Maybe this section of the docs isn't the right place, since it's more of an how-to, but i wanted to share it anway.
If this looks good, I can polish it further or add more examples. Just an idea, let me know what you think! :)

### DIFF
--- a/docs/features/import_export.md
+++ b/docs/features/import_export.md
@@ -54,6 +54,320 @@ It is maintained with new fields added and contains all data to transfer your re
 It is also one of the few recipe formats that is actually structured in a way that allows for
 easy machine readability if you want to use the data for any other purpose.
 
+## JSON Import/Export
+
+This application features a very versatile import and export feature in order to offer the best experience possible and allow you to freely choose where your data goes.
+
+### Recipe JSON Format
+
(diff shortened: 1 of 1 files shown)


vabene1111:
hi, thanks for your PR, i really like it and I think it can help a lot of people. What I am unsure about is if this is the right place for it. The main reason is that I lack the time to properly maintain something like this trough updates. This means it will at some point become deprecated and might confuse people. 

What do you think about just leaving the section about that JSON import/export is possible in the docs and create a "discussion" in the show and tell area https://github.com/TandoorRecipes/recipes/discussions/categories/show-and-tell which is linked int the docs as something like "community guide to importing and exporting JSON with AI" so people can find it and see the date it was created. Its also easy for people to post about their experiences and share for example their promts.

PR #3448, opened 2024-12-24
Update system.html: Django Recipes > Tandoor

igorsantos07:
I guess this got left out when you renamed the project :)


vabene1111:
thanks, it did in fact got lost there :)

Issue #3494, opened 2025-01-17
Paprika.py imports 0 byte images from photo_url despite response code


mitcdh:
### Tandoor Version

:latest

### Setup

Others (please state below)

### Reverse Proxy

Traefik

### Other

Podman / Quadlets

### Bug description

Hi, I think this is a fairly simple bug related to this code in cookbook/integration/paprika.py:

````python
# Paprika exports can have images in either of image_url, or photo_data.
# If a user takes an image himself, only photo_data will be set.
# If a user imports an image, both will be set. But the photo_data will be a center-cropped square resized version, so the image_url is preferred.

# Try to download image if possible
try:
    if recipe_json.get("image_url", None):
        url = recipe_json.get("image_url", None)
        if validate_import_url(url):
            response = requests.get(url)
            self.import_recipe_image(recipe, BytesIO(response.content))
except Exception:
    pass
# If no image downloaded, try to extract from photo_data
if not recipe.image:
    if recipe_json.get("photo_data", None):
        self.import_recipe_image(recipe, BytesIO(base64.b64decode(recipe_json['photo_data'])), filetype='.jpeg')
return recipe
````

The function validate_input_url only looks to check that the url is valid so the request will proceed and the image imported no matter if it's an empty response. This results in 0 byte images being saved in mediafiles and returned.

I would believe in this case that the embedded photo_data should be preferred rather than nothing.

I've tested the following which provides the desired functionality:

````python
    if response.status_code == 200 and len(response.content) > 0:
        self.import_recipe_image(recipe, BytesIO(response.content))
````

### Relevant logs

```shell

```

PR #3497, opened 2025-01-17
Prevent paprika.py from importing empty image_url content

mitcdh:
Addresses #3494 with proposed resolution.


vabene1111:
yes, that looks good, thank you!

PR #3522, opened 2025-01-30
refactored docs to reflect upcoming changes to vue3;

SlickMoose:
-refactored installation page to be more transparent
-added Vue 3 integration in isntallation page
-added information for django_vite in order django to pull assets from 5173

### DIFF
--- a/docs/contribute/contribute.md
+++ b/docs/contribute/contribute.md
@@ -51,5 +51,4 @@ See [here](/contribute/documentation) for more information on how to add documen
 For the truly ambitious, you can help write code to fix issues, add additional features, or write your own scripts using
 Tandoor's extensive API and share your work with the community.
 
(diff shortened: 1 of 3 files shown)


vabene1111:
thanks for updating the docs. We should change the python version to 3.12 this way we cant forget it at some point.

since this is the documentation for the vue 3 version dont you think it sould be added to the vue 3 branch?

You recommend changing the hot reload options. I dont have much knowledge about that config option, do you think I should change it by default?

SlickMoose:
You are right, it should be merged with vue3 branch, but it would be nice to have them also in gh pages as `https://docs.tandoor.dev/` is the first place where people looking at

Hot reloading is neat, you dont need to rebuild after each change, server reload after you save file. It speed up development tenfold.

PR #3523, opened 2025-01-30
Disable DEBUG and DEBUG_TOOLBAR by default

Honza-m:
The [documentation](https://docs.tandoor.dev/system/configuration/#debug) states that the Django DEBUG mode is disabled by default, and rightly points out that it should not be left enabled in production. 

This change aligns the code with the documentation, ensuring that people who run the app with default .env file don't end up with an unsecure configuration.


vabene1111:
makes sense, thanks

PR #3553, opened 2025-02-17
Add support for Python 3.13

el-han:
Building Tandoor Recipes with Python 3.13 currently fails. I'd suggest the following changes to add support for Python 3.13:

- bump `psycopg2` to 2.9.10 since 2.9.9 does not support Python 3.13 yet
- replace `imghdr` with `filetype` as `imghdr` got removed from Python 3.13 after deprecation


mbarks314 (commented):
No merge

PR #3606, opened 2025-03-21
build: Optimize Dockerfile 

theProf:
> [!NOTE]
> This PR has been moved to https://github.com/TandoorRecipes/recipes/pull/3919

---

## Description
Refactor the Dockerfile for readability, while optimizing for a slimmer final image by reducing the layers and stripping the build dependencies from the production image

## Background
I wanted to start contributing to issues, but got side tracked during setup. This is a work in progress as it will require more testing. The image itself relies on external dependencies not included in the Dockerfile (looking at the CI pipelines) and since I haven't got the dev environment up and running yet, unsure which dependencies are solely for building and which are required for production, but err-ed on the side of caution rather than optimization.

## Comparison of sizes
| Repository                        | Arch                    | Size   |
| --------------------------------- | ----------------------- | ------ |
| theProf:build/docker-optimization | linux/arm64             | 784MB  |
| theProf:build/docker-optimization | linux/amd64             | 157MB  |
| TandoorRecipes:feature/vue3       | linux/arm64             | 2.83GB |
| TandoorRecipes:feature/vue3       | linux/amd64             | 237MB  |
| theProf:build/docker-optimization | linux/arm64,linux/amd64 | 1.56GB |
| TandoorRecipes:feature/vue3       | linux/arm64,linux/amd64 | 3.78GB |
| ghcr.io/tandoorrecipes/recipes    | linux/amd64             | 3.36GB |

## Questions
I didn't see anywhere that was tagged with contribution / discussion around DevOps, CI/CD, building, etc, but if I missed it please feel free to point me in the correct direction! 

EDIT: Created an Issue for larger discussion that can be [found here](https://github.com/TandoorRecipes/recipes/issues/3612)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -107,10 +107,17 @@ RUN <<EOF
 EOF
 
 FROM base AS runner
-COPY --link --from=builder /opt/recipes/ ./
-COPY --link --from=node-builder /src/recipes/vue3/build ./cookbook/static/vue3/
+
+RUN <<EOF
+addgroup -g 1000 tandoor
+adduser -u 1000 -G tandoor -H -D tandoor
+EOF
+
+COPY --link --chown=1000:1000 --from=builder /opt/recipes/ ./
+COPY --link --chown=1000:1000 --from=node-builder /src/recipes/vue3/build ./cookbook/static/vue3/
 
 # This port will be used by gunicorn.
 EXPOSE 8080
+USER 1000:1000
 
 ENTRYPOINT ["/opt/recipes/boot.sh"]


Dockerfile
```diff
@@ -1,44 +1,75 @@
...
+  postgresql-client \
+  postgresql-libs \
+  zlib
 
-ENV DOCKER true
+FROM base AS deps
 
-#This port will be used by gunicorn.
-EXPOSE 8080
+ENV DOCKER=true
```
theProf, line 22: Where is this env used, and for what purpose? didn't find it searching the source

vabene1111:
hi, thanks, i would like to optizmize the build process. To be honest I don't really have dev and production dependencies well sorted, I just add them to docker whenever something fails to build and the mess has been growing for years. 

The vue-3 branch ist the one where all development is happening right now and its probably best to optimize that one from the beginning on so testing can happen gradually while the alpha tests start and nothing on the existing pipeline breaks. 

One issue I have had for years is building for the different architectures. With the vue-3 branch I dropped support for arm/v7 because it was a pain to maintain and would just break every few month. Still there are architecture related issues for other arm build as well which sometimes cause the x86 build, which is the most important one, to fail. Properly separating these two would be great but I think it might require different tags on docker hub, not sure. 

Feel free to play around and ask if you have any questions, would be great if someone could sort this mess.

theProf:
> To be honest I don't really have dev and production dependencies well sorted, I just add them to docker whenever something fails to build and the mess has been growing for years.
no worries, understood. I'll help get them sorted

> The vue-3 branch ist the one where all development is happening right now and its probably best to optimize that one from the beginning
Glad I picked the right branch to base from! 

> so testing can happen gradually while the alpha tests start and nothing on the existing pipeline breaks
Where is the best place to have discussions around the existing pipelines and how they are used? I'll have some general questions on how/where things are used as I get up to speed and wouldn't want to pollute this PR thread is something like an Issue or Discussion with a label would be better for more QA/advice/back and forth? 

> One issue I have had for years is building for the different architectures
It can be tough! Especially if you have any dependencies on libc, then you start getting into glibc (common default of Ubuntu for example) vs musl libc (default of Alpine). I'd be happy to help sort out the multi-arch building. 

One question I did have is: what is the process for getting the development environment setup from scratch? 

I wasn't sure if everything is supposed to be in the Dockerfile after building/running, or if the Dockerfile is just the backend and vue3 needs to be started separately, or if [TandoorRecipes/open-tandoor-data](https://github.com/TandoorRecipes/open-tandoor-data) needed to be pulled in as well, etc.

If you have any sort of "minimum commands to run" to get started, it would be greatly appreciated.

Dockerfile
```diff
@@ -1,44 +1,94 @@
-FROM python:3.13-alpine3.21
+FROM node:20.19-alpine AS node-deps
```
theProf, line 1: what version(s) of `node` should this be built with?

Dockerfile
```diff
@@ -1,44 +1,94 @@
...
 
-#Install all dependencies.
-RUN apk add --no-cache postgresql-libs postgresql-client gettext zlib libjpeg libwebp libxml2-dev libxslt-dev openldap git
+COPY --link ./vue3/package.json yarn.lock ./
 
-#Print all logs without buffering it.
-ENV PYTHONUNBUFFERED 1
+# TODO: use --frozen-lockfile
+# RUN yarn install --frozen-lockfile
+RUN yarn install
```
theProf, line 8: Adding back `--frozen-lockfile` will be a requirement before merge, but uncommenting for now while cleaning up and moving forward

vue3/package.json
```diff
@@ -35,5 +35,6 @@
     "vite": "^5.4.7",
     "vite-plugin-vuetify": "^2.0.4",
     "vue-tsc": "^2.0.26"
-  }
+  },
+  "packageManager": "yarn@1.22.22"
```
theProf, line 39: what version of `yarn` do you want to support? it feels like this should be a monorepo, since `vue3` is referencing `vue`, which would require upwards of yarn v1 (current latest at `yarn@4.7.0`). even if not wanting to use yarn workspaces, upgrading from v1 would give more reliability/performance surrounding node modules

Dockerfile
vabene1111: i dont really care, I usually use some LTS version for years and once it runs out upgrade. I think I have tried everything with 22 on some machine so that should work

vue3/package.json
vabene1111: sure, feel free to upgrade. As with node I have never tought about this version. The `vue` folder will disappear at some point, just left it in because the old stuff is not yet cleaned out.

vabene1111:
> One question I did have is: what is the process for getting the development environment setup from scratch?

clone the repo, install yarn/pip requirements, start the yarn dev server and start the django dev server, run migrations. That should be all thats needed to get a hot reloading dev setup to work.

> discussion

feel free to open an issue, I get notified so I will respond there. Alternatively I could offer a quick discord call to explain stuff. 

Generally speaking the whole system is just "what worked" for the last years, I have tried a few times to optimize for different things but overall there isn't much reason for any particular decision besides that it worked, so feel free to change whatever you want, it just needs to work :)

theProf:
> Alternatively I could offer a quick discord call to explain stuff
That is a very kind offer, I will happily take you up on this! 

> feel free to open an issue
Also did so [here](https://github.com/TandoorRecipes/recipes/issues/3612) (and added link in the PR description), but a realtime chat may be more efficient upfront

Dockerfile
```diff
@@ -1,44 +1,94 @@
...
-    venv/bin/pip install -r requirements.txt --no-cache-dir &&\
-    apk --purge del .build-deps
-
-#Copy project and execute it.
-COPY . ./
+# Print all logs without buffering: https://stackoverflow.com/a/59812588
+ENV PYTHONUNBUFFERED=1
+
+# Install all dependencies.
+RUN apk add --no-cache \
```
wilmardo, line 27:
```suggestion
RUN --mount=type=cache,target=/etc/apk/cache \
  apk add \
```

make use of the 'newish' run volumes to cache the packages inbetween builds. Mostly an easy speedup for local dev.

Dockerfile
```diff
@@ -1,44 +1,94 @@
...
+  postgresql-libs \
+  zlib
+
+FROM base AS deps
+
+ENV DOCKER=true
+
+COPY --link requirements.txt ./
+
+RUN <<EOF
```
wilmardo, line 45:
```suggestion
RUN --mount=type=cache,target=/etc/apk/cache --mount=type=cache,target=/root/.cache <<EOF
```
`/root/.cache` is where the pip cache lives

Dockerfile
```diff
@@ -1,44 +1,94 @@
...
+  if [ $(apk --print-arch) = "armv7" ]; then
+    printf "[global]\nextra-index-url=https://www.piwheels.org/simple\n" >/etc/pip.conf
+  fi
+
+  # remove Development dependencies from requirements.txt
+  sed -i '/# Development/,$d' requirements.txt
+
+  apk add --no-cache --virtual .build-deps \
```
wilmardo, line 53:
```suggestion
  apk add --virtual .build-deps \
```

Dockerfile
```diff
@@ -1,44 +1,94 @@
...
+  /opt/recipes/venv/bin/python -m pip install --upgrade pip
+  venv/bin/pip debug -v
+  venv/bin/pip install wheel==0.45.1
+  venv/bin/pip install setuptools_rust==1.10.2
+
+  if [ $(apk --print-arch) = "aarch64" ]; then
+    curl https://sh.rustup.rs -sSf | sh -s -- -y
+  fi
+
+  venv/bin/pip install -r requirements.txt --no-cache-dir
```
wilmardo, line 83:
```suggestion
  venv/bin/pip install -r requirements.txt
```
related to the caching in the RUN command (proposed change)

Dockerfile
```diff
@@ -48,9 +98,17 @@ COPY . ./
 #            CMD [ "/usr/bin/wget", "--no-verbose", "--tries=1", "--spider", "http://127.0.0.1:8080/openapi" ]
 
 # collect information from git repositories
-RUN /opt/recipes/venv/bin/python version.py
-# delete git repositories to reduce image size
-RUN find . -type d -name ".git" | xargs rm -rf
+RUN <<EOF
+  /opt/recipes/venv/bin/python version.py
+  # delete git repositories to reduce image size
+  find . -type d -name ".git" | xargs rm -rf
```
wilmardo, line 104: I would add .git in the `.dockerignore` file instead of removing it. This is kinda meh since this only removes it from the final image the `.git` is already in the COPY layer which would be avoided when using the dockerignore.

Dockerfile
```diff
@@ -1,44 +1,94 @@
...
+  echo -n "INPUT ( libldap.so )" >/usr/lib/libldap_r.so
+
+  python -m venv venv
+  /opt/recipes/venv/bin/python -m pip install --upgrade pip
+  venv/bin/pip debug -v
+  venv/bin/pip install wheel==0.45.1
+  venv/bin/pip install setuptools_rust==1.10.2
+
+  if [ $(apk --print-arch) = "aarch64" ]; then
+    curl https://sh.rustup.rs -sSf | sh -s -- -y
```
wilmardo, line 80: Couldn't this be replaced by a simple `apk add rust`? 
The package is available:
https://pkgs.alpinelinux.org/packages?name=rust&branch=v3.21&repo=&arch=aarch64&origin=&flagged=&maintainer=

Then it is nicely within the virtual-deps as well.

Dockerfile
wilmardo: related to the caching in the RUN command (proposed change)

wilmardo (commented):
Great work @theProf <3

Did a quick review with some other newish Docker syntax, nice use of the new `--link` already!

Feel free to ping for any questions or a final review when you are done. I now just did a quick review in the Github interface no local checkout.

It would IMO also be nice to setup an unprivilged user and switching to that. Quick copy paste to setup a 1000:1000 user and group without `useradd` or `adduser` (and remembering the different syntaxes hehe):

```Dockerfile
FROM alpine as builder
RUN <<PASSWD cat > /etc_passwd && <<GROUP cat > /etc_group
app:x:1000:1000:app:/:
PASSWD
app:x:1000:app
GROUP

FROM alpine

COPY --from=builder /etc_group /etc/group
COPY --from=builder /etc_passwd /etc/passwd

USER app
```

theProf:
> Great work @theProf <3
> 
> Did a quick review with some other newish Docker syntax, nice use of the new `--link` already!
> 
> Feel free to ping for any questions or a final review when you are done. I now just did a quick review in the Github interface no local checkout.
> 
> It would IMO also be nice to setup an unprivilged user and switching to that. Quick copy paste to setup a 1000:1000 user and group without `useradd` or `adduser` (and remembering the different syntaxes hehe):
> 
> ```dockerfile
> FROM alpine as builder
> RUN <<PASSWD cat > /etc_passwd && <<GROUP cat > /etc_group
> app:x:1000:1000:app:/:
> PASSWD
> app:x:1000:app
> GROUP
> 
> FROM alpine
> 
> COPY --from=builder /etc_group /etc/group
> COPY --from=builder /etc_passwd /etc/passwd
> 
> USER app
> ```

Thanks for the review! I'm glad I kept this as a Draft PR as I feel it best to split it. I've realized there are 2 goals and want to try my best not to mix them (as it will end up being a large and difficult to review PR):
1. Update and organize the Dockerfile to build a production ready image with clarity on dependencies and build steps
2. Optimize the build

I mention as I'm 100% on board for things like RUN mount caches (have it in a local branch), but that falls under 2 and will have impacts to the build pipeline as well (to actually utilize the cache). The reason I want to focus on 1 is to gain clarity on things like:  https://github.com/TandoorRecipes/recipes/pull/3606/files#r2033544477
> I would add .git in the .dockerignore file instead of removing it

I thought the same! I then found out there _may_ be git submodules used by the pipelines [such as here](https://github.com/TandoorRecipes/recipes/blob/develop/.github/workflows/build-docker-open-data.yml#L43). And that's the kind of thing I'd like to move into the Dockerfile as a source of truth on all dependencies and build steps.

Once the Dockerfile is organized and a reliable source of truth, it will be much easier to add optimizations with confidence.

boveloco (commented):
Can we take the time to set a user to dockerfile so it won't run as ROOT?

Dockerfile
edysli: I agree. Removing files added in an earlier layer doesn't actually remove them from the image. The files are still there even if they aren't visible, so it doesn't reduce the container image's size.

edysli:
> IMO the non root running of the image would be part of this. At this moment you can't run this image on Kubernetes where Pod Security is enforced.

I'm in that situation too, trying (and so far failing) to run everything under the _restricted_ [Pod Security Standard](https://kubernetes.io/docs/concepts/security/pod-security-standards/). So I'd very much appreciate the non-root user.

@wilmardo actually, a user account doesn't even need to exist in the container image. `USER 10000:10000` is enough. That's how we run our containers at work and it's not an issue unless the application tries to look up the username it's running under.

edysli:
Thank you @theProf ! :smile: I think those are good improvements to the clarity of the container build process.

theProf:
Sorry I haven't been responsive, had some personal life going on. Great to see all the activity and feedback!

I'll take a crack at these (and creating a checklist to split things up) tomorrow morning.

theProf:
Took a bit longer as I had to get reacquainted with the code. Only pushed a rebase from feature/vue3. But putting update in the issue

wilmardo:
@theProf this has probably been closed by accident when the branch got deleted. Would you mind reopening?

theProf:
> @theProf this has probably been closed by accident when the branch got deleted. Would you mind reopening?

thanks for the nudge. initially thought it was closed out intentionally. Have had some personal life events so haven't been so active. I'll open this back up and likely stack some changes with @wilmardo PRs [ref](https://github.com/TandoorRecipes/recipes/pull/3904)

PR #3611, opened 2025-03-22
Remove maxEntries from image cache in sw.js

mikhail5555:
I believe the hardcoded 20 that was in there is dated, browsers nowdays should be smart enough to know how many images they can/cant cache.

### DIFF
--- a/vue/src/sw.js
+++ b/vue/src/sw.js
@@ -36,12 +36,7 @@ setCatchHandler(({event}) => {
 registerRoute(
     ({request}) => request.destination === 'image',
     new CacheFirst({
-        cacheName: 'images',
-        plugins: [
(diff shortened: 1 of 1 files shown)


theProf:
have you had any issues or noticed slow downs due to this config?

vabene1111:
since this is a "manual" chache and not part of the browsers automatic caching mechanism I limited it back then because it will basically become infinitely huge over time if not limited (its just local storage). This also should not affect the normal browser image chaching logic, its just meant for offline usage. 

To be honest the whole chache stuff to show when the user is offline is rather hackily put together and does not work super well. I do plan on re-visiting this at some point but not sure when I will have the time to do so. For now I would rather not change anything because I dont have the capacity to manage that change.

PR #3614, opened 2025-03-25
docs: mobile apps updated

stritti:
The refered mobile app is discontinued and another one is suggested there.


vabene1111:
awesome, thanks for the PR

PR #3637, opened 2025-04-08
feat: introduce STATIC_ROOT env var

wilmardo:
You like the already existing `MEDIA_ROOT`

My usecase is to have both the media and staticfiles root outside of the `/opt/recipes/` directory to avoid `subPath` mounting on Kubernetes.

It felt missing since `MEDIA_URL` and `MEDIA_ROOT` exist but only `STATIC_URL`.

### DIFF
--- a/docs/system/configuration.md
+++ b/docs/system/configuration.md
@@ -132,6 +132,13 @@ This can either be a relative path from the applications base path or the url of
 STATIC_URL=/static/
 ```
 
+#### Static root
+
+> default `<basedir>/staticfiles` - options `/some/other/media/path`.
+
+Where staticfiles should be stored on disk. The default location is a
+`staticfiles` subfolder at the root of the application directory.
+
 #### Media URL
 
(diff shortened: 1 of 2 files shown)


wilmardo:
@vabene1111 rebased against the `feature/vue3` branch since I am also testing with the latest 2.0.0-alpha release. Seems that branch is the most up to date as well.

Let me know if not correct :)

vabene1111:
hi, thanks for the PR. We can get this merged, I see no reason not to allow people changing this. 

The only thing I am worried about: you changed "staticfiles" to `os.path.join(BASE_DIR, "staticfiles")` as the default. This should probably be the same for almost everyone but I am a tiny bit nervous to break someones setup by changing this. On the other hand it feels like it makes sense to use the base dir. 

What do you think?

wilmardo:
Thanks for the review! Are you sure I am changing something as default? I previously couldn't be changed so that is now the default for then the env var isn't set.

Or am I overlooking something?

vabene1111:
you are correct, I just looked wrong. Sorry for the confusion.

PR #3645, opened 2025-04-10
add small chip indicating if you are selecting existing item or creaing a new one

mikhail5555:
Before:
https://github.com/user-attachments/assets/74f1be49-5fc6-4a0a-80e0-330b0b76e3d1

After:
https://github.com/user-attachments/assets/37718952-8b56-41d7-a924-1f5b4f373d1e

### DIFF
--- a/vue3/src/components/inputs/ModelSelect.vue
+++ b/vue3/src/components/inputs/ModelSelect.vue
@@ -45,6 +45,12 @@
               </v-chip>
             </div>
           </template>
+
+          <template #clear="{ clear }" v-if="props.canClear">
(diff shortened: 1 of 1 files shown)


vabene1111:
ah that is good, I tought I wanted to check if a custom slot works but must have forgotten. Thanks for the PR and effort in fixing this!

I will probably change it so on mobile just an icon is displayed. Do you think existing should always be written? It looks nice on the other hand its kinda clear and I think the old behavior was to only show new ones right ?

vabene1111:
perfect, when shall I send you the list of the other things for the expert 😂 but jokes aside, thanks a lot I think this really improves the component. 

I will merge this for the next update :)

Issue #3676, opened 2025-04-26
Alpha 2 – Advanced search not working


jonaskohl:
### Tandoor Version

885a3169558fb978af74d0370e0279997c962c9a

### Setup

Docker / Docker-Compose

### Reverse Proxy

Others (please state below)

### Other

nginx 1.22.1

### Bug description

The advanced search page does not work. It shows an empty page. It worked in version b08c119e57d1d257bd5fa1e622683addb2199c10

### Relevant logs

```shell
ReferenceError: VNumberInput is not defined
    setup https://rezepte.XXXXXXXXXX.de/static/vue3/assets/SearchPage-CnTG2U7r.js:1
    callWithErrorHandling https://rezepte.XXXXXXXXXX.de/static/vue3/assets/main-BEbQZ3zs.js:17
    setupStatefulComponent https://rezepte.XXXXXXXXXX.de/static/vue3/assets/main-BEbQZ3zs.js:18
    setupComponent https://rezepte.XXXXXXXXXX.de/static/vue3/assets/main-BEbQZ3zs.js:18
    j https://rezepte.XXXXXXXXXX.de/static/vue3/assets/main-BEbQZ3zs.js:18
    B https://rezepte.XXXXXXXXXX.de/static/vue3/assets/main-BEbQZ3zs.js:18
    S https://rezepte.XXXXXXXXXX.de/static/vue3/assets/main-BEbQZ3zs.js:18
    le https://rezepte.XXXXXXXXXX.de/static/vue3/assets/main-BEbQZ3zs.js:18
    run https://rezepte.XXXXXXXXXX.de/static/vue3/assets/main-BEbQZ3zs.js:10
    runIfDirty https://rezepte.XXXXXXXXXX.de/static/vue3/assets/main-BEbQZ3zs.js:10
    callWithErrorHandling https://rezepte.XXXXXXXXXX.de/static/vue3/assets/main-BEbQZ3zs.js:17
    flushJobs https://rezepte.XXXXXXXXXX.de/static/vue3/assets/main-BEbQZ3zs.js:17
    promise callback*queueFlush https://rezepte.XXXXXXXXXX.de/static/vue3/assets/main-BEbQZ3zs.js:17
    queuePostFlushCb https://rezepte.XXXXXXXXXX.de/static/vue3/assets/main-BEbQZ3zs.js:17
[...]

PR #3677, opened 2025-04-26
Fix missing VNumberInput import on advanced search page

jonaskohl:
See #3676


vabene1111:
thanks for the PR! I manually fixed it because the number input is no longer in labs and can/must be imported from the main components directory.

PR #3695, opened 2025-05-03
Support reading S3_SECRET_ACCESS_KEY from file

tpansino:
(I can re-target this against `feature/vue3` if that would be better.)

Thank you for the excellent project! Looking forward to the Vue 3 UI.


vabene1111:
thanks for the PR, will get it merged into the base branch and then merge that with vue3

PR #3707, opened 2025-05-10
Fix doc updating postgress

EifX:
Fixed the documentation regarding updating postgres. Following changes were made:

- removed `sudo` to be more uniform in this documentation. The user should know in which right environment they had installed tandoor
- updated the `docker-compose` executable to `docker compose`
- added to step 6 the very neccessary information, that only `db_recipes` should be started in this point


vabene1111:
thank you, looks good to me. I will get it merged with one of the next updates

PR #3724, opened 2025-05-19
Feature/connector manager bulk insert

mikhail5555:
image
Checked by running locally, deleting/adding individual items also still works as expected

### DIFF
--- a/cookbook/connectors/connector.py
+++ b/cookbook/connectors/connector.py
@@ -11,10 +11,10 @@ class UserDTO:
     first_name: Optional[str]
 
     @staticmethod
-    def try_create_from_user(instance: Optional[User]) -> 'UserDTO':
+    def create_from_user(instance: User) -> 'UserDTO':
         return UserDTO(
             username=instance.username,
-            first_name=instance.first_name if instance.first_name and len(instance.first_name) > 0 else None
+            first_name=instance.first_name if instance.first_name else None
         )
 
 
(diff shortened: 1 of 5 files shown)


cookbook/connectors/connector.py
```diff
@@ -1,6 +1,43 @@
...
+from cookbook.models import ShoppingListEntry, User, ConnectorConfig
+
+
+@dataclass
+class UserDTO:
+    username: str
+    first_name: Optional[str]
+
+    @staticmethod
+    def try_create_from_user(instance: Optional[User]) -> 'UserDTO':
```
x-stp, line 14: Minor: This will throw if `instance` is ever `None`. Consider guarding early and returning `None`, or enforcing a stricter type upstream.

cookbook/connectors/connector.py
```diff
@@ -10,20 +47,18 @@ def __init__(self, config: ConnectorConfig):
         pass
 
     @abstractmethod
-    async def on_shopping_list_entry_created(self, space: Space, instance: ShoppingListEntry) -> None:
+    async def on_shopping_list_entry_created(self, instance: ShoppingListEntryDTO) -> None:
```
x-stp, line 50: Nice: much cleaner to pass DTOs instead of raw ORM objects, especially in async context.

Consider documenting in the docstring that the DTO may omit related fields (e.g. `created_by`) if not available or preloaded..

cookbook/connectors/homeassistant.py
```diff
@@ -88,19 +81,19 @@ async def close(self) -> None:
...
-            item += f" {shopping_list_entry.unit.name})"
+        if shopping_list_entry.base_unit:
+            item += f" {shopping_list_entry.base_unit})"
+        elif shopping_list_entry.unit_name:
+            item += f" {shopping_list_entry.unit_name})"
         else:
             item += ")"
 
     description = "From TandoorRecipes"
-    if shopping_list_entry.created_by.first_name and len(shopping_list_entry.created_by.first_name) > 0:
```
x-stp, line 103: Small nit: `len(...) > 0` is redundant in py: non-empty strings are already truthy.

Also: if this runs against a DTO with `created_by=None`, it will raise an `AttributeError`. Might be worth guarding earlier or clarifying that `created_by` is always expected.

Could alternatively use structural matching (3.10+):

```python
match shopping_list_entry.created_by:
    case UserDTO(first_name=str() as name) if name:
...
```

vabene1111: but the name is actually optional, you can just call the models get_label (or something like that) to get how the username should look (either username or if given first/last name)

vabene1111 (approved):

x-stp:
LGTM

Issue #382, opened 2021-02-03
time display option


Tuetenk0pp:
**Is your feature request related to a problem? Please describe.**
Sometimes you have a recipe with ridiculous amounts of time going into it, e.g. a Baguette recipe with 72h proofing time.

**Describe the solution you'd like**
I'd like to have an option to turn on time conversion so that whenever I give a time in minutes, it would display a readable equivalent:
45 min -> 45 min
90 min -> 90 min
150 min -> 2 h 30 min
120 min -> 2 h
4320 min -> 72 h

**Describe alternatives you've considered**
Fractions:
150 min -> 1 1/2 h

**Additional context**
Wether you would like 90 min to display as 90 min, 1h 30 min or as 1,5h is just personal preference. Or is there a standard somewhere that this could follow?

vabene1111:
yes this is definitely something that needs to be added to make longer waiting times easier to read. Its probably also very easy to implement but i haven't gotten around to it yet. I think the standard would be that everything over 60 minutes is converted to hours and minutes.

Thanks for reporting.

hotdog218:
I also would love to see this implemented! Would be nice for bread recipes that have lots of proofing stages.

wisdomkhan:
I would like to get started on this one please. Guide me.

vabene1111:
You will need to find the appropriate code location (likely recipe view. Vue) and write a utility function to convert the displayed time to a more convenient to read format.to test your changes look at the link I send your previously.

wisdomkhan:
https://github.com/vabene1111/recipes/blob/493d553050e9ebd581f6c4a67b70d860b50a34c2/cookbook/views/views.py#L154

This one?

vabene1111:
No you want to work on a frontend component, those are either HTML or vue files. The particular one you are looking for is  Recipe View.vue

If you are not sure how to implement this into the application you could just write the Javascript function and PR it to utils.js and I can implement it into the frontend later

wisdomkhan:
`function timeConvert(n){
    var num = n;
    if (num <= 90) {
        return num + " min";
    }
    else {
        var hours = num / 60;
        var rhours = Math.floor(hours);
        var mins = (hours - rhours) * 60;
        var rmins = Math.round(mins);
        if (mins == 0) {
            return rhours + " h";
        }
        else {
            return rhours + " h " + rmins + " min";
        }
    }
}`

Something like this?

vabene1111:
[...]

PR #3751, opened 2025-06-06
Improved timer finish time display: adding "+Xd" suffix for recipes with ridiculous amounts of time

mkieledunsche:
as discussed in  #382


vabene1111:
thank you. Please localize this as discussed in the issue. Not sure if the format works for all cultures/languages, maybe an icon could also work.

PR #3815, opened 2025-07-07
Cookbookapp images import

caffeinated-tech:
The current version of the cookbookapp integration imports the cookbookapp logo, instead of the actual recipe image.
My change filters out any branding images and imports the first valid image instead.

The  changes to `devcontainer/Dockerfile' were required to run the devcontainer environment. See the comment for details.

I also fixed broken links in the contributor documentation while reading it.

If any of these aren't desired, I can remove them from the PR.

Thanks,
Liam


.dockerignore
```diff
@@ -29,3 +29,4 @@ vue/babel.config*
 vue/package.json
 vue/tsconfig.json
 vue/src/utils/openapi
+venv
```
vabene1111, line 32: i dont really think venv should be ignored because all dependencies are installed in venv when the conainer builds and when we ignore that layer it wont make it into the final image I think ?

vabene1111 (changes requested):

vabene1111:
thanks for the PR, I have one comment as mentioned above, everything else looks good to me

vabene1111 (approved):

caffeinated-tech:
@vabene1111 I've tried setting up the dev container locally to run the tests which are failing, but am having some difficulty with my setup.

Having read the failing test, I'm not sure how my changes could have impacted it. Is there a chance it was an intermittently failing test?

Here is the failed CI step for reference: https://github.com/TandoorRecipes/recipes/actions/runs/16127433006/job/45693933220?pr=3815

vabene1111:
dont worry about the test, if I don't think you changed anything that could break it, might be they were just broken when you branched. 

Its approved so will get this merged for one of the next releases

Oettinger77:
I just updated Tandoor to beta 5 and tried this import, but it failed.
0 from 0 recipes.

Cookbookapp updatet their site last week or so, could that be the cause?


Regards

caffeinated-tech:
@Oettinger77 There is a good chance the CookBookApp update broke something. There isn't a specific scraper for cookbookapp, so the default one from https://github.com/hhursev/recipe-scrapers is being used. Could you share a sample recipe export which is failing for you? I'm hoping to write a cookbookapp scraper eventually and this would be useful.

Oettinger77:
@caffeinated-tech I just read in the documentation "CookBookApp can export .zip files containing .html files."
Now i have 3 options for export, in my first try i chose cookbook files.
image

Export of 137 recipes as cookbook files is 287 kb big, txt 112 kb and PDF 21 mb.
What do you need? :)

Regards

caffeinated-tech:
@Oettinger77 I've just taken a look over those three export formats and it doesn't look promising. The old HTML export was great, but with these 3 only the PDF export includes all info along with any photos.

No need for you to pass on any exports, I can test with my own files which are giving the same results.

vabene1111:
some sites have the option to export different formats, maybe you can choose between HTML and PDF?

Sadly some apps purposely use cryptic/non machine readable formats to lock their users into their app which is really sad. You can probably use the AI importer but that's unnecessarily complicated

PR #3822, opened 2025-07-14
FoodEditor margin fix

SlickMoose:
added no-gutters to remove negative margin from food edit card title;
added padding left 4px to indent food icon & title;


vabene1111:
hi, thanks for the PR. I like the change in general but preferred it with pl-2 instead of pl-4 so I manually made the changes.

Issue #3836, opened 2025-07-18
ingredient editor dropdowns (including unit, ingredient,grocery list) completely white in dark theme


jstrvr:
### Tandoor Version

beta-5

### Setup

Docker / Docker-Compose

### Reverse Proxy

Nginx Proxy Manager (NPM)

### Bug description

<img width="683" height="261" alt="Image" src="https://github.com/user-attachments/assets/1c82fd42-8428-4812-a999-0c3efa5f3655" />

Light theme seems to work fine.

<img width="446" height="222" alt="Image" src="https://github.com/user-attachments/assets/27cf8239-160e-4c28-ba5f-abca38bab134" />

### Relevant logs

```shell

```

vabene1111:
fixed

PR #3848, opened 2025-07-25
fix: improved dark mode contrast for dropdowns in ShoppingListView

meeraiswar:
This PR addresses issue #3836 by applying scoped dark-mode styles to dropdowns in `ShoppingListView.vue`.

✔️ Sets background to `#2b2b2b` and text to `#f0f0f0`  
✔️ Ensures visual consistency and accessibility in dark theme  
✔️ Targets Vuetify components (`v-select`, `v-input`, etc.) used in grocery list UI

### DIFF
--- a/vue3/src/components/display/ShoppingListView.vue
+++ b/vue3/src/components/display/ShoppingListView.vue
@@ -378,5 +378,21 @@ function loadSupermarkets() {
 </script>
 
 <style scoped>
+/* Force dark mode styling for dropdowns */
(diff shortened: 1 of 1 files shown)


vabene1111:
i cannot reproduce your issue on the version you mention. Even if, the fix you provided does not apply to the all the components you posted pictures of. 

Please check if you might had an older version or some chache issues, should be fine with beta 5 i think

PR #3861, opened 2025-07-31
Fixed k8s setup for tandoor 2

MyDigitalLife:
I'm running Tandoor in k8s with the default setup described in the documentation. As its not tested i noticed It didn't work after updating to tandoor 2. 

Checking the code the init container has the collectstatic_js_reverse command. The dependency that was needed was removed from the requirements.txt with the upgrade so it worked after removing the command in the init container.

### DIFF
--- a/docs/install/k8s/50-deployment.yaml
+++ b/docs/install/k8s/50-deployment.yaml
@@ -63,7 +63,6 @@ spec:
           source venv/bin/activate
           echo "Updating database"
           python manage.py migrate
-          python manage.py collectstatic_js_reverse
           python manage.py collectstatic --noinput
           echo "Setting media file attributes"
           chown -R 65534:65534 /opt/recipes/mediafiles


MeneDev:
Can confirm I have it running like that now.

I haven't followed the v2 development, but isn't nginx now part of the image? A bit unclear to my, why we still need it?
When I look at the diff in https://github.com/TandoorRecipes/recipes/commit/7ccdb90f9b5555d2ea5dfe3c0a3aa78532d5a845 for the "plain" docker-compose setup, I would expect quite a simplification here as well?

vabene1111:
perfect, yes this command is no longer required. @MeneDev is correct, an additional nginx is no longer needed, users will only need a reverse proxy (if they want).

Do you want to review the setup regarding this or should I merge the current state?

Thanks for the PR :)

MyDigitalLife:
I'll have a look tonight to review the setup.

MyDigitalLife:
Had a look and there are a few issues that I noticed. Some are fixed in this pr: https://github.com/TandoorRecipes/recipes/pull/3904. After this is merged I can have a second look and see if there are any other issues to make this setup work again without the dedicated nginx config.

edysli:
The `ServiceAccount` resource can also be removed because it is unnecessary. This application doesn't need to talk to the k8s API server.

pfaelzerchen:
@MyDigitalLife I tried the same last weekend. The problem is, that you seem to need to start nginx from a root process within the recipes container. Otherwise, it won't run as nginx user and therefore won't have write access to the nginx-folders and thus fail to start.

But when the container needs root privileges it probably does not matter to extract the init to a rootful container, and you can deploy everything just as in the docker example - or stay with a seperate nginx container (which can run unprivileged and with no write access).

PR #3890, opened 2025-08-03
Add Ability to Split/Merge Steps in Recipe Editor View

### LINKED ISSUE #3802: Add way to merge recipe steps in editing mode

There does not seem to be a way to merge recipe steps (and their ingredients) when editing a recipe.

Add a feature to not just move down or delete a step but to merge it and it's associated ingredients with the one below/above just like it is possible when initially importing it from a URL.

_No response_

### Additional context

_No response_

birthdayattack:
Closes #3802 

This adds buttons for: 
- Splitting step into individual steps based on newline
- Merging all steps into one step

into the recipe editor view. This reuses the logic already present in the recipe importer view.


vabene1111:
thanks for the PR. Since the import Recipe and the Recipe in the Editor share the mostly the same datastructure can you please verify if the same functions can be used to perform the operations because duplicating the functions is not good. 

Best would be in some utility file/class.

birthdayattack:
Good point. I've pulled out the logic into a separate step_utils file.

The Step and SourceImportStep interfaces are similar, but separate, and those files are autogenerated by OpenAPI. So I added a StepLike interface into the same step_utils, to act as a generic Step object. 

I also went ahead and updated the merge code to use map and flatMap instead of the existing forEach. I noticed a newline gets appended to the merge instruction each time you click the button.

vabene1111:
perfect, thats exactly what I preferred as a solution.  Will get this merged with the next update

vabene1111:
hi, for some reason the GitHub diff viewer became really bad and its hard for me to see where the changes were made. Would you mind merging your branch locally and updating based on the changes I have made to the RecipeEditor so I can merge more easily?

birthdayattack:
hm, looks like the diff view is still being difficult. I did merge the changes locally. Let me know if there are any formatting reasons why the diff might be confused. I'm using the formatting extensions [mentioned in the docs for VSCode.](https://docs.tandoor.dev/contribute/vscode/)

vabene1111:
awesome, thank you 👍

vabene1111:
ok something broke my ability to merge this or at least properly see the diffs. Implemented this manually now. Thansk still for the PR

PR #3904, opened 2025-08-05
Docker optimizations

wilmardo:
Couple quick fixes in the Dockerfile where ee7d6110862a3518ace46e4d426651faa0ed64a0 is the most important one.

In the current setup there is a possibility for zombie processes since Docker only cares about the PID1 and other processes might not get killed since Bash isn't the greatest in forwarding the signals.

Using a small processmanager like tini solves all these issues :)

The rest are just little convenience fixes or little cleanups (armv7 support dropped, rust works fine on aarch64 from repo now)

Will followup this PR with some more optimizations but this is the most important one.


edysli:
Thanks for starting to improve the container image! :smile: I think there is more to clean, but at least that's a start.

And yes indeed, having a proper init process is crucial here.

wilmardo:
@edysli yes I was very temped to a big refactor but that is for after my vacation ;)

I will for sure follow this up with more PRs

vabene1111:
Thanks for the PR, looks good on first sight. I have a bit of vacation now as well but will definitely get back to this.

PR #3905, opened 2025-08-05
fix: remove python-ldap workaround, not needed anymore since openldap >3.4.2

wilmardo:
As seen in the releasenotes:
>This is a minor release to provide out-of-the-box compatibility with the merge
of libldap and libldap_r that happened with OpenLDAP's 2.5 release.

https://github.com/python-ldap/python-ldap/releases/tag/python-ldap-3.4.2

Tandoor is currently on 3.4.4 so this shouldn't be needed anymore. This is just based on the docs, please test this :)


vabene1111:
manually merging this as merging the other one broke this, thanks for the PR and testing   👍

Issue #3612, opened 2025-03-22
[Code] Improvements to project toolchain and build system


theProf:
### Is your feature request related to a problem? Please describe.

With the migration to vue3, it seems like a good time to modernize the project toolchain and build system as well. The project currently doesn't have well defined / explicit dependencies and are set on a 'what gets it working basis'. Excellent to get up and running, but leads to less reliable / stable development and building as the project ages. 

### Describe the solution you'd like

I have more experience in the JS ecosystems than the python ones, and will lean towards that side in solutions.

Proposed changes (that I would participate in implementing):
1. Toolchain versioning within the code
    1. Examples would be defining the package manager with version and supported node engines within the package.json
1. Upgrade Yarn
    1. Current version is `yarn@4.7.0`
    1. Support for monorepos, but this isn't a necessity. But would help with separation of concerns and cross project dependency management
    1. More efficient with package management, even when using more classic `nodeLinker: node-modules` for better reliability, smaller `node_modules` size (resulting image size), and ease of upgrading dependencies
1. Docker upgrade and optimization ([initial PR Draft found here](https://github.com/TandoorRecipes/recipes/pull/3606)]
    1. This will also help serve as documentation of dependencies and steps for building / development
    1. Help slim down the resulting production image
    1. Increased stability and ease of upgrading deps
    1. Will help with multi-platform image building along with point 3
1. CI/CD polishing
    1. Will need some changes to support the above changes
    1. The above changes lend to having faster, more stable, and multi-platform builds

This is pretty much a project in and of itself, but would aid in both lowering barriers for user contributions to code, as well as, making a more reliable and efficient running image.

### Additional context

Wanted a single location for open discussion so created this issue to bring it out of the initial PR.

Personal Note:
This has been a god send of an App for my not-a-good-cook ADHD self, and would love to give back to something that's improved my life. Thank you to everyone who has had a hand in the creation!

vabene1111:
Thank you for the kind words and your willingness to help. As discussed in the PR lets have a quick chat on discord for the initial start, my name is vabene1111, you can also find me over the tandoor discord.

edysli:
Is there a pythonic way of distinguishing between development and production dependencies in `requirements.txt`? That would probably be more convenient than deleting the dev dependencies from this file at build time.

smilerz:
[...]

PR #3919, opened 2025-08-08
 build: Optimize Dockerfile

theProf:
## Description
Refactor the Dockerfile for readability, while optimizing for a slimmer final image by reducing the layers and stripping the build dependencies from the production image

## Background
I wanted to start contributing to issues, but got side tracked during setup. This is a work in progress as it will require more testing. The image itself relies on external dependencies not included in the Dockerfile (looking at the CI pipelines) and since I haven't got the dev environment up and running yet, unsure which dependencies are solely for building and which are required for production, but err-ed on the side of caution rather than optimization.

## Discussion
Created an Issue for larger discussion that can be [found here](https://github.com/TandoorRecipes/recipes/issues/3612)

## Questions
I didn't see anywhere that was tagged with contribution / discussion around DevOps, CI/CD, building, etc, but if I missed it please feel free to point me in the correct direction! 

##  Relevant Links
- The [original PR](https://github.com/TandoorRecipes/recipes/pull/3606) was against a branch that was merged and closed. Couldn't figure out how to re-open and re-target the base, so I've just opened a new PR. 
- I like the work being done over in [another optimization PR](https://github.com/TandoorRecipes/recipes/pull/3904). I'm happy to incorporate the work, or wait until it is merged to rebase this PR. One issue I'll want to test with this solution is the user permissions (if not running as root). I was happy with `nginx` not being incorporated into the image itself (i use traefik as a reverse proxy), but if people like it incorporated, it may be more flexible to run it as a service rather than part of a `boot.sh`. [s6 overlay](https://github.com/just-containers/s6-overlay) has been great for this kind of thing

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -21,12 +21,6 @@ mediafiles
 staticfiles
 db.sqlite3
 pytest.ini
-vue/**/*.vue
-vue/**/*.ts
 **/.openapi-generator
 mkdocs.yml
-vue/babel.config*
-vue/package.json
-vue/tsconfig.json
-vue/src/utils/openapi
 venv
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,9 @@
 FROM python:3.13-alpine3.21 AS base
 WORKDIR /opt/recipes
 
+#Print all logs without buffering: https://stackoverflow.com/a/59812588
+ENV PYTHONUNBUFFERED=1
+
 #Install all dependencies.
 RUN apk add --no-cache \
   gettext \
@@ -18,9 +21,6 @@ RUN apk add --no-cache \
   zlib
 
 FROM base AS deps
-
-#Print all logs without buffering it.
-ENV PYTHONUNBUFFERED=1
 
 ENV DOCKER=true
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,25 +17,32 @@ WORKDIR /opt/recipes
 
 COPY requirements.txt ./
 
-RUN \
-  if [ `apk --print-arch` = "armv7" ]; then \
-  printf "[global]\nextra-index-url=https://www.piwheels.org/simple\n" > /etc/pip.conf ; \
+RUN <<EOF
+  if [ $(apk --print-arch) = "armv7" ]; then
+    printf "[global]\nextra-index-url=https://www.piwheels.org/simple\n" >/etc/pip.conf
   fi
 
-# remove Development dependencies from requirements.txt
-RUN sed -i '/# Development/,$d' requirements.txt
-RUN apk add --no-cache --virtual .build-deps gcc musl-dev postgresql-dev zlib-dev jpeg-dev libwebp-dev openssl-dev libffi-dev cargo openldap-dev python3-dev xmlsec-dev xmlsec build-base g++ curl && \
-  echo -n "INPUT ( libldap.so )" > /usr/lib/libldap_r.so && \
-  python -m venv venv && \
-  /opt/recipes/venv/bin/python -m pip install --upgrade pip && \
-  venv/bin/pip debug -v && \
-  venv/bin/pip install wheel==0.45.1 && \
-  venv/bin/pip install setuptools_rust==1.10.2 && \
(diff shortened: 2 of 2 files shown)


vabene1111:
thanks for the update. I will merge #3904 because it is kinda needed/a bugfix. After that I will review your PR. Please note that I made some changes to the build/container setup to allow for the integrated nginx. once I merged the other PR it would be great if you could check that everything is still working as expected. 

Also with the change to non root user: we need to be 1000% sure that this does not break anything like media file access or what not. As far as I am aware it is not needed/not heavily recommended to run the container as a non root users, its just a nice to have  and I simply don't have the capacity for any kind of headache for something that is not absolutely needed.

vabene1111:
ok so the other changes to the docker file are done. Can you update your PR?

Also what about the non root user, are we sure that this works 100%?

theProf:
> ok so the other changes to the docker file are done. Can you update your PR?
> 
> 
> 
> Also what about the non root user, are we sure that this works 100%?

Thanks, I will update when I get the chance. The nginx changes aren't tested yet and I assume they're breaking changes to what's here. A multi-service image will be more complex permission wise than the previous single service. nginx will likely want to be root so just setting a `user: 1000:1000` field would likely fail on startup.

ignas2526:
Do not add things to the Dockerfile that should be done on the host. Things like `addgroup` and `adduser` are bad. Just let everyone set their own UID and GID in the compose. Also, hardcoded `USER 1000:1000` is bad. I run every container with its own user and group since its more secure.

The point of running a container as non-root is to give it fewer privileges, not more. The container should do the least amount of things. I've been running Tandoor Recipes as a non-root, read-only, with all capabilities dropped for a couple of months now, after doing some minor adjustments to the Dockerfile and boot.sh.


Since FS is read-only, run folder cannot be modified, so in boot.sh `nginx` -> `nginx -g "pid /tmp/nginx.pid;"` and `/run/tandoor.sock` -> `tmp/tandoor.sock`.


I've created user and group on the host `groupadd --system do-recipes-app && useradd --system --gid do-recipes-app --no-create-home --shell /sbin/nologin do-recipes-app`. Next I set these on the host os `export D_RECIPES_APP_UID=$(id -u do-recipes-app); export D_RECIPES_APP_GID=$(id -g do-recipes-app)` I have a sh file in the profile.d so it gets set on the startup.

I removed `envsubst '$MEDIA_ROOT $STATIC_ROOT $TANDOOR_PORT' < /opt/recipes/http.d/Recipes.conf.template > /opt/recipes/http.d/Recipes.conf`  and replaced it with a static `/opt/recipes/http.d/Recipes.conf`. Since entire FS is now read-only and I don't see the point for specifying different mounting paths inside the container.


Here's my docker compose...
```
services:
  app:
    image: vabene1111/recipes:2.3.3
    user: ${D_RECIPES_APP_UID}:${D_RECIPES_APP_GID}
    container_name: recipes_app
    restart: unless-stopped
    env_file: .env
    volumes:
      - recipes_mediafiles:/opt/recipes/mediafiles
      - recipes_staticfiles:/opt/recipes/staticfiles
    tmpfs:
      - /tmp:mode=770,size=64m,uid=${D_RECIPES_APP_UID},gid=${D_RECIPES_APP_GID}
      - /var/lib/nginx/logs:mode=770,size=64m,uid=${D_RECIPES_APP_UID},gid=${D_RECIPES_APP_GID}
    deploy:
      resources:
          limits:
            cpus: '0.5'
            memory: 1G
    cap_drop:
      - ALL
    read_only: true
    depends_on:
      - db

volumes:
  # DO NOT run docker volume create.
  # docker container run --rm -v recipes_mediafiles:/data debian:bookworm /bin/sh -c "touch /data/.initialized && chown -R ${D_RECIPES_APP_UID}:${D_RECIPES_APP_GID} /data"
  # docker container run --rm -v recipes_staticfiles:/data debian:bookworm /bin/sh -c "touch /data/.initialized && chown -R ${D_RECIPES_APP_UID}:${D_RECIPES_APP_GID} /data"
  recipes_mediafiles:
    external: true
  recipes_staticfiles:
    external: true
```

PR #3935, opened 2025-08-13
fix(vite): Configure Vite to use subfolder

### LINKED ISSUE #3903: Tandoor V2: Webapp does not respect subfolder in URL
### Tandoor Version

2.0.1

### Setup

Docker / Docker-Compose

### Reverse Proxy

Others (please state below)

### Other

nginx

### Bug description

When first loading the webapp with a subfolder, the main page displays a 404 error.

Image

If I click on home, everything appears, but then the subfolder in the URL bar is removed.

Image

When trying to load a recipe directly without the subfolder, I get a 404 from nginx. (This is expected and correct behavior.)

Image

When trying to load a recipe directly with the subfolder, I get a 404 in the application.

Image

It seems that the vue router is not respecting the subfolder in the URL. This is evident as when I click on something in the webapp after it loads, the subfolder is removed from the URL and things work as expected. (sort of, as I can't directly load content without clicking through the webapp.)

In Tandoor v1, I can directly load content with the subfolder in the URL.

I do have X-Script-Name set correctly to the subfolder. This configuration worked flawlessly under Tandoor v1.

### Relevant logs

```shell

```

AdrienPoupa:
Fixes #3903 

In `vue3/vite.config.ts`, as recommended in the docs: https://vite.dev/guide/build.html#relative-base we use a relative path. I was able to confirm it was not introducing issues when building the app and serving it from the dev server.

Then for SPA navigation, we need to init the router with the base URL. This is done when instantiating the router, so we retrieve the base URL injected by the backend and extract the path if there's one, or defaulting to `undefined`, which is the default behaviour if there is none: https://router.vuejs.org/api/functions/createWebHistory.html#Function-createWebHistory-

### DIFF
--- a/vue3/src/apps/tandoor/main.ts
+++ b/vue3/src/apps/tandoor/main.ts
@@ -63,8 +63,12 @@ TANDOOR_PLUGINS.forEach(plugin => {
     routes = routes.concat(plugin.routes)
 })
 
+const basePath = localStorage.getItem("BASE_PATH")
+const pathname = basePath?.startsWith("http") ? new URL(basePath).pathname : undefined
+const base = pathname === "/" ? undefined : pathname
+
 const router = createRouter({
-    history: createWebHistory(),
+    history: createWebHistory(base),
     routes,
 })
 
--- a/vue3/vite.config.ts
+++ b/vue3/vite.config.ts
@@ -7,7 +7,7 @@ import {VitePWA} from "vite-plugin-pwa";
 
 // https://vitejs.dev/config/
 export default defineConfig({
-    base: '/static/vue3/',
+    base: './',
     plugins: [
         vue({
             template: {transformAssetUrls}


vabene1111:
thank you for the PR and the good description. This looks good. Will get it merged and tested with my setup.

vabene1111:
So i had to change the base in the vite config to be conditional otherwise the dev server does not work properly with django 

https://github.com/TandoorRecipes/recipes/commit/5faa74a75d5910f9a8a9deba0ba2005d851dfce0#diff-cd16465af3aa93e4e0a5971e3576cbdb6c4bc853122a6f8efee5a824e280406aR12

was that change just because you happend to stumble about the absolute path and wanted to change it to relative or is it needed for sub path setups? With the fix it works but if its not needed I can revert it and reduce some complexity maybe.

AdrienPoupa:
If I remember correctly the fix was needed, but I had only tested it with a yarn build, then publishing the static files, then starting Django.

I was not sure how to run the whole stack locally since the dev documentation seemed a bit out of date, especially for Vue 3.

I think the condition is probably needed for both cases to work.

Thanks for the review and merging!

vabene1111:
ok perfect, thanks for the feedback. The update is out now so feel free to test if everything works the way you want it to work

AdrienPoupa:
> ok perfect, thanks for the feedback. The update is out now so feel free to test if everything works the way you want it to work

It works great, thanks for the V2 the UI is much nicer.

PR #3956, opened 2025-08-21
Update devcontainer to work with new vue3 UI.

### LINKED ISSUE #3925: VSCode instructions unavailable
### Documentation link

https://docs.tandoor.dev/contribute/guidelines/#install-and-configuration

### Affected section

Other (please state below)

### Other

Contributing Code

### Issue description

Both links under "Install and Configuration" are dead.
(I've found them in the Github 1.x tag, but still can't get the VSCode devcontainer to work. Any hints?)

c0mputerguru:
Fixes #3925

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,12 +1,7 @@
-FROM python:3.10-alpine3.18
+FROM python:3.13-alpine3.22
 
 #Install all dependencies.
-RUN apk add --no-cache postgresql-libs postgresql-client gettext zlib libjpeg libwebp libxml2-dev libxslt-dev openldap git yarn
-
-# Fix libxml error from xmlsec https://github.com/xmlsec/python-xmlsec/issues/257#issuecomment-1738620862
-RUN echo "https://dl-cdn.alpinelinux.org/alpine/v3.15/community/" | tee -a /etc/apk/repositories
-RUN echo "https://dl-cdn.alpinelinux.org/alpine/v3.15/main" | tee -a /etc/apk/repositories
-RUN apk add --no-cache libxml2-dev=2.9.14-r2  xmlsec-dev=1.2.33-r0
+RUN apk add --no-cache postgresql-libs postgresql-client gettext zlib libjpeg libwebp libxml2-dev libxslt-dev openldap git yarn libgcc libstdc++ nginx tini envsubst nodejs npm
 
 #Print all logs without buffering it.
 ENV PYTHONUNBUFFERED 1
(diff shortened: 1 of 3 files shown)


c0mputerguru:
The devcontainer is functional again now that I mirrored the changes to the main Dockerfile into the devcontainer.  I had to clean up the tasks as well to remove the old vue tasks.  The 'yarn dev' task seems to run just fine, but doesn't actually show anything.  I'm not too familiar with the vue3 dev process; is that something that is supposed to work?

vabene1111:
thanks for fixing this. Not sure what yarn dev does, I only use yarn run for development and yarn build for building. I think that is the way it should be for the devcontainer as well.

c0mputerguru:
The question I had was for vue3 development, not overall development.  There are 3 scripts in the vue3 directory:

```
  "scripts": {
    "dev": "vite",
    "build": "vite build --emptyOutDir",
    "preview": "vite preview"
  },
```

I don't see anything defined as a run target, and when I run `yarn run` I only get the option to select one of those 3 options.  If I run the `dev` command which just runs `vite`, it starts a server, but when I go to the port in my browser, I get a 404.  That means its correctly connecting to the server, but not actually able to browse to anything.

The only time I used the devcontainer I was doing python development, not frontend, so I never used any of the vue tasks, but just copied from what was in https://docs.tandoor.dev/contribute/installation/.  There it says `yarn serve` is how to start the development frontend server, but that doesn't seem to work.

Do you use any of these?  If not, we should remove it from the docs since its broken, and I can remove those items from the devcontainer as well.

vabene1111:
I am sorry for my previsous comend, i looked it up and it is wrong. Serve was the old way, now it is `yarn run dev` as you thought.

The difference for the 3 modes are:
dev: hot reload server for development
build: file build for production
preview: (i think) is a hybrid of both that serves files for development but like they were build. Never really used this and never really needed this, so no need to include it for now I think.

One thing that might have caused you troubles is that the `yarn run dev` needs to happen before django is started because django checks the dev server port to see if a dev server is running. If so it connects to it if not it uses the static files. For performacne reasons this is only done once during startup and if the order is wrong there will be no static files (as yarn runs in dev mode) but django will try to get them. 

I guess the docs need to be updated , will look into that.

c0mputerguru:
Ah, ok.  Sounds like the vue dev server works in conjunction with django
server, but not on its own.  The devcontainer supports running both, so I
think it works, but I'll check it out tomorrow.

vabene1111:
thanks, once you checked let me know and I will merge this

c0mputerguru:
Looks like `DEBUG` has to be set true as an environment variable for django to look for the vite server.  I added that, cleaned up some unnecessary vscode task dependencies, and updated the documentation a bit to try to make it clearer what needs to be done.

c0mputerguru:
I was able to test this out and it indeed works.  This should be ready to merge now.

PR #3960, opened 2025-08-23
Allow video file types to be uploaded

dertasiu:
In the previous version of Tandoor I could upload some videos in the steps to document some specific process. Now this seems impossible.
Although I can still download the videos that were uploaded in the latest version.

This pull request adds two "file" types in the allowed list, .mp4 and .mov.

### DIFF
--- a/cookbook/helper/image_processing.py
+++ b/cookbook/helper/image_processing.py
@@ -37,7 +37,7 @@ def get_filetype(name):
 
 def is_file_type_allowed(filename, image_only=False):
     is_file_allowed = False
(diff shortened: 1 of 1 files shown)


vabene1111:
thank you, that makes sense

spidersouris:
It would also be great if videos were directly embedded onto the recipe pages when they are associated with a step, so that they could be played when checking instructions, e.g. with [`<video>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/video)

Issue #3943, opened 2025-08-16
Login leads to internal server error


icedieler:
### Tandoor Version

2.0.0

### Setup

Manual Setup

### Reverse Proxy

Others (please state below)

### Other

nginx

### Bug description

I've upgraded my Tandoor installation from 1.5.19 to 2.0.0. Now I can no longer login because the application crashes with an unhandled exception:

```
allauth.usersessions.models.UserSession.DoesNotExist: UserSession matching query does not exist.
```

Thus I am unable to login and see my recipes. Please see the full log below.

What I did to resolve this issue was changing the password of the user. I also ran the `migrate.py` script multiple times according to the manual installation steps.

So what am I missing?

### Relevant logs

```shell
[2025-08-16 14:35:27 +0200] [2648048] [DEBUG] POST /accounts/login/
MainThread ERROR 2025-08-16 14:35:27,625 django.request Internal Server Error: /accounts/login/
Traceback (most recent call last):
  File "/var/www/tandoor/lib/python3.13/site-packages/django/db/models/query.py", line 916, in get_or_create
    return self.get(**kwargs), False
           ~~~~~~~~^^^^^^^^^^
  File "/var/www/tandoor/lib/python3.13/site-packages/django/db/models/query.py", line 637, in get
    raise self.model.DoesNotExist(
        "%s matching query does not exist." % self.model._meta.object_name
    )
allauth.usersessions.models.UserSession.DoesNotExist: UserSession matching query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/var/www/tandoor/lib/python3.13/site-packages/django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
psycopg2.errors.NotNullViolation: null value in column "ip" of relation "usersessions_usersession" violates not-null constraint
DETAIL:  Failing row contains (4, 2025-08-16 12:35:27.623239+00, null, 2025-08-16 12:35:27.623248+00, x3ej38ks3l0mu4jq79heq12equwfbjym, Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605...., {}, 1).


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/var/www/tandoor/lib/python3.13/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
[...]

PR #3999, opened 2025-09-03
Update nginx configuration for manual setup

icedieler:
Since v2 tandoor added user session tracking which requires the reverse proxy to add an extra header.

This change adds the `X-Forwarded-For` header to the example nginx configuration. This header fixes the issue described in #3943.


vabene1111:
thanks for the PR

PR #4005, opened 2025-09-05
add descriptive css

tomtjes:
This adds human-readable css classes like they existed in Tandoor 1. These are not necessarily compatible with 1.x though. They follow this pattern:

`section__element`, e.g. `recipe__title` or `mealplan__item`
`element--variant`, e.g. `unit--grams` or `keyword--dessert`

Some `style` attributes have been moved from template/HTML tags to CSS.

Deprecates  `step__scalable-num_scaled_{up,down,false}`  in favor of  `.scaled--{up,down,default} .step__scalable-num` where the `.scaled` class is applied to the entire recipe.

The formatting changes were automatically done by the devcontainer.

### DIFF
--- a/vue3/src/components/display/RecipeView.vue
+++ b/vue3/src/components/display/RecipeView.vue
@@ -30,57 +30,57 @@
         </v-card>
       </v-card>
 
-            <!-- only display values if not all are default (e.g. for external recipes) -->
-            <v-card class="mt-1 recipe__meta" v-if="recipe.workingTime != 0 || recipe.waitingTime != 0 || recipe.servings != 1">
-                <v-container>
-                    <v-row class="text-center text-body-2">
-                        <v-col class="pt-1 pb-1 recipe__worktime" :class="'worktime--' + String(recipe.workingTime)">
-                            <i class="fas fa-cogs fa-fw mr-1"></i> {{ recipe.workingTime }} min<br/>
-                            <div class="text-grey">{{ $t('WorkingTime') }}</div>
-                        </v-col>
-                        <v-col class="pt-1 pb-1 recipe__waittime" :class="'waittime--' + String(recipe.waitingTime)">
-                            <div><i class="fas fa-hourglass-half fa-fw mr-1"></i> {{ recipe.waitingTime }} min</div>
-                            <div class="text-grey">{{ $t('WaitingTime') }}</div>
-                        </v-col>
-                        <v-col class="pt-1 pb-1 recipe__servings" :class="'servings--' + String(recipe.servings)">
+      <!-- only display values if not all are default (e.g. for external recipes) -->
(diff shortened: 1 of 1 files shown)


vabene1111:
hi @tomtjes thanks for the PR and the continued support of this project. Sadly i will not merge this. 

First of all I hate the formatting. I know its in the docs and the devcontainer but it just doen't work for me. Smilerz tried to make a standard for this project but everytime I try to work with it I think about abandonning this project alltogether and I simply do not want to waste another second on this. 

Second this is a very nieche feature/requirement and I do not want to maintain it. I also do not want to leave it broken all the time as I did in tandoor 1 when I just ignored it when refactoring stuff. Also I do not want the extra complexity of computed properties and special functions for CSS.

I am sorry you did all this work and I am sorry that this is not going to be in tandoor but this is the way it is. Next time i recommend talking to me before doing that, I could have saved you the work.

tomtjes:
Hi @vabene1111, I appreciate the detailed response.

I completely get the formatting issue. Happy to uninstall those plugins and undo all of that.

You're right, I should have discussed this beforehand, got a little over-excited to finally implement it, after a partial implementation in v1 was cut short when development on v2 started. However, my contribution was well received then, which is why I saw this as re-implementing and completing a feature from v1. But I get it, that was a long time ago and preferences can change.

I agree that it's rare that someone would use custom css, but I feel like without these classes, custom css will be even less used as it's pretty hard to target specific elements. The current page source makes it difficult to figure out, whether you're on a recipe page or a meal plan page, for instance.

I'd like to suggest to reduce the complexity of this PR and only add the non-computed classes that are at least making the building blocks of the site easier to target for custom css.

_(One of my favorite use cases involved a computed class though, just FYI. Since property calculation requires a unit, I had to use `1 piece egg` or `1 each egg` where you'd normally say `1 egg`. So I hid the unit `piece` or `each` using css.)_

vabene1111:
thanks for the response. Again I really really hate closing PR's like this because I know someone spend time and was motivated to do something and it would help at least them having it. 

I can agree to non computed for now only. I cannot guarantee that I will take proper care when updating to retain the classes but will try my best. 

If you could do the PR without formatting that would be a blessing, would make merging 1000 times easier (and yes I know it would be even better if I would just accept the formatters 🙈 )

regarding your other comment: That kinda sounds like we need a "dont display" flag for units, right? I think we can at least discuss that, not sure if it should be implemented or not but a discussion might not hurt.

tomtjes:
Thank you, and no worries: I also had a lot of fun exploring how everything works and what is and isn't possible to do.

I'll work on a minimal version of the PR. If there are future bugs related to those CSS classes, feel free to assign them to me.

I created a feature request for the unit issue.

vabene1111:
thank you for your understanding i know you have been around a long time and am really sorry for not merging this. Looking forward to the simplified PR.

PR #4006, opened 2025-09-05
fixed localization function call in Timer.vue

mkieledunsche:


vabene1111:
sadly this fix is not correct, but I fixed the code, thanks for the attemt.

mkieledunsche:
@vabene1111 Thans for your patience and fixing the code!

PR #4013, opened 2025-09-07
Make logo consistent across end user systems

tomtjes:
On systems that don't have access to the logo font (e.g. Safari), the logo used a default serif font. I've converted the text to an svg path, so it should display the same on any system.


vabene1111:
perfect, thank you 👍

PR #4039, opened 2025-09-15
Optimize Dockerfile and build process

wilmardo:
Somewhat of a simpler implementation of #3919

@theProf I cherry-picked some of your work to not discard your work :)

* Moves the yarn build into the Dockerfile so that a simple `docker build` ends up as a working image instead of missing the compiled frontend
* Moves the version.py executing into the CI so that the `.git` directory can be omitted from the container
* Reduce the amount of files in the final image by just the bare minimum


vabene1111:
thanks for this and the other PR's
I do have some feature updates I want to finish before playing with the setup again and I need to think about some preferences regarding this. Will get back to this but might be a bit, sorry.

wilmardo:
No worries, let me know if there are any questions

also @theProf feel free to shoot some comments. I decided on less stages to keep the Dockerfile more simple and readable

vabene1111:
So I have (finally) merged the non root PR. Now there are two open PRs regarding optimizing the Dockerfile with mainly @wilmardo and @theProf working on them (this one and #3919)

One thing that I think was somewhat unclear to you when you first did them was that the frontend build needs to work at runtime because otherwise the plugin system does not work (also something I think might not work with the non root pr @wilmardo is that possible? - i can live with plugins only working with root for now).

I personally have two things I would like to improve 
- allowing to override the nginx config again (probably through templates). My current solution sucks/is super hacky but I had no time and it worked
- maybe finding a way to not always rebuild the frontend if the plugins haven't changed. Ideally I would be able to start the container with a special flag to trigger the build or maybe I add something to do the build trough the UI?

Does anyone of you or you both (or anyone else in this thread) feel like creating one PR that respects these things and optimizes the container? If yes feel free to give it a go and also feel free to contact me for any questions, I would be very greateful.

wilmardo:
yes sure! I will try to find some time upcoming weekend to incorporate your feedback. That was exactly what I was looking for. Let me come back with an updated PR :)

PR #4040, opened 2025-09-15
feat: make nginx respect ipv6 disable fixes #3996

### LINKED ISSUE #3996: nginx won't start if ipv6 is disabled
### Tandoor Version

2.0.3

### Setup

Docker / Docker-Compose

### Reverse Proxy

No reverse proxy

### Other

_No response_

### Bug description

if ipv6 is disabled nginx won't start

try 
```
nginx -t

2025/09/01 09:27:53 [emerg] 82#82: socket() [::]:80 failed (97: Address family not supported by protocol)
nginx: [emerg] socket() [::]:80 failed (97: Address family not supported by protocol)
```

comment out ipv6 listener in  /etc/nginx/http.d/Recipes.conf and Recipes.conf.template

```
server {
  listen 80;
#  listen [::]:80 ipv6only=on;
...
...
...
}

```

and reboot container without recreating it 

`reboot`

nginx started successfully

and all images from media are displayed correctly

### Relevant logs

```shell
n/a
```

wilmardo:
feat: make nginx respect ipv6 disable fixes #3996

### DIFF
--- a/boot.sh
+++ b/boot.sh
@@ -103,6 +103,9 @@ ipv6_disable=$(cat /sys/module/ipv6/parameters/disable)
 
 # prepare nginx config
 envsubst '$MEDIA_ROOT $STATIC_ROOT $TANDOOR_PORT' < /opt/recipes/http.d/Recipes.conf.template > /opt/recipes/http.d/Recipes.conf
+if [ "$ipv6_disable" -eq 1 ]; then
+    sed -i '/listen \[::\]:80/d' /opt/recipes/http.d/Recipes.conf
+fi
 
 # start nginx
 echo "Starting nginx"


vabene1111:
thanks for the PR but if a change to the nginx config fixes your issue you can just mount the config to your host system and change it there, no need to have an extra piece of code running for everyone on every boot :) 

or am I missing something?

wilmardo:
Not my issue but @cpritschow but since I was already touching the codebase I thought I will get a quick fix in.

>thanks for the PR but if a change to the nginx config fixes your issue you can just mount the config to your host system and change it there, no need to have an extra piece of code running for everyone on every boot :)
or am I missing something?

No that is pretty valid, the only thing is that the gunicorn startup command is ipv6 disable aware (https://github.com/TandoorRecipes/recipes/blob/develop/boot.sh#L113) and nginx currently isn't. This change would have unified that behaviour.

On most setups the `sed` will be skipped to the impact is negligible.

PR #4041, opened 2025-09-16
refactor: move staticfiles collection from runtime to build stage

wilmardo:
This will remove the `STATIC_ROOT` env var, I will update the documentation accordingly after this is merged.

I personally don't see any problems doing this, it is now guaranteed that each user runs exactly the same setup since it build into the container.

What was the reason that this collection existed on runtime instead of buildtime?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,9 @@ COPY . ./
 RUN rm -rf /etc/nginx/http.d && \
     ln -s /opt/recipes/http.d /etc/nginx/http.d && \
     ln -sf /dev/stdout /var/log/nginx/access.log && \
-    ln -sf /dev/stderr /var/log/nginx/error.log
+    ln -sf /dev/stderr /var/log/nginx/error.log && \
+    # collect staticfiles
+    venv/bin/python manage.py collectstatic --noinput
 
 # commented for now https://github.com/TandoorRecipes/recipes/issues/3478
 #HEALTHCHECK --interval=30s \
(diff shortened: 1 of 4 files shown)


vabene1111:
It used to be that the yarn build is run outside the dockerfile which you already fixed so that one is taken care of. 

One relatively new feature tough requires this: The plugin system. The plugin system builds the frontend when booting and needs to collect the static files afterwards. It is a lot better than the Tandoor v1 Plugin system that required building an entire container for every plugin combination you wanted, yet I am not very happy with it building on every restart even when nothing was changed. 

The 100% best option would be a plugin installer in the UI but I am not sure if that is too much. It would also require leaving node and git available in the runtime. 

If you have any ideas let me know.

wilmardo:
Ah that was a feature I wasn't familiar with, makes sense though why the collect static was at runtime haha. Do you have a quick example how you would setup a plugin?

vabene1111:
sorry for the late reply. you must mount the plugins folder `./plugins:/opt/recipes/recipes/plugins` and clone the plugin (e.g. https://github.com/TandoorRecipes/open_data_plugin) into it. then set `PLUGINS_BUILD=1` in your env for the container to rebuild the frontend on every boot.

edysli:
> I personally don't see any problems doing this, it is now guaranteed that each user runs exactly the same setup since it build into the container.

I think it makes sense to integrate those static files into the container image, because they are part of the application's code (they only change when the application code changes). Mastodon does it this way, for example, and it saves quite a bit of time not having to do this when the container starts.

As for plugins, could they ship their own static files so the entire set doesn't have to be rebuilt but only added to?

vabene1111:
> As for plugins, could they ship their own static files so the entire set doesn't have to be rebuilt but only added to?

no, sadly this is not the case. I wanted plugins to be able to integrate directly with the pages in the main application. This means they must be present at build time. What would be great would be a means to detect if there are any changes to the plugins and only rebuild then or if you could pass a parameter like --build to the container startup to only rebuild then but I am not sure if that is feasible. 

The problem with that I think is the external build dir and  the cleanup flag for the vite build.

PR #4042, opened 2025-09-16
feat: make it possible to switch to a non-root user

wilmardo:
These changes make it possible to start the container as a non-root user. The only prerequisite is an external database since the non-root user can't create the `db.sqlite3` in the projectroot anymore (this can be fixed by changing the path, if needed I can expose this as en env var).

None of these changes should impact existing users (or new for that matter) since the image itself doesn't change user and it rootful still. We can't really move away from that easily since the users will need to chown all the mediafiles to be owned by the user.

However these changes will give users the option to switch to a non-root user :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,12 +30,16 @@ RUN apk add --no-cache --virtual .build-deps gcc musl-dev postgresql-dev zlib-de
 #Copy project and execute it.
 COPY . ./
 
-# delete default nginx config and link it to tandoors config
-# create symlinks to access and error log to show them on stdout
-RUN rm -rf /etc/nginx/http.d && \
-    ln -s /opt/recipes/http.d /etc/nginx/http.d && \
-    ln -sf /dev/stdout /var/log/nginx/access.log && \
+RUN <<EOF
+    # delete default nginx config and link it to tandoors config
+    rm -rf /etc/nginx/http.d
+    ln -s /opt/recipes/http.d /etc/nginx/http.d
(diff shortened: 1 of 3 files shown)


vabene1111:
I like the approach, it will finally get people of my chest complaining about missing non root functionality without annoying all the people that do not care. As with the other, will get back to this, thanks!

ghost:
Hi dear,
I assume this MR doesnt allow us to set uid/gid for the process in the container, right?

wilmardo:
Yes when this is in the image you should be able to set any uid/gid you want. It does require that you use an external database and not sqlite.

ghost:
@wilmardo Thanks for explaining.

Meanwhile I found out how to use with sqlite in docker-compose:

```
...
    user: 1006:1006
...
    volumes:
      - /home/tandoor/dockervolumes/staticfiles:/opt/recipes/staticfiles
      - /home/tandoor/dockervolumes/mediafiles:/opt/recipes/mediafiles
      - /home/tandoor/dockervolumes/data:/opt/recipes/data
      - /home/tandoor/dockervolumes/sec.txt:/opt/recipes/sec.txt:ro
...
    environment:
      - SECRET_KEY_FILE=/opt/recipes/sec.txt
      - DB_ENGINE=django.db.backends.sqlite3
      - POSTGRES_DB=/opt/recipes/data/db.sqlite3
```

The key fact is to not directly bind an (empty) sqlite3 file, but to bind a whole dir, where the user 1006 (in my case) has read/write permissions.

PR #4068, opened 2025-09-24
Clean up commit history with interactive rebase

Dalnu:
This PR demonstrates squashing messy commits into a clean, semantic commit.

Before:
- Multiple noisy commits: "fix", "oops", "more things", "added stuff", etc.

After:
- 1 atomic commit with a clear message:
  "Consolidate notes and docs updates"

Why:
- Easier to understand history
- Professional commit log
- Groups related changes together

### DIFF
--- a/DEV_NOTES.txt
+++ b/DEV_NOTES.txt
@@ -1,0 +1,2 @@
+"Tip: Always preheat oven." 
+"Use metric units" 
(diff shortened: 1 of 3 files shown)


vabene1111:
thank you, I do know how git works but I personally prefer to let my way of learning/testing be documented both for me and maybe even others to refer back. One could occasionally put more thought into the commit messages though, I agree with that :)

Issue #4084, opened 2025-09-29
Shopping list does not show all items


nickinzon:
### Tandoor Version

2.2.7

### Setup

Docker / Docker-Compose

### Reverse Proxy

Nginx Proxy Manager (NPM)

### Bug description

When I add a meal to my shopping list, it only adds a fraction of the necessary ingredients to the list. The number of added ingredients varies. For the same recipe once there were 5 items added, once 6 items. But never all the items necessary. It does not matter, if the meal is added to the shopping list via the meal planner or within the shopping list itself.

### Relevant logs

```shell

```

vabene1111:
I have been using this feature in Tandoor 2 for a long time and never had such an issue. Can you open your browser debug tools and see if there are any issues shown there?

Can you set `DEBUG=1` on your server and see if there are any error messages in the console when adding something to shopping?

c0mputerguru:
I'm on a slightly older version and experiencing a similar issue.  If I add a recipe to my meal plan, it shows some of the items, but some of the other items don't show.  If I go to the recipe tab on the shopping list, look at the recipe (showing the ingredients), and then go back to the shopping list tab, those missing ingredients then show up.  If I browser refresh and go to the shopping list, the item goes missing again.  So I think it probably has something to do with how the items for the shopping list are being loaded.

I'll look to debug this a bit later on my instance, but I'm also running an older build (2.0.1).

c0mputerguru:
Couple more pieces of info:

- I've upgraded to 2.2.7 and I still have the same issue.
- If I look at the API requests when I load the shopping cart, it seems that the missing item is not returned as a part of the shopping-list-entry API call.  That call is paginated, and I went through the json responses for all of the pages and that particular entry is not there.
- If I reduce the shopping_recent_days user preference (mine is defaulted to 14) to 2, then things seem to load fine.

So I suspect there's a bug in the implementation of the API call that returns all the shopping list entries where it drops entries in some cases.  I reviewed the code, and I can't seem to find any reason items would be dropped.

c0mputerguru:
Oooh, found this in the logs:

```
/opt/recipes/venv/lib/python3.13/site-packages/rest_framework/pagination.py:207: UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: <class 'cookbook.models.ShoppingListEntry'> QuerySet.ppaginator = self.django_paginator_class(queryset, page_size)
```

[...]

PR #4089, opened 2025-09-29
Provide stable ordering of shopping list entries for pagination

c0mputerguru:
Pagination without a stable ordering can result in missed entries if the order loaded for each page request is different.  I believe this was causing missed entries in the shopping list. 

See #4084 for more information.


vabene1111:
thanks for the PR, I decided to finally look into the warnings I always thought did not really effect tandoor. I am still not sure why the database would suddenly decide to change the query response ordering in between requests but I added default ordering trough the model meta options to all models so this issue should not occur again.  b1ef35e41564c2b6a13bd4201fefd1343ff71c53

Still thanks for implementing the fix for the shopping endpoint.

PR #4091, opened 2025-09-30
Requests using date should be in local timezone, not UTC as tandoor is timezone aware

c0mputerguru:
Tests were passing in a date that was always in UTC, and then that date was being interpreted on the server as in the local timezone.  This was causing sporadic test failures; when the local timezone was a different day than UTC, the tests would fail because "today in UTC" was "yesterday in local", so finding meal plans that were "today in UTC" would not include the meal plan that was "today in local".

I verified this by locally setting the timezone to something very different (Sydney Australia).  Without using timezone.localtime(timezone.now()) the test would fail, but using it the tests now pass.

This test bug was found when proposing a fix for #4084

### DIFF
--- a/cookbook/tests/api/test_api_meal_plan.py
+++ b/cookbook/tests/api/test_api_meal_plan.py
@@ -99,19 +99,19 @@ def test_list_filter(obj_1, u1_s1):
 
     response = json.loads(
         u1_s1.get(
-            f'{reverse(LIST_URL)}?from_date={(timezone.now() + timedelta(days=2)).strftime("%Y-%m-%d")}'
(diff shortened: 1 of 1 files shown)


c0mputerguru:
Really I think what would be better is not having tests depend on "now" but have fixed times, but that is a much more extensive change than I'm willing to spend time on right now.  That would make tests much more deterministic, but would also potentially mask timezone translation issues.

vabene1111:
thank you, I think this is already a good improvement. The tests need a lot of work which I currently cannot provide so we have to live with sub optimal ones, but they already help so I think thats probably 80/20 :)

Issue #4117, opened 2025-10-05
Tandoor Sudenly Died


Akusho:
### Tandoor Version

latest

### Setup

Docker / Docker-Compose

### Reverse Proxy

Nginx Proxy Manager (NPM)

### Bug description

Hello,

During automatic update by Watchtower, Tandoor went offline. Believing it might be some database corruption, I restored to several earlier snapshots, but the issue persists. 

Portainer states that all containers are running and healthy.

Trying to access the front-end locally (not through NGINX PM) displays: 

"Tandoor Recipes is not yet available 502
Services are still trying to start.
Please allow up to 3 minutes after you started the application on your server.
If this status persists, check the application or docker logs for further information.
After checking and trying everything mentioned in the docs, you can request help on the project's GitHub page.
"
Attached logs.

Any idea what might be the issue? 

### Relevant logs

```shell
Tandoor-NGINX-Recipes log: https://hst.sh/viqadikosi.yaml

Tandoor-Recipes-DB log: https://hst.sh/yuzezevoju.yaml

Tandoor-Recipes-Web log: https://hst.sh/aqavomicos.properties
```

Xyaren:
Try naming your container `web_recipes`.

The contaner intenral nginx is using that as a hostname for the backend.
https://github.com/TandoorRecipes/recipes/blob/develop/nginx/conf.d/Recipes.conf#L22

It's not really greate that the container uses it's name for resoltion.
I'll make a PR to fix that.

Akusho:
Hello,
Thank you for the suggestion. Doesn't seem to have helped. Still the same 502 error:
web_recipes log: https://hst.sh/abajowikej.properties
NGINX log: https://hst.sh/orogedolah.yaml

skullone:
I have the same issue.  All containers are up and running but I get the same errors as Akusho.

My setup is as follows:

docker container for Tandoor.  My opnsense router uses ACME and Caddy for Let's Encrypt certs and the reverse proxy.  This has not been an issue until this morning after version 2.3 was pushed.

inverimus:
I tried changing the hostname to web_recipes and that didn't fix anything.

Akusho:
See this:
https://github.com/TandoorRecipes/recipes/issues/4120
I missed Tandoor changing the container structure. 

Now I use this compose:
```
services:
  db_recipes:
    container_name: Tandoor-Recipes-DB
    restart: unless-stopped
    image: postgres:16-alpine
    volumes:
[...]

PR #4118, opened 2025-10-05
Use loopback in nginx to point to own backend instead of container name

Xyaren:
Fixes tandoor not being available when container is not named exatcly `web_recipes`.

Fixes https://github.com/TandoorRecipes/recipes/issues/4117


vabene1111:
thanks for the PR but this is obsolete. This config is no longer used and was only present for backward compatability reasons. If you setup according to the tandoor 2 setup instrctions the config in http.d will be used which uses unix sockets and thus is completely Independent of any container naming choices

PR #4154, opened 2025-10-12
fix(vue): correct pagination to show page numbers instead of recipe numbers in BookViewPage

racehd:
The v-pagination component had both @update:model-value and @next/@prev handlers, causing pagination to increment twice on desktop (mdAndUp=true). The @next/@prev handlers were redundant since @update:model-value already handles all pagination changes.

Removed @next/@prev to fix issue. Verified changes by building local docker image and testing both mobile and desktop viewports.

### DIFF
--- a/vue3/src/pages/BookViewPage.vue
+++ b/vue3/src/pages/BookViewPage.vue
@@ -33,9 +33,9 @@
 
         <v-row>
             <v-col class="text-center">
-                <v-pagination :model-value="page + 1"
-                              @update:model-value="value => page = Math.max(value - 1, 0)"
-                              :length="totalItems"
+                <v-pagination :model-value="currentPageNumber"
+                              @update:model-value="value => page = (value - 1) * recipesPerPage"
+                              :length="totalPages"
                 ></v-pagination>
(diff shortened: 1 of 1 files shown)


racehd:
On closer look, it appears the BookViewPage list showed the number of recipes, not number of pages as I had initially thought. So the old behavior was functional, but confusing since on desktop this made it looked like you were moving 2 "pages" at a time but you were actually just moving 2 recipes. My initial PR would've been a regression since on desktop you were then only navigating one recipe at at time despite seeing 2 recipes at time. 

To resolve the underlying issue, I made a 2nd commit to list the number of pages instead of the number of recipes. Now on desktop and mobile it will correctly show the number of pages, and navigate 1 page at a time via 1 handler instead of duplicated handlers as before. 

Verified my changes by building local docker image and testing both mobile and desktop viewports. 

visual of new PR BookViewPage on desktop: 
2025-10-24-160344_hyprshot
and using arrow / keyboard once ->  
2025-10-24-160408_hyprshot
(Correctly moved one page, with 2 new recipes)

PR #4189, opened 2025-10-26
feat: show recipe author in the recipe card

djstini:
I've added a pill to the recipe card that shows the recipes author.


vabene1111:
hi thanks for the PR but I will have to close this. Its not that this was not added because I wasn't able to do this myself but because I decided so. The cards are overloaded quickly and thus I limit what is displayed.

In a future release I might add some settings to let the user decide what to display but for now I dont want this changed.

djstini:
If i took the time to make a space-option for showing it or not, would you merge it?

vabene1111:
no sorry, I want this to be handled as a whole feature block with a proper editor and settings ui to toggle the different options on and off, its no help for me to just get this done.

PR #4198, opened 2025-10-31
feat(ui): add footer with contribute link and hover animation

nihalsingh373:
Added a new footer to tandoor.html with a 'Contribute' link and hover animation for better UI polish and visibility of open-source info.


vabene1111:
thank you but this is not wanted. There is already a github link trough the help menu and this isn't even in the frontend

PR #4216, opened 2025-11-09
Adding port to the docker_compose.yml

vspaziani:
Updating to include port in docker_compose.yml. Without this I do not think the docker container has a mechanism for displaying the webpage on the internal port.


Daan5556:
Are you sure you made the correct change? I think you meant to add the port mapping to the web container instead of the database.

vspaziani:
> Are you sure you made the correct change? I think you meant to add the port mapping to the web container instead of the database.

Actually, I took a second look. You are correct, copy and paste error. Sorry, I don't know how to edit the pull request... I'll close this and resubmit with the correct location. Thanks for catching that.

Issue #4177, opened 2025-10-22
Django_vite Core Exception after Super user Sign-in


doylelew:
### Tandoor Version

TandoorRecipes/recipes:develop

### Setup

Docker / Docker-Compose

### Reverse Proxy

No reverse proxy

### Bug description

After forking from Develop and building a docker image you can create a super user at port 80, then once you try to login as that super user you get a server error 500 from not being able to locate the favicon.

### Relevant logs

```shell
ThreadPoolExecutor-0_0 ERROR 2025-10-22 19:33:08,001 django.request Internal Server Error: /favicon.ico
2025-10-22T17:33:08.006986912Z Traceback (most recent call last):
2025-10-22T17:33:08.006991402Z   File "/opt/recipes/venv/lib/python3.13/site-packages/django/core/handlers/exception.py", line 55, in inner
2025-10-22T17:33:08.006995139Z     response = get_response(request)
2025-10-22T17:33:08.006997818Z   File "/opt/recipes/venv/lib/python3.13/site-packages/django/core/handlers/base.py", line 197, in _get_response
2025-10-22T17:33:08.007000958Z     response = wrapped_callback(request, *callback_args, **callback_kwargs)
2025-10-22T17:33:08.007003583Z   File "/opt/recipes/cookbook/views/views.py", line 49, in index
2025-10-22T17:33:08.007006227Z     return render(request, 'frontend/tandoor.html', {})
2025-10-22T17:33:08.007009134Z   File "/opt/recipes/venv/lib/python3.13/site-packages/django/shortcuts.py", line 25, in render
2025-10-22T17:33:08.007012187Z     content = loader.render_to_string(template_name, context, request, using=using)
2025-10-22T17:33:08.007014707Z   File "/opt/recipes/venv/lib/python3.13/site-packages/django/template/loader.py", line 62, in render_to_string
2025-10-22T17:33:08.007017586Z     return template.render(context, request)
2025-10-22T17:33:08.007020105Z            ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
2025-10-22T17:33:08.007022527Z   File "/opt/recipes/venv/lib/python3.13/site-packages/django/template/backends/django.py", line 107, in render
2025-10-22T17:33:08.007025288Z     return self.template.render(context)
2025-10-22T17:33:08.007027672Z            ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
[...]

PR #4222, opened 2025-11-11
Added Yarn install and Build commands to DockerFile

doylelew:
Based on the info from this ["Django_vite Core Exception after Super user Sign-in " issue](https://github.com/TandoorRecipes/recipes/issues/4177). I have included the necessary npm install of yarn, then the yarn install and build commands to the docker file before the entry point. This solves the broken state of building from source without the github actions. I am not very familiar with github actions so let me know if it breaks your "Build Docker Container" action.


vabene1111:
thank you but I do not want this to be part of the docker file. As far as I know it will increase the size due to the inclusion of the libraries. I also dont want to change anything in the build process as it works quite nicely as it is. Feel free to add instructions to the documentation if they are not there yet on how to manually build the container.

PR #4225, opened 2025-11-11
Update docker-compose.yml

vspaziani:
Updating Web Recipes container with external port. Without this I do not think the docker container has a mechanism for displaying the webpage on the internal port.


BenWiederhake:
Thanks! I ran into this bug this morning, and just wanted to make this exact PR. Funny timing :D

Issue #4220, opened 2025-11-10
Allow sharing pictures & PDF's to PWA


basert:
### Is your feature request related to a problem? Please describe.

The PWA already supports link shares via `share_target`: https://github.com/TandoorRecipes/recipes/blob/d09e629415b863f690166fab25a9b7cbe7816d41/cookbook/views/views.py#L435

The `share_target` API also allows the application to receive files: https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/share_target#receiving_shared_files

The current solution should be extended to accept pictures & PDF documents. These should be forwarded to the AI import.

### Describe the solution you'd like

Extend the `share_target` configuration to allow imports of files. This probably requires better handling in the backend, ie. it should directly send the picture to the AI and get the pre-filled view.

### Additional context

I would be happy to start working on a PR for this solution if we agree this is a good addition!

vabene1111:
that is a good idea, I like it. Anyone feel free to take a hit at it.

basert:
> that is a good idea, I like it. Anyone feel free to take a hit at it.

I gave it a shot, please take a look :)

PR #4227, opened 2025-11-12
feat(share): add option to share files to PWA

basert:
* adjusts the `manifest.json` file to configure `files` option
* add service-worker handling to intercept the POST request, write the file to shared cache and redirect to the vue app
* adds handling in vue app to read the file from cache and preload the AI import view

This implements https://github.com/TandoorRecipes/recipes/issues/4220

### DIFF
--- a/cookbook/views/views.py
+++ b/cookbook/views/views.py
@@ -434,12 +434,18 @@ def web_manifest(request):
         }],
         "share_target": {
             "action": "/recipe/import",
-            "method": "GET",
-            "enctype": "application/x-www-form-urlencoded",
+            "method": "POST",
+            "enctype": "multipart/form-data",
             "params": {
                 "title": "title",
                 "url": "url",
-                "text": "text"
+                "text": "text",
+                "files": [
+                    {
+                        "name": "data",
+                        "accept": ["image/*", "text/plain", "application/pdf"]
+                    }
+                ]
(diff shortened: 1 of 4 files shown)


vabene1111:
thank you for your solution and effort. This is more complex than I had hoped it to be but looks like it works (i guess you probably  tested it).

Do you have any ideas on how this could be done "easier" (like just populating the file upload with the proper path, not sure if that works). I will leave this open for a few days, maybe someone has an idea or maybe I have one, if not I will get it merged, I do like the feature.

basert:
Yeah, unfortunately it's not as simple as opening a webpage with the file pre-bootstrapped. And I can see why, the file might be owned by the app itself and possibly doesn't even exist on the phones filesystem. So the app shares the file with the browser by `POST`ing it to the PWA.

I indeed tested it, but it's quite a fight to get the local copy running on my phone :D I had to expose it via a public hostname with valid HTTPS certificate.

To just test the internal handling (e.g. service-worker, upload page, ...), you can use this test html:
```
<!DOCTYPE html>
<html lang="en">
<body>
<form action="http://localhost:8000/recipe/import" method="post" enctype="multipart/form-data">
    <input id="data" type="file" name="data">
    <button type="submit">Upload</button>
</form>
</body>
```

Issue #3787, opened 2025-06-30
Keep servings if recipe is opened from the Meal Planner


FuzzyWereBeast:
### Is your feature request related to a problem? Please describe.

Right now, the meal planner shows you the recipe and how many servings you have planned.  when you click on the recipe, the servings are reset to what was entered when the recipe was created.

### Describe the solution you'd like

It would be nice to have the recipe automatically show the amount of servings entered in the meal plan.

vabene1111:
- [ ] add GET parameter to recipeview
- [ ] adjust link from meal plan cards (home screen and meal planner)

PR #4237, opened 2025-11-18
open meal plan recipe with right amount of servings

Orycterope:
Fix https://github.com/TandoorRecipes/recipes/issues/3787

Adds a '?servings=42' query param to RecipeViewPage, and propagates it to child views via a prop. The ingredientFactor is computed based on this param if it is present, and defaults to recipe servings otherwise.

This query param is set when comming from:

* Home page's HorizontalRecipeWindow
* MealPlanEditor

This commit also makes the RecipeView's Activity form reactive on the number of servings, before creating a comment.


vabene1111:
thanks for the PR, i will merge it. If you have some time you could use the vueUse query param bindings to make the RecipeViewPage query binding without so much code but other than that everything looks good enough so i can merge it :)

PR #4255, opened 2025-11-26
Correct misspellings of `through`

gordonjb:
In a number of places in the codebase, the word `through` is misspelt as `trough`


vabene1111:
thank you

Issue #4186, opened 2025-10-25
feat: Sort the recipe property ingredient window descending about amount


dsm1212:
### Is your feature request related to a problem? Please describe.

So, it's minor but when I want to see how ingredients contributed to a property I'm always looking for the primary contributers and I have to scan the list of values because the list is the same order as the ingredients in the recipe. 

### Describe the solution you'd like

The main reason to pop that window is to see where the total is coming from so I'd prefer if the list was sorted descending by the number/amount. For example, looking at calories, the first entry shown should be the ingredient with the most calories. 

### Describe alternatives you've considered

Just scan the list, it's not a huge deal, but since the main reason to do that click is to see what contributes to it so sorting will make this easier.

vabene1111:
i like the idea.

deeptikam:
Hello! Me and a few other group members are planning to try and solve this issue.

PR #4294, opened 2025-12-08
Sort Recipe Property Ingredient Window

deeptikam:
Issue Solved : #4186 
### New Feature
I implemented a 'Sort Decreasing' Button for the property ingredient window on a recipe so that users can sort and see which ingredients are contributing the most in a recipe. It doesn't showcase in property windows that have a warning in them, since not all values are there for sorting. Where the button is showcased, selecting it changes the color to signal that it is selected to the user, and the ingredients are sorted by decreasing percentage. Deselecting the button sets the ingredients back to the original order.

### Reasoning
Helps users identify primary contributors to a recipe. It can be helpful for figuring out which ingredients to substitute or take out to reach goals such as calorie goals.

### Implementation 
**Functions**

New constants 'originalFoodValues' and 'isSorted' are established. 'originalFoodValues' stores the original ingredient order so that it can be called when deselecting the sorting button. 'isSorted' is used for styling purposes. 

Function sortFoodValues creates a new list using the original food values and sorts it in decreasing order based on the percentage amount. It also updates 'isSorted' as needed.

**Styling**

A button is added to the property window card header with a thin outline and alignment with the text above it. Clicking the button styles it using the primary vue3 styling given. The button is not showcased when the property window has a warning.

### Testing
Manual Testing was done to ensure that the new functionality worked and was showcased only when appropriate. 

### Images 

Property Window open, showcasing an unselected button and ingredients that are not sorted. 
Screenshot 2025-12-05 at 5 23 34 PM

Property Window open, showcasing a selected button and ingredients sorted in decreasing order by percentage. 
Screenshot 2025-12-05 at 5 23 41 PM

Button not shown in Property Windows where there are warnings
Screenshot 2025-12-05 at 5 23 53 PM

### DIFF
--- a/vue3/src/components/display/PropertyView.vue
+++ b/vue3/src/components/display/PropertyView.vue
@@ -39,8 +39,11 @@
         <v-card v-if="dialogProperty" :loading="loading">
             <v-closable-card-title :title="`${dialogProperty.propertyAmountTotal} ${dialogProperty.unit} ${dialogProperty.name}`" :sub-title="$t('total')" icon="$properties"
                                    v-model="dialog"></v-closable-card-title>
-            <v-btn @click="sortFoodValues" :variant="isSorted ? 'flat' : 'outlined'" color="primary" size="small" class="rounded-pill d-print-none px-3 align-self-start ms-2 text-body-2 text-medium-emphasis">
-                Sort Decreasing</v-btn>
+            <v-btn 
+                v-if="dialogProperty && !dialogProperty.missingValue"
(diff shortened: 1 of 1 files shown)


vabene1111:
hi,. thanks for the PR. Do you think it makes sense to have this as a button? I think we could just always sort the list when its opened as I do not really see a reason to view it without being sorted 🤔

tecosaur:
Bump @deeptikam :slightly_smiling_face:

Issue #4280, opened 2025-12-03
Search by Rating (<=) results incorrect


OllieUK:
### Tandoor Version

2.3.6

### Setup

Docker / Docker-Compose

### Reverse Proxy

Traefik

### Bug description

When searching for recipes using '*Rating (<=)*' the results incorrectly show results of '*Rating (>=)*'. A (partial) list of recipes with ratings with fewer than *x* stars is currently only possible through multiple searches of '*Rating (exact)*'.

### Relevant logs

```shell

```

newhacker1746:
I attempted to fix this in #4303

PR #4303, opened 2025-12-11
fix: search by rating filter: <=

newperson1746:
**Summary**
Fix the bug in the recipe search functionality where the <= "lte" rating filter was incorrectly applying >= "gte". 

**Detailed description**
As seen in https://github.com/TandoorRecipes/recipes/issues/4280, searching for recipes with a rating of 3 stars or less would return results for those of 3 stars or more instead. This is incorrect. 

The rating filter is defined in `cookbook/helper/recipe_search.py` and the query being run for the "lte" filter was actually the same as the "gte" filter. This fixes the issue. Also re-enable a commented test case that was failing but now passes with this fix, expanding the comment explaining how the test works. This test covers all of the rating filters.

**File changes:**
`cookbook/helper/recipe_search.py`
  - `rating_filter` method: correct `self._rating_lte` case
 
`cookbook/tests/other/test_recipe_full_text_search.py` 
  - `test_search_count`: uncomment the test case for the "lte=3" filter and explain why it works

Fixes https://github.com/TandoorRecipes/recipes/issues/4280

### DIFF
--- a/cookbook/helper/recipe_search.py
+++ b/cookbook/helper/recipe_search.py
@@ -430,7 +430,7 @@ class RecipeSearch():
         elif self._rating_gte:
             self._queryset = self._queryset.filter(rating__gte=int(self._rating_gte))
         elif self._rating_lte:
-            self._queryset = self._queryset.filter(rating__gte=int(self._rating_lte)).exclude(rating=0)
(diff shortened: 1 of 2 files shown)


vabene1111:
thanks for the PR, smilzers PR fixed this as well and I merged that first but still thank you for taking the time to fix this

PR #4316, opened 2025-12-21
Fixed saffron export not working

buergi:
Due to a typo in the frontend code the Saffron export was not working.


vabene1111:
thank you

PR #4317, opened 2025-12-21
Improved error info during import

buergi:
Forwarded error messages by `RecipeExportSerializer` to a proper exception, this way errors during importing via the Default (Tandoor) integration don't look like this

```-------------------- 
ERROR 
'NoneType' object has no attribute 'keywords'
--------------------
-------------------- 
ERROR 
'NoneType' object has no attribute 'keywords'
--------------------
Imported 0 recipes.
```

but like this

```
-------------------- 
ERROR IMPORTING 1.zip
description: Ensure This Field Has No More Than 512 Characters.
--------------------
-------------------- 
ERROR IMPORTING 264.zip
keywords[1].name: Ensure This Field Has No More Than 64 Characters.
--------------------
Imported 0 recipes.
```

Could be improved even further but I consider it is quite an improvement.

### DIFF
--- a/cookbook/integration/default.py
+++ b/cookbook/integration/default.py
@@ -27,11 +27,32 @@ class Default(Integration):
         return recipe
 
     def decode_recipe(self, string):
+        def extract_error_messages(data, path=""):
+            errors = []
+            if isinstance(data, dict):
+                for key, value in data.items():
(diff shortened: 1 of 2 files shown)


vabene1111:
awesome, thank you

PR #4334, opened 2026-01-02
Add default_unit field to space for property calculations

### LINKED ISSUE #4009: Default unit for countable foods w/ option to hide it

For property calculations to work, ingredients require a unit. `1 lemon` doesn't work, it has to be something like `1 count lemon`. 

- Add a space setting for a **default unit for countable foods**
  It could be considered to repurpose the existing "default unit" setting for this, as I think it is not very clear what the setting currently does. Afaik it only pre-populates the unit field when manually adding ingredients to recipes.
- Add a checkbox to **toggle whether or not the unit gets displayed** in ingredient lists or is only visible in editors and property tables

- Hide certain units using CSS
- Live with the awkwardness of `1 count lemon`

### Additional context

Continuing the discussion from #4005

silasg:
## Summary

- Add `default_unit` field to Space model for use in property calculations when an ingredient has no unit
- When an ingredient has no unit but the space has a default unit configured, use it for property calculations
- Show `missing_conversion` details (not `missing_unit`) when default unit exists but conversion is missing -- just like for a configured unit without conversion
- Refactor property calculation logic to extract `_try_calculate_property` helper method

### Backend

- **Model**: Added `default_unit` ForeignKey to Space (nullable, SET_NULL)
- **Serializer**: Moved `UnitSerializer` earlier in file to use in `SpaceSerializer`
- **Property Helper**: Use space's default_unit when ingredient has no unit
- **Property Helper**: Extracted `_try_calculate_property` to reduce code duplication

### Frontend

- **SpaceEditor**: Added UI to select default unit
- **Translations**: Added English and German translations

## Test Plan

- [x] Added 4 new test cases covering:
  - Default unit with working conversion
  - No default unit (marks `missing_unit`)
  - Default unit but missing conversion (shows `missing_conversion` details)
  - Default unit matches property unit (no conversion needed)
- [x] Migration tested

Partially fixes #4009 (I did not touch the display logic)

## Notes

I am neither a Vue nor a Python developer. So I used an LLM to help me with the implementation, reviewed and revised the results. But still I don't know what I don't know so please review the PR carefully.


silasg:
Any chance to get this (or another fix for the issue) merged soon? TBH I don't know if I will start using my payed SaaS Tandoor without it and would of course cancel otherwise.

vabene1111:
Hi, thank you very uch for your PR. 

To be honest I do not think that this behavior is correct. If I do not add a unit the property calculation should not just assume that something is in the default unit, most people leave out the unit when they mean pieces. 

Also when editing a recipe the default unit is automatically added so one would have to manually remove it and then it would be even more unintuitive if the property calculation function would default back to it. 

The only thing I see here is that if you import recipes and they have no unit given one could default to the default unit. But there again most sources for imports are pretty good at returning a unit if there is an actual unit of measurement and only leave it blank  if its pieces or similar. 

So maybe, as discussed in #4009 a default import unit could do the trick or maybe an automation. 

What do you think?

silasg:
Thanks for the reply @vabene1111. Maybe I got something wrong.

> If I do not add a unit the property calculation should not just assume that something is in the default unit, most people leave out the unit when they mean pieces.

That was exactly what I had in mind: If there is no unit in a recipe and the space is configured to use e.g. pieces as a default for that case, the property calculation should do exactly that: treat "no unit" as "pieces".

> The only thing I see here is that if you import recipes and they have no unit given one could default to the default unit. But there again most sources for imports are pretty good at returning a unit if there is an actual unit of measurement and only leave it blank if its pieces or similar.

Yes, adding the unit during import would be the obvious alternative. But I see some problems with that:

1. Users probably do not want to see the "pieces" written everywhere. At least I do not want. Of course, one could add a UI option to automatically hide this in display if it is the default unit as suggested. But I wonder what is the purpose to first assign a unit everywhere during import and then hide it again everywhere when displaying a recipe?! Also, it would be needed to change other UIs like kitshn App as well. Makes no sense at all. But there are more problems:
2. If we assign the unit during import, what about existing recipes without a unit? For those it would not work. 
3. Of course we could go and assign the unit to their ingredients all at once when the default unit is set. But again, this introduces more edge case and problems: 
4. What if I by accident clicked the wrong Unit, e.g. grams? Now every missing unit in the space was filled with grams and we can only fix every single one manually -- because we cannot know which one was grams originally and which one just changed by accident.

So overall, I think using the option only for property calculation is the simplest and cleanest approach. It is simple and does neither change any persisted state, nor current behaviour. It can easily be changed or turned off, if one wants so.

But as I said, maybe I got something wrong. For me, from a user perspective, I just wanted to have full features (including property calculation) without the need to manually add "pieces" (and see) everywhere.

vabene1111:
thank you for your feedback. I sadly do still disagree. I will not change the property calculation behavior to simply treat no unit as pieces.

The default unit should basically always be "grams" or whatever is the standard in the US. Using pieces is almost always a bad idea as the size of a piece changes over time. 

What I think could work is a separate "default property calculation" setting. It defaults to null and can be set to pieces or something else you want. It would need to be a space level setting and would introduce additional complexity for the user so I am not sure if this is actually desirable/better than the Solution in #4009 

@tomtjes since you brought up the other issue and have used tandoor for a long time, do you have any opinion on this?

silasg:
Again thanks for your kind reply. I still think we talk at cross purposes.

> What I think could work is a separate "default property calculation" setting. It defaults to null and can be set to pieces or something else you want. It would need to be a space level setting and would introduce additional complexity for the user

I _thought_ I had implemented exactly that. Maybe it is just a naming thing -- like often in software development -- that we have different concepts in mind when we say 'default unit'? I my naming is misleading, of course I would be happy to adjust it (and also other aspects of the feature). 

Edit: maybe it helps if I explain what I tried to achieve from my personal user perspective: I have lots of recipes with ingredients without any unit. Probably most people have. That is no problem for any feature in Tandoor except property calculation. I tried to fill this gap without the need to change e.g. "3 eggs" to "3 pieces eggs" in every recipe manually. If there would have been an automatic conversion for that, I would probably never looked at GH but would have just used that conversion. But now, when I had to think about the issue myself before trying an implementation, I thought keeping the recipes like "3 eggs" is even better than converting them, for the reasons explained above.

vabene1111:
ok, I understood it differently, if that is the case I must think about it again but that could be something that works for me

silasg:
Cool, thanks for considering. I think I read somewhere you are German, too?! If so, I am happy to have a chat/call in German as well to simplify the discussion and avoid missunderstandings.

vabene1111:
thank you I am. For the sake of others joining in on the discussion I would keep public threads in English. On discord though you can contact me in German if needed :)

silasg:
>  must think about it again but that could be something that works for me

hi @vabene1111, any update about this yet? And if so, what wording would you prefer instead of 'default unit' for the behaviour we discussed?

vabene1111:
I will get back to this when I am ready. I did a lot today but this is a bigger thing so I need some time to properly think this trough.

silasg:
I don't want to annoy you @vabene1111 , so for now I decided to run my own fork with this small change. Of course I will be happy continue the discussion and (re-)open a PR if/when you are open to include this or a comparable feature in the future. In this case _please_ let me know. I still think it is a tiny change with almost no added complexity in the code and for users who do not want this, full backwards compatibility but a huge value for everyone who wants properties also for ingredients without units.

vabene1111:
Hi, i will reopen this PR as it is still relevant. I am sorry that its taking so long but I am simply no where near having enough time to work on all things so I have to prioritise. That means sometimes things take a long time to get merged, sorry about that.

silasg:
Hi @vabene1111 sure, I am happy to hear you are still considering this. No worries, don't rush. Feel free to reach out any time, in case you want to continue the discussion or want me to make some changes to the PR. I just reverted it to its original state, since I did some changes for my local fork in this branch in the last days. I will do so in another one now, so the PR stays as it was.

tomtjes:
> @tomtjes since you brought up the other issue and have used tandoor for a long time, do you have any opinion on this?

Sorry, I missed this. It seems like this PR is better thought through than my approach. I support it.

Issue #4346, opened 2026-01-05
SOCIAL_DEFAULT_ACCESS and SOCIAL_DEFAULT_GROUP do not appear to work in the current release


mattinx:
### Tandoor Version

2.3.6

### Setup

Docker / Docker-Compose

### Reverse Proxy

No reverse proxy

### Bug description

After configuring a new installation of Tandoor and configuring SOCIAL_PROVIDERS=allauth.socialaccount.providers.discord, SOCIAL_DEFAULT_ACCESS=1, and SOCIAL_DEFAULT_GROUP=user in my docker-compose .env, and adding a new Social Provider through the Django admin interface, I'm able to login successfully via Discord. However, upon login, a new user space is created, and no access is granted to the shared space.

Digging through the code a bit, it looks like SOCIAL_DEFAULT_ACCESS and SOCIAL_DEFAULT_GROUP are only referenced within the space_overview view. However, it looks like the only references to the space_overview view were commented out by 723b74509ff8d7253148f957e9965e515ba87d96 - so I'm guessing it's been broken since then, unless I've missed something.

### Relevant logs

```shell

```

PR #4351, opened 2026-01-07
Add SOCIAL_DEFAULT_ACCESS support back in

mattinx:
Move code to handle SOCIAL_DEFAULT_ACCESS and SOCIAL_DEFAULT_GROUP settings into helper functions to restore functionality removed by commit 723b74509ff8d7253148f957e9965e515ba87d96 as identified by issue #4346

No change has been made to the logic of these settings - when SOCIAL_DEFAULT_ACCESS is enabled, users are added to the first space with the group specified by SOCIAL_DEFAULT_GROUP.

Further scope for enhancement would be to parse SOCIAL_DEFAULT_ACCESS as an integer (as indicated by the documentation) and use this as the space ID, rather than using Space.objects.first().

Another potential enhancement would be to check space membership on every login, and add the user to the configured space if they are not currently a member.

### DIFF
--- a/cookbook/helper/permission_helper.py
+++ b/cookbook/helper/permission_helper.py
@@ -491,3 +491,15 @@ def create_space_for_user(user, name=None):
         user_space.groups.add(Group.objects.filter(name='admin').get())
 
         return user_space
+
+def add_social_user_default_space(user, active=True):
+    # If enabled, add a user that has logged in with a social account to the default group
+    # Group is defined by settings.SOCIAL_DEFAULT_ACCESS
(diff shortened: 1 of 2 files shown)


vabene1111:
thank you, I would accept `SOCIAL_DEFAULT_ACCESS ` to be changed to ID based space retrieval. Do you want to implement that before I merge it?

vabene1111:
ok thanks for the PR, I think @smilerz implemented it at the same time, just in a bit of a different way #4395 

closing this for now, if anything does not work after the next update let me know

PR #4352, opened 2026-01-08
Add support to hide login form when social login is enabled

mattinx:
When using social login as the primary login method, it is sometimes desirable to hide the regular login form so users are forced to login via the social login methods.

This patch adds an additional setting controlled by the boolean environment variable HIDE_LOGIN_FORM, which when enabled prevents the display of the login form, forgot password, and sign-up buttons, as log as there is at least one available social login provider.

This behaviour can by bypassed by adding ?form=1 to the login URL, allowing the use of a local admin account


vabene1111:
thank you!

PR #4357, opened 2026-01-11
Fix URL import returning empty results and improve image download compatibility

N0mikon:
## Summary

This PR fixes two related issues with recipe importing:

### 1. URL Import Fix (line ~2394)

**Problem:** "Import from URL" on `/recipe/import` page returns empty results (200 with no recipe data), while the bookmarklet import works fine for the same URLs.

**Root Cause:** The existing code used `requests.get()` to fetch page HTML, then passed it to `scrape_html()`. Many modern recipe sites (Food Network, AllRecipes, etc.) inject their JSON-LD schema data via JavaScript. Since `requests.get()` only fetches raw HTML without executing JavaScript, the schema data was missing and `scrape_html()` couldn't find any recipe data.

The bookmarklet worked because it captures `document.documentElement.outerHTML` from the browser - the fully-rendered DOM after JavaScript execution.

**Fix:** Use `scrape_html(html=None, org_url=url, online=True, supported_only=False)` which lets recipe_scrapers handle the HTTP request internally with:
- Proper User-Agent handling
- Site-specific request logic
- Better compatibility with various recipe sites

### 2. Image Download Fix (line ~1657)

**Problem:** Recipe images failing to download from some CDNs and image hosting services.

**Root Cause:** The User-Agent was set to Firefox 86 (released February 2021), which some servers reject or serve different content to.

**Fix:** Updated to modern Chrome User-Agent with proper headers:
- Modern Chrome 120 User-Agent string
- `Accept` header with image MIME types
- `Accept-Language` header
- `Referer` header for sites that validate referrers

## Test Plan

- [x] Tested URL import with AllRecipes, Food Network URLs
- [x] Verified bookmarklet still works
- [x] Tested image downloads from various CDNs
- [x] Built and ran develop branch with patches in Docker

## Changes

- `cookbook/views/api.py`: 2 locations modified (10 insertions, 7 deletions)

---
🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/cookbook/views/api.py
+++ b/cookbook/views/api.py
@@ -1654,8 +1654,12 @@ class RecipeViewSet(LoggingMixin, viewsets.ModelViewSet, DeleteRelationMixing):
                 try:
                     url = serializer.validated_data['image_url']
                     if validate_import_url(url):
+                        # Use modern headers for better compatibility
                         response = requests.get(url, headers={
-                            "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:86.0) Gecko/20100101 Firefox/86.0"})
+                            "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36",
+                            "Accept": "image/avif,image/webp,image/apng,image/svg+xml,image/*,*/*;q=0.8",
+                            "Accept-Language": "en-US,en;q=0.9",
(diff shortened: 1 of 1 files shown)


vabene1111:
please do not post AI generated PRs without properly reviewing/understanding them. 

While updating the User agent makes sense setting the accept langauge header to en-US makes no sense in an internationally used application. One could extract that from the server settings, the request or leave it out. 

Also this is a temporary solution, maybe it would be even better to just use the headers from the request so its always a "current"/"supported" user agent.

PR #4366, opened 2026-01-18
fix(import): truncate mealie recipe names to 128 chars

rajacs50:
While importing mealie export, I ran across this ERROR: value too long for type character varying(128)
A massive amount of whitespace followed by the actual title caused this error.
Fixes by stripping white spaces, and truncates the title to the accepted 128 characters.


vabene1111:
very nice, thank you

PR #4369, opened 2026-01-19
fix: Contradictory scheme headers with external reverse proxies

### LINKED ISSUE #4187: Contradictory scheme headers - how to debug?
### Issue

After removing the nginx container and switching my Bunkerweb ReverseProxy to the Django container, I receive

web_recipes-1  | [2025-10-26 12:14:44 +0100] [33] [WARNING] Invalid request from ip=: Contradictory scheme headers
web_recipes-1  | 172.31.0.10 - - [26/Oct/2025:11:14:44 +0000] "GET /system HTTP/1.0" 400 149 "-" "Mozilla/5.0

Unfortunately I can't load the system page too.   I tried to set the needed headers like

X-Forwarded-Proto https
X-Forwarded-Protocol https

but still get the same error.  

Any hints on how to go on?  

### Tandoor Version

Latest

### OS Version

Debian 12

### Setup

Docker / Docker-Compose

### Reverse Proxy

Others (please state below)

### Other

Bunkerweb

### Environment file

```shell
# random secret key, use for example `base64 /dev/urandom | head -c50` to generate one
SECRET_KEY=<removed>

# allowed hosts (see documentation), should be set to your hostname(s) but might be * (default) for some proxies/providers
ALLOWED_HOSTS=*

# add only a database password if you want to run with the default postgres, otherwise change settings accordingly
DB_ENGINE=django.db.backends.postgresql
POSTGRES_HOST=db_recipes
POSTGRES_DB=djangodb
POSTGRES_PORT=5432
POSTGRES_USER=djangouser
```

### Docker-Compose file

```shell
networks:
  tandoor:
  bw-services:
    external: true
services:
  db_recipes:
    restart: always
    image: postgres:16-alpine
    networks:
      - tandoor
    volumes:
      - ./postgresql:/var/lib/postgresql/data
    env_file:
      - ./.env

  web_recipes:
    restart: always
    image: ghcr.io/tandoorrecipes/recipes:2.3.1
    env_file:
      - ./.env
    volumes:
      - staticfiles:/opt/recipes/staticfiles
      # Do not make this a bind mount, see https://docs.tandoor.dev/install/docker/#volumes-vs-bind-mounts
      - ./mediafiles:/opt/recipes/mediafiles
    networks:
      - tandoor
      - bw-services
    expose:
      - 80
    depends_on:
      - db_recipes

volumes:
  nginx_config:
  staticfiles:
```

### Relevant logs

```shell
web_recipes-1  | Starting gunicorn
web_recipes-1  | [2025-10-26 11:14:20 +0000] [7] [INFO] Starting gunicorn 23.0.0
web_recipes-1  | [2025-10-26 11:14:20 +0000] [7] [INFO] Listening at: unix:/run/tandoor.sock (7)
web_recipes-1  | [2025-10-26 11:14:20 +0000] [7] [INFO] Using worker: gthread
web_recipes-1  | [2025-10-26 11:14:20 +0000] [31] [INFO] Booting worker with pid: 31
web_recipes-1  | [2025-10-26 11:14:20 +0000] [32] [INFO] Booting worker with pid: 32
web_recipes-1  | [2025-10-26 11:14:20 +0000] [33] [INFO] Booting worker with pid: 33
web_recipes-1  | Running django-vite in production mode (no HMR)
web_recipes-1  | Running django-vite in production mode (no HMR)
web_recipes-1  | Running django-vite in production mode (no HMR)
web_recipes-1  | [2025-10-26 12:14:44 +0100] [33] [WARNING] Invalid request from ip=: Contradictory scheme headers
web_recipes-1  | 172.31.0.10 - - [26/Oct/2025:11:14:44 +0000] "GET /system HTTP/1.0" 400 149 "-" "Mozilla/5.0 (Windows
```

schw4rzlicht:
Fixes an issue if external reverse proxies send the `X-Forwarded-Protocol` and `X-Forwarded-Proto` headers at the same time.

Fixes #4187

### DIFF
--- a/http.d/Recipes.conf.template
+++ b/http.d/Recipes.conf.template
@@ -19,6 +19,8 @@ server {
   # pass requests for dynamic content to gunicorn
   location / {
     proxy_set_header Host $http_host;
+    proxy_set_header X-Forwarded-Protocol "";
+
     proxy_pass http://unix:/run/tandoor.sock;
 
     # param needed by django allauth sessions to log IP


vabene1111:
so first of all thank you. Second the conf.d folder is no longer used by tandoors internal nginx, it now uses the http.d and the other is just still around for legacy reasons. So if you tested this and it worked something needs to be wrong with your setup. 

Also tandoor works for all normal reverse proxies, I dont really understand why you would need to suddenly change anything about that?

schw4rzlicht:
Hey there :)

Sorry, I didn't know that I edited the wrong file. I just fixed that.

The reason I propose this change is to improve compatibility when a reverse proxy sends the two mentioned headers at the same time. When that happens, gunicorn complains about it. See #4187 for details. In that specific case the reverse proxy used is bunkerweb which in its core is just a hardened nginx reverse proxy.

vabene1111:
thank you. I still dont really see this as a tandoor issue. Why should I unset a header that some previous proxy fails to configure correctly?

I need to make the nginx config editable again somehow in the deployment process, probably using templates or something, but I dont think that I can merge this. 

Sorry if this is not what you prefer but I don't know enough about this to be sure that it can break anything else and would rather not break something else in the process.

PR #4389, opened 2026-01-29
fix: replace naive datetime.now() with timezone.now() (#3975)

### LINKED ISSUE #3975: test_search_date fails near date boundaries
### Tandoor Version

2.0.2

### Setup

Docker / Docker-Compose

### Reverse Proxy

No reverse proxy

### Other

_No response_

### Bug description

Set a fixed reference datetime (e.g., reference_date = datetime(2025, 8, 28, 1, tzinfo=timezone.utc))
Create all recipes with explicit UTC datetimes relative to reference_date
Use the same reference for the query boundary
Example:

```
from datetime import datetime, timedelta, timezone
from freezegun import freeze_time


@pytest.fixture
def found_recipe(request, space_1, accent, unaccent, u1_s1, u2_s1):
    reference_date = datetime(2025, 8, 28, 0, 0, 0)


@freeze_time("2025-08-28 12:00:00", tz_offset=0)
def test_search_date(found_recipe, recipes, param_type, result, u1_s1, u2_s1, space_1):
    # force updated_at to equal created_at datetime
    with scope(space=space_1):
        for recipe in Recipe.objects.all():
            Recipe.objects.filter(id=recipe.id).update(updated_at=recipe.created_at)

    reference_date = datetime(2025, 8, 28, 0, 0, 0)

```

smilerz:
Resolves #3975 

The test_search_date test used datetime.now() (naive) while the fixture used timezone.now() (aware). Near midnight or with timezone offsets, these resolve to different dates causing flaky assertions.

Also fixed the same issue in production code across 7 files:
- serializer.py: InviteLink rate-limit query
- views.py: cache entry and invite link validation
- AllAuthCustomAdapter.py: signup token validation and rate limiting
- integration.py: import description and export filename
- provider/{local,nextcloud,dropbox}.py: monitor.last_checked

Minor lint fixes: removed unnecessary f-string prefixes, fixed membership test operator, removed unused variable assignment.

### DIFF
--- a/cookbook/helper/AllAuthCustomAdapter.py
+++ b/cookbook/helper/AllAuthCustomAdapter.py
@@ -1,10 +1,10 @@
-import datetime
 from gettext import gettext as _
 
 from allauth.account.adapter import DefaultAccountAdapter
 from django.conf import settings
 from django.contrib import messages
 from django.core.cache import caches
+from django.utils import timezone
 
 from cookbook.models import InviteLink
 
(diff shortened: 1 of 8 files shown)


vabene1111:
thank you.

Feedback
1. is it intended that you delete the workflow files? I think this might be a merge error
2. in the InviteLinkEditor: I have just added a onBefore and onAfter save method to the base editor system if you want to use that instead of having to add a custom save method (see RecipeEditor for reference) this commit 155a8df1ab7442addc091c28464f16d767d997b0

Issue #4374, opened 2026-01-24
Inconsistent times and timezones for meal plans


c0mputerguru:
### Tandoor Version

2.3.6

### Setup

Docker / Docker-Compose

### Reverse Proxy

SWAG

### Bug description

TL;DR; I hate timezones.

I've been using the ical API to be able to sync to a google calendar for a bit.  Looks like between the last time I upgraded and now, there was a change to add meal type time information into the ical entries (https://github.com/TandoorRecipes/recipes/commit/09487a0e9478768f4e0fae7cab0e91258a63ad2c).

Prior to that change, I was getting correct info for my ical entries, and after I was getting things shifted by 1 day.  I.e. I would have an entry for Jan 24, and the ical entry would show Jan 25.  Note I am using TZ=US/Pacific on my instance as I am in the US Pacific timezone.

I think I found 2 bugs:

### Inconsistent storage of from_date and to_date for meal plans

Doing some preliminary digging, it seems like storing of the datetime in the database/backend is inconsistent.  In the meal plan UI, I can create an entry for Jan 28 Dinner (6pm).  If I then list the meal plans, I get the following:

```
{
	"22": {
		"id": 1212,
		"title": "test2",
...
		"from_date": "2026-01-28T00:00:00-08:00",
		"to_date": "2026-01-28T00:00:00-08:00",
	}
}
```

Note that the from_date is set to midnight, but still includes the timezone offset.  So the UTC time is actually 2026-01-29T08:00:00Z.

If I then open the item up and click save, and then reload, I get the following:

```
{
	"22": {
		"id": 1212,
		"title": "test2",
...
		"from_date": "2026-01-28T18:00:00-08:00",
		"to_date": "2026-01-28T18:00:00-08:00",
...
	}
}
```

So simply opening the item and saving changes the from and to dates and updates the time to the meal type time.  I think my personal habit of "must constantly hit save" means that after creating the item, I always hit save, which meant that all my entries always got updated to the meal type time.  Which is why I never experienced what was described in https://github.com/TandoorRecipes/recipes/issues/4236.

What I would expect is that things would be consistent; simply opening and saving an unchanged item should not change any of the fields.  So either the original create call should offset the time stored in the database based on the meal type the same way an update does, or the update should not do that offset.  Fixing this inconsistency helps create a clearer path forward for the issue below.

[...]

PR #4398, opened 2026-01-31
#4374 — Time picker & iCal timezone

smilerz:
Fixes: #4374

  **Root Cause**

  Meal plan time handling had multiple interrelated problems across the stack:

  iCal export produced naive datetimes. meal_plans_to_ical() used datetime.combine(p.from_date, p.meal_type.time) to build event timestamps. combine() extracts the date portion from from_date
  (which is stored as UTC) and pairs it with meal_type.time (a naive time object), producing a naive datetime with no timezone info. This meant iCal clients had no timezone context and would
  render events in whatever local zone they assumed, the UTC date component was treated as a local date so events near midnight could shift to the wrong calendar day, and if meal_type.time was
  null only the date was emitted with no time at all.

  The frontend mutated shared state in the calendar view. MealPlanView.vue called startDate.setHours() directly on the fromDate/toDate objects from the Pinia store to apply mealType.time for
  calendar rendering. setHours() mutates the Date in-place, so this silently corrupted the store's data. Any subsequent save or drag operation would persist whatever time the calendar view had
  stamped, not what the user intended.

  There was no way to set or view time in the editor. The meal plan editor had no time picker. Users could only set a date. The only time that existed came from meal_type.time, applied
  inconsistently — backend iCal used it, frontend calendar mutated it in, but the actual from_date/to_date fields stored on the model had no meaningful time.

  Drag-and-drop discarded time. When dragging a meal plan to a new date, the code set mealPlan.fromDate = targetDate directly. targetDate comes from the calendar component as a midnight Date
  object, so the existing time on the meal plan was silently replaced with 00:00.

  **Changes**

  Backend iCal export (cookbook/views/api.py): Removed datetime.combine() entirely — from_date and to_date are already timezone-aware DateTimeFields and now carry the correct time. They are used
   directly as dtstart/dtend, and icalendar serializes aware datetimes with proper VTIMEZONE components automatically. When to_date <= from_date (same-moment events), dtend is set to from_date +
   60 minutes so events are visible in calendar clients. Removed the meal_type.time dependency from iCal — time is now on from_date itself, so the export has one source of truth.

  Backend serializer default time (cookbook/serializer.py): Added _apply_default_time() static method to MealPlanSerializer.create() with a priority chain: (1) explicit time — if the incoming
  datetime has a non-midnight local time, it's preserved as-is; (2) meal type time — if the datetime is midnight local (i.e. the frontend sent a date-only string like 2026-06-15) and the meal
  type has a time field, that time is applied; (3) noon fallback — if neither condition is met, defaults to 12:00 local time so events don't land at midnight. This runs on from_date and to_date
  during create() only — updates preserve whatever time is already stored.

  Frontend time picker in MealPlanEditor (MealPlanEditor.vue): Added time picker using v-text-field + v-menu + v-time-picker (same pattern as MealTypeEditor.vue). Date and time fields are
  side-by-side on desktop (sm+ breakpoint) using v-row no-gutters, stacked on mobile. Custom CSS hides individual Vuetify field overlays and applies a single shared background to eliminate the
  visual seam between date and time fields. applyTimeToEditingDates() applies the selected time to both fromDate and toDate via Luxon .set() (immutable, no mutation). A meal type watcher updates
   the time picker when the user changes meal type. New meal plans initialize to noon today, then apply meal type time if available, then apply any itemDefaults from the calendar. Existing items
   initialize the time picker from fromDate. updateDate() calls applyTimeToEditingDates() to re-stamp the selected time when the date range changes.

  Frontend calendar view fixes (MealPlanView.vue): Removed the setHours() mutation — planItems computed property now passes fromDate/toDate directly to the calendar component since time is
  stored correctly on the model. Drag-and-drop extracts hour/minute/second from the existing fromDate via Luxon and applies them to the target date using .set(), so moving a 17:00 dinner to
  Tuesday keeps it at 17:00.

  **Tests**
  - test_ical_event_timezone_aware
  - test_ical_event_minimum_duration
  - test_create_date_only_gets_noon_default
  - test_create_date_only_with_meal_type_time
  - test_create_explicit_time_preserved

### DIFF
--- a/cookbook/serializer.py
+++ b/cookbook/serializer.py
@@ -1384,11 +1384,39 @@ class MealPlanSerializer(SpacedModelSerializer, WritableNestedModelSerializer):
     def in_shopping(self, obj):
         return obj.shoppinglistrecipe_set.count() > 0
 
+    @staticmethod
+    def _apply_default_time(dt, meal_type_obj):
+        """Apply default time to a datetime that has no explicit time (midnight local).
+
+        Priority: explicit time > meal_type.time > noon fallback.
+        Returns the datetime unchanged if it already has a non-midnight local time.
+        """
+        local_dt = timezone.localtime(dt)
+        if local_dt.hour != 0 or local_dt.minute != 0 or local_dt.second != 0:
+            return local_dt
+
+        if meal_type_obj and meal_type_obj.time:
+            return local_dt.replace(hour=meal_type_obj.time.hour, minute=meal_type_obj.time.minute, second=0)
+        return local_dt.replace(hour=12, minute=0, second=0)
+
     def create(self, validated_data):
         validated_data['created_by'] = self.context['request'].user
 
+        meal_type_obj = None
+        meal_type_data = self.context['request'].data.get('meal_type')
+        if isinstance(meal_type_data, dict):
+            meal_type_id = meal_type_data.get('id')
+        else:
+            meal_type_id = meal_type_data
+        if meal_type_id:
+            meal_type_obj = MealType.objects.filter(pk=meal_type_id, space=self.context['request'].space).first()
+
+        validated_data['from_date'] = self._apply_default_time(validated_data['from_date'], meal_type_obj)
+
         if 'to_date' not in validated_data or validated_data['to_date'] is None:
             validated_data['to_date'] = validated_data['from_date']
+        else:
+            validated_data['to_date'] = self._apply_default_time(validated_data['to_date'], meal_type_obj)
 
         add_to_shopping = False
         try:
--- a/cookbook/views/api.py
+++ b/cookbook/views/api.py
@@ -3164,14 +3164,10 @@ def meal_plans_to_ical(queryset, filename):
         event['uid'] = p.id
 
         start_date_time = p.from_date
-        end_date_time = p.from_date
+        end_date_time = p.to_date if p.to_date else p.from_date
 
-        if p.to_date:
-            end_date_time = p.to_date
-
-        if p.meal_type.time:
-            start_date_time = datetime.datetime.combine(p.from_date, p.meal_type.time)
-            end_date_time = datetime.datetime.combine(p.to_date, p.meal_type.time) + datetime.timedelta(minutes=60)
+        if end_date_time <= start_date_time:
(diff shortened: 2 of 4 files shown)


vabene1111:
hi, thanks for the PR, some feedback

1. I think you have worked with AI, which is fine for me, but please keep the PR description short. By the time I have read this wall of text I have implemented the feature myself. 
2. this is a lot of code and complexity added to the UI. I added times to the meal type because I tought it would make it easier to display entries into a calendar once I finally get around to add proper ical sync endpoints. Wouln't the easies solution be to export to export the time on the ICal entry in the time configured in TZ in the settings? this would be like one line of code, work and not introduce any complexity? or am I missing something?

smilerz:
> hi, thanks for the PR, some feedback

> 2. this is a lot of code and complexity added to the UI. I added times to the meal type because I tought it would make it easier to display entries into a calendar once I finally get around to add proper ical sync endpoints. Wouln't the easies solution be to export to export the time on the ICal entry in the time configured in TZ in the settings? this would be like one line of code, work and not introduce any complexity? or am I missing something?

The time on the meal plan gets mutated in several places that destroys the default time on the meal plan.  My thought was if time is important why not allow setting it directly.  
If you only want time to be applied if it is configured on the meal plan I can change the code to do that consistently.

vabene1111:
I really hate timezones .. If we add time to the UI we would kinda also need to add a timezone selector and then we are starting to build a calendar application .. 

but we cant really not do it or we have the stupid date shifting issue again ..

Maybe just select type > time gets populated > you can change time > its respected for the from/to date > its exported in ical

Will look at the code again

smilerz:
> I really hate timezones .. If we add time to the UI we would kinda also need to add a timezone selector and then we are starting to build a calendar application ..

I hate time - huge pain.  That being said, I think that it is reasonable to make the assumption that timezone is always the users  timezone as detected by the browser - I can't think of any reason to support setting timezone directly.

vabene1111:
yes, I merged and looked at it and I think its fine. The only thing I see is that it might annoy users that feel obligated to set a time every time when they edit a meal plan (we germans do like to fill our forms :D) but I guess thats the price to pay. 

Thanks for the pr

PR #4399, opened 2026-01-31
Clarify and improve import/export documentation

loucasal:
- Clarified which Mealie exports are supported
- Corrected capitalization, punctuation and spelling throughout the document (including in app names)

Not done yet, but could be done if desired:
- Reorder the list of integrations alphabetically (or according to other criteria)
- Make sure integrations have the same order in the overview and in the page
- ~~Add links from the overview to the different sections of the page~~ I noticed there is a ToC in the sidebar, so this one won't be needed.


vabene1111:
thank you. I think the order is fine for now.

PR #4422, opened 2026-02-09
Move Norwegian locale from nb_NO to nb to match Django language code

Ziggiz:


vabene1111:
fixed by smilerz along  with some other stuff, still thanks for the PR

PR #4424, opened 2026-02-09
Update Language folder in Dev for Norwegian to support Django "nb"

Ziggiz:
Locale folder name and config in po


smilerz:
I think this should already exist now - can you confirm?

PR #4425, opened 2026-02-09
Beta Norwegian language folder to support Django

Ziggiz:


smilerz:
I think this should already exist now - can you confirm?

PR #4426, opened 2026-02-10
Fix unit conversion query explosion

poikilotherm:
On my instance, after upgrading to 2.5.0, I experienced two huge N+1 query problems with unit conversions. Loading a recipe took about 17 seconds on my HP Proliant Micro Gen 8. Please find my fixes attached. 

I did apply this to my instance and it works, but I am in no way familiar with your codebase, so wont be able to add any tests etc.

Let me know if I can help with anything else to get this PR merged.

Thanks for your hard work on all of this!

### DIFF
--- a/cookbook/helper/unit_conversion_helper.py
+++ b/cookbook/helper/unit_conversion_helper.py
@@ -123,14 +123,14 @@ class UnitConversionHelper:
             while queue:
                 current = queue.pop(0)
                 for c in current.unit.unit_conversion_base_relation.all():
-                    if c.space_id == self.space.id:
+                    if self.space and c.space_id == self.space.id:
                         r = self._uc_convert(c, current.amount, current.unit, ingredient.food)
                         if r and r.unit.id not in visited_unit_ids:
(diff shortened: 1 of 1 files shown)


vabene1111:
thank you very much. Might be related to some django model fetching. If this helped I do have an idea why. would also resolve #4430 i think

poikilotherm:
If it helps, happy to copy & paste the SQL middleware it put in place to learn who's the culprit generating the N+1. Thanks for merging, happy to see this project accepting external contributions!

Also: I did put some explanation into the commit messages to document the "why". Maybe these help?

vabene1111:
oh very nice, i never read commit messages of PR's but maybe I should. Though I do understand it, has to do with the way django handles query and model retrievals, just did not have much time and could not test with a complete reproducible example. 

Thanks for the contribution, I am always happy about things like this. Bug fixes and QOL sutff is always great. New features are also ok but I ask to discuss them first with me and try to keep the complexity down as its usually me how has to maintain them for many years to come :)

PR #4427, opened 2026-02-10
Correctly document options for `DB_ENGINE`

mark-monteiro:
Although PostgreSQL is the recommended engine, the default option is actually SQLite (presumably for backwards compatibly).

See: [recipes/settings.py#L501](https://github.com/TandoorRecipes/recipes/blob/master/recipes/settings.py#L501)


vabene1111:
you are correct, since i always had postgres in the sample .env I assumed loading that as the default but that is not true and it might confuse people.

mark-monteiro:
Hey @vabene1111 just checking in to see if you perhaps forgot to click merge on this after commenting?

vabene1111:
thanks for the ping, not forgotten, just did not have the time to merge and properly credit you in the changelog yet.

Issue #4443, opened 2026-02-14
Mealie 1.x Import Bugs


Mardis:
### Tandoor Version

2.5.0

### Setup

Docker / Docker-Compose

### Reverse Proxy

No reverse proxy

### Bug description

2 bugs I've found in the import process:

1. Importing when a step has an empty name causes a crash leading to recipes imported without steps, images, etc.
2. Importing recipes with ingredient sections breaks the order/sections when moved over as order is not preserved on import

### Relevant logs

```shell

```

vabene1111:
thanks for the report. It would help me a lot to get one example each (preferably in different files) so I can reproducde and fix this easier.

Mardis:
Went ahead and tried my hand at a quick fix: https://github.com/TandoorRecipes/recipes/pull/4454 

**Out of Order Example:**
Original in Mealie::
![Image](https://github.com/user-attachments/assets/e04e1ded-df7c-48fa-98c0-e60c9bf79a62)

Out of order import:
![Image](https://github.com/user-attachments/assets/901a1916-0e14-49ff-ada3-498052bd7398)

In order fix: 
![Image](https://github.com/user-attachments/assets/f06cc111-61a0-4e0e-92ac-8e0b52a4c915)


**Null Step Title Example:**
From my exported `database.json` here is an example of a step that was causing the import to fail
```json
{"created_at": "2026-02-13T06:07:30.456916", "update_at": "2026-02-13T06:07:30.456918", "id": "8a3af780-a560-4da6-bb57-aa766dbcf5e5", "recipe_id": "4bcff776-efa5-483b-a7a1-050c37915210", "position": 0, "type": "", "title": null, "text": "Combine all ingredients.", "summary": ""}
```
Before null step title fix:
![Image](https://github.com/user-attachments/assets/85694b36-6438-4267-9723-8c515d20e30e)

PR #4454, opened 2026-02-18
Fixing Mealie 1.x import to default null step name to empty string and retain ingredient ordering

Mardis:
Added 2 fixes to the Mealie 1.x integration as identified in: https://github.com/TandoorRecipes/recipes/issues/4443
* Added support to default `null` step names to empty strings
* Retain ingredient order to protect ingredient section label position


vabene1111:
awesome, thanks, will merge this for the next release

PR #4464, opened 2026-02-21
Update German translations for 'cosmetic' term

nepda:
As a German, the term "Kosmetisch" is correct, but not very common for things like user interface / design settings.


vabene1111:
Hmm I am not really a fan of "erscheinungsbild" .. (i am german as well). Its probably often "Anzeige" or something. Windows calls it "Personalisieren". 

Since I do not liek Erscheinungsbild mor than Kosmetisch I will leave it for now.

PR #4472, opened 2026-02-25
fix: add missing JSON keyword to AI properties prompts and improve error handling

### LINKED ISSUE #4384: AI properties method fails and the exception message is not passed
### Tandoor Version

2.3.6

### Setup

Kubernetes

### Reverse Proxy

Others (please state below)

### Other

ingress-nginx

### Bug description

Tested with gpt-5-mini.

[aiproperties(self, request, pk)](https://github.com/TandoorRecipes/recipes/blob/develop/cookbook/views/api.py#L1103)

Two issues:

1. The prompt is missing the word 'JSON', causing the request to throw an exception.
2. The try/except block has an indent error so this exception is never passed to the user.

Fixing the try/except block allows the error to propogate correctly. Adding the word 'JSON' to the prompt allows the method to complete sucessfully.

### Relevant logs

```shell

```

stuckvgn:
## Summary

Fixes #4384

The `aiproperties` action on both `FoodViewSet` and `RecipeViewSet` uses `response_format: {"type": "json_object"}` with litellm, but the prompt text never contains the word **JSON**. Many LLM providers (OpenAI, Azure, etc.) require the word "JSON" to appear somewhere in the messages when using structured JSON output mode -- without it, the API returns a `BadRequestError`.

### Changes

1. **Added "JSON" to prompt text** in both `FoodViewSet.aiproperties` and `RecipeViewSet.aiproperties` -- changed "in the following format" to "in the following JSON format" and appended "Return only valid JSON." This matches the pattern already used by the file/text import prompts elsewhere in the same file.

2. **Added generic `Exception` handler** to the `try/except` blocks in all three AI action methods (`FoodViewSet.aiproperties`, `RecipeViewSet.aiproperties`, and `AiStepSortView`). Previously, only `LitellmTimeout` and `BadRequestError` were caught -- any other exception (e.g., `JSONDecodeError` from malformed AI response, unexpected litellm errors) would propagate as an unhandled 500 with no useful information returned to the user. The new handler logs the full traceback server-side and returns a descriptive error message to the client.

### Testing

- Verified the prompt text change is consistent with the existing AI import prompts in the same file (lines 2668 and 2686) which already include "JSON"
- The generic exception handler follows the same pattern as the existing `BadRequestError` handler but with `HTTP_500_INTERNAL_SERVER_ERROR` status
- No new dependencies or migrations required

### DIFF
--- a/cookbook/views/api.py
+++ b/cookbook/views/api.py
@@ -1203,7 +1203,7 @@ class FoodViewSet(LoggingMixin, TreeMixin, DeleteRelationMixing):
                     "content": [
                         {
                             "type": "text",
-                            "text": "Given the following food and the following different types of properties please update the food so that the properties attribute contains a list with all property types in the following format [{property_amount: <the property value>, property_type: {id: <the ID of the property type>, name: <the name of the property type>}}]."
+                            "text": "Given the following food and the following different types of properties please update the food so that the properties attribute contains a list with all property types in the following JSON format [{property_amount: <the property value>, property_type: {id: <the ID of the property type>, name: <the name of the property type>}}]. Return only valid JSON."
                                     "The property values should be in the unit given in the property type and for the amount specified in the properties_food_amount attribute of the food, which is given in the properties_food_unit."
                                     "property_amount is a decimal number. Please try to keep a precision of two decimal places if given in your source data."
                                     "Do not make up any data. If there is no data available for the given property type that is ok, just return null as a property_amount for that property type. Do not change anything else!"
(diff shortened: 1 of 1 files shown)


samtuckerdavis:
Friendly ping — this adds a missing JSON keyword to the AI properties prompts and improves error handling. CLA is signed, small fix. @vabene1111 would appreciate a look when convenient.

vabene1111:
thanks for the PR, will get it merged. 

As I only work on this in my free time and have a lot of other things to do at the moment it often takes a bit for me to answer, but I always look at everything :)

PR #4483, opened 2026-03-01
Implement Food.diet field and activate FoodInheritField inheritance

### LINKED ISSUE #2022: Allergen/diet flagging

_No response_

Hi, love the project so far! I have a number of food allergies and it's really annoying to filter recipes manually - a feature that would make recipe management much easier for me is to be able to define an allergen, like say mushrooms, and have it flagged in some way after import, is this something that would be feasible? Thank you!

_No response_

### Additional context

_No response_

stuckvgn:
## Summary

The `diet` field was scaffolded in `FoodInheritField` during the initial migration (`FoodInheritField(name='Diet', field='diet')`), but has been excluded from `Food.inheritable_fields` with a TODO comment and has had no corresponding model field — so the API and UI have had nowhere to store or filter by it.

This PR completes the scaffolding by wiring up the full stack.

## Changes

**`cookbook/models.py`**
- Adds `Food.diet` as a nullable `CharField(max_length=32, choices=DIET_CHOICES)` with choices: `VEGAN`, `VEGETARIAN`, `PESCATARIAN`, `GLUTEN_FREE`, `LACTOSE_FREE`, `HALAL`, `KOSHER`
- Removes `'diet'` from the `inheritable_fields` exclusion list (`substitute` remains excluded as it is a M2M and behaves differently)
- Extends `Food.reset_inheritance` to propagate `diet` down the food tree using the same null-guard pattern as `supermarket_category`: an unset diet on a parent does not cascade

**`cookbook/signals.py`**
- Extends `update_food_inheritance` to apply `diet` parent → child and parent → direct children, following the same guard used for `supermarket_category`
- Extracts the repeated `instance.inherit_fields.values_list(...)` call to a local variable to avoid N redundant queries per child

**`cookbook/serializer.py`**
- Adds `'diet'` to `FoodSerializer.Meta.fields`

**`cookbook/views/api.py`**
- Adds `?diet=` query parameter filter to `FoodViewSet.get_queryset`
- Adds `@extend_schema_view` OpenAPI documentation for the new parameter

**`cookbook/migrations/0238_food_diet_field.py`**
- Additive: `null=True, blank=True, default=None` — no effect on existing rows, no data loss, no default imposed

## Backward compatibility

Existing `Food` records get `diet=NULL`. The field is optional in the serializer. No existing API consumers are affected.

Closes #2022

### DIFF
--- a/cookbook/migrations/0238_food_diet_field.py
+++ b/cookbook/migrations/0238_food_diet_field.py
@@ -1,0 +1,30 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('cookbook', '0237_remove_mealtype_mt_unique_name_per_space_and_more'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='food',
+            name='diet',
+            field=models.CharField(
+                blank=True,
+                choices=[
+                    ('VEGAN', 'Vegan'),
+                    ('VEGETARIAN', 'Vegetarian'),
+                    ('PESCATARIAN', 'Pescatarian'),
+                    ('GLUTEN_FREE', 'Gluten Free'),
+                    ('LACTOSE_FREE', 'Lactose Free'),
+                    ('HALAL', 'Halal'),
+                    ('KOSHER', 'Kosher'),
+                ],
+                default=None,
+                max_length=32,
+                null=True,
+            ),
+        ),
+    ]
--- a/cookbook/models.py
+++ b/cookbook/models.py
@@ -773,8 +773,26 @@ class Unit(ExportModelOperationsMixin('unit'), models.Model, PermissionModelMixi
 class Food(ExportModelOperationsMixin('food'), TreeModel, PermissionModelMixin):
     # TODO when savings a food as substitute children - assume children and descednants are also substitutes for siblings
     # exclude fields not implemented yet
-    inheritable_fields = FoodInheritField.objects.exclude(field__in=['diet', 'substitute', ])
+    inheritable_fields = FoodInheritField.objects.exclude(field__in=['substitute', ])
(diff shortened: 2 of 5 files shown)


samtuckerdavis:
## CI Failures are pre-existing in `develop`

The 13 failing tests are **not caused by this PR**. They are pre-existing failures introduced into the `develop` branch by a recent upstream commit before this PR was opened.

I confirmed this by checking the develop branch CI directly — it fails with the **exact same 13 tests**:

**10 `test_makenow_filter` failures** (`AttributeError: '' object has no attribute 'user_space'`):
- Caused by the `make meal type space global` commit to `develop` which changed `recipe_search.py` line 529 from `[*self._request.user.get_shopping_share(), self._request.user]` to `self._request.user_space.household.values_list('user_id', flat=True)`
- The test mocks in `test_makenow_filter.py` use `type('', (object,), {'space': space_1, 'user': user1})()` which has no `user_space` attribute
- Additionally, `household.values_list('user_id', flat=True)` appears to call `values_list` on a `Household` FK instance rather than a queryset (the pattern used elsewhere in `views/api.py` is `UserSpace.objects.filter(household=...)`)

**2 `test_api_meal_type` failures** (guest `g1_s1` getting 403 instead of 200/201):
- `test_list_permission[arg1]` and `test_add[arg1]` — the "space global" meal type change appears to have altered permissions for guest users

**1 `test_api_food.py::test_onhand` failure** (`assert False is True`):
- The `food_onhand` PATCH is not being saved — likely related to the household sharing changes in `develop`

This PR only adds the `diet` field to `Food`, a migration, and related serializer/signal/API changes. None of the failing test files are touched by this PR.

The `develop` branch CI has been broken since ~2026-02-28. Once those upstream issues are resolved, this PR should be ready for re-review.

samtuckerdavis:
Closing this PR — on reflection, this change may not be the right fit for this project. Thanks for your time!

PR #4491, opened 2026-03-04
fix: propagate user language to Luxon, Vuetify, and vue-simple-calendar

MrAdam:
Dates across the app were formatted using the browser/OS locale instead of the user's selected language in Tandoor.

## Changes
- Set `Settings.defaultLocale` in `setLocale()` so all `DateTime.toLocaleString()` calls use the correct locale
- Derive Vuetify's initial locale from the HTML `lang` attribute at startup so date pickers render correctly
- Pass the vue-i18n locale to `vue-simple-calendar` so meal plan calendar day headers follow the user's language

## Future improvement
Vuetify provides a `createVueI18nAdapter` that would keep Vuetify's locale automatically in sync with vue-i18n, removing the need for the HTML `lang` workaround. This requires migrating to vue-i18n's composition mode (`legacy: false`), which is a larger refactor.

### DIFF
--- a/vue3/src/components/display/MealPlanView.vue
+++ b/vue3/src/components/display/MealPlanView.vue
@@ -5,6 +5,7 @@
                 <!-- TODO add hint about CTRL key while drag/drop -->
                 <!-- TODO multi selection? date range selection ? -->
                 <calendar-view
+                    :locale="locale"
                     :show-date="calendarDate"
                     :items="planItems"
                     class="theme-default"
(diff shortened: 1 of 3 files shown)


vabene1111:
hi, thanks for the PR. I will manually merge this as @smilerz did some work on the locales which breaks merging. 

Feel free to send me a message if you want to refactor i18n to the new API, I never quite understood their docs about which way of setting things up is which version. I was happy when the whole async loading thing worked 😂

PR #4497, opened 2026-03-06
docs: add note about local storage only on https

probablyVincentIp:
Added a note about the need for https for PWA local storage for users with less experience with service workers

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -4,6 +4,7 @@ Many of those shortcomings are planned to be fixed in future release but simply 
 
 ## Is there a Tandoor app?
 Tandoor can be installed as a progressive web app (PWA) on mobile and desktop devices. The PWA stores recently accessed recipes locally for offline use.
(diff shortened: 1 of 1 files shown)


vabene1111:
while there might be some local settings that do not work when local storage is not available in an insecure context recipes are always stored on the server in a database regardless if http or https is used. 

I am not sure which issue you encountered and what exactly caused it but this seem to me to be not correct. Generally there is not really any reason to do anything with http today. 

Still thank you for trying to help and updating the docs.

probablyVincentIp:
@vabene1111 when a user runs Tandoor on a local network and calls it from a different device (eg `http://192.168.1.xzy`), this will not provide a secure connection without additional network configuration, and the browser does't have this whitelisted (ie `localhost`).

Advanced users will be able to recognize the issue but if you are having a quick "lets spin a container up to see if it works" it might confuse users why saving to local storage does not work

PR #4508, opened 2026-03-09
Feature/inline ingredient notes

lewisburgess:
Included a feature to show the ingredients notes in-line instead of the original bubble tooltip. The setting can be switched in the Settings - Cosmetic - Preferences section between the original behaviour and the in-line notes. 

image


smilerz:
This is included in #4601 which provides greater flexibility on for user controlled behavior

Issue #4448, opened 2026-02-16
Books' Order has no effect


Matt-Gore:
### Tandoor Version

2.5.3

### Setup

Docker / Docker-Compose

### Reverse Proxy

No reverse proxy

### Bug description

The Books page shows books from newest to oldest instead of respecting the configured ordering.

### Relevant logs

```shell

```

PR #4515, opened 2026-03-13
respect book ordering by default

Matt-Gore:
fixes https://github.com/TandoorRecipes/recipes/issues/4448


vabene1111:
thanks for the PR

PR #4519, opened 2026-03-14
fix: default book ordering uses 'order' field instead of 'id' (#4448)

### LINKED ISSUE #4448: Books' Order has no effect

smilerz:
Resolves #4448

- Default order_field changed from id to order
- Added -id secondary sort for stable ordering when order values tie


vabene1111:
i think someone else already did this so closing, thanks

Issue #4547, opened 2026-03-22
Invalid HTTP Response Code in apiRecipeShoppingUpdate


jonsch318:
### Tandoor Version

2.5.3

### Setup

Kubernetes

### Reverse Proxy

Others (please state below)

### Other

Envoy Gateway -> Nginx for static content 

### Bug description

When sending such an http request either manually or with the kitshn app with correct id's and vars i got an 

502 Bad Gateway upstream connect error or disconnect/reset before headers. reset reason: protocol error

from my envoy gateway.

After some investigation this is due to it rejecting the incorrect response code 204 (NO CONTENT) of the apiRecipeShoppingUpdate request that has a body content.

Although the API documentation states a 200 response the request handler response with a 204 and a msg content.
 
https://github.com/TandoorRecipes/recipes/blob/92d6bdc4dac2a958aa0457719ee510fa2cb9f23f/cookbook/views/api.py#L1854-L1883 

```bash
curl --request PUT \
  --url https://$HOST/api/recipe/{id}/shopping/ \
  --header 'authorization: Bearer $TOKEN' \
  --header 'content-type: application/json' \
  --data '{
    "id": "15",
    "ingredients": [
      163
    ],
    "serving": "1"
  }'
```

### Relevant logs

```shell
$IP - - [22/Mar/2026:12:56:18 +0100] "PUT /api/recipe/21/shopping/ HTTP/1.1" 204 119 "-" "curl/8.19.0"
```

PR #4548, opened 2026-03-22
fix: invalid response code in apiRecipeShoppingUpdate

jonsch318:
This fixes the incorrect 204 NO_CONTENT response, where a content is given in apiRecipeShoppingUpdate.

Since the API schema also specifies a 200 OK code this should be used.


jonsch318:
Fixes https://github.com/TandoorRecipes/recipes/issues/4547

PR #4559, opened 2026-03-27
feat: add Bring shopping list connector

i-ilak:
Implementation following discussion in #911 

One-way sync from Tandoor to [Bring](https://www.getbring.com) via the connector framework. 

First contribution here, feedback very welcome!

# Design choices

## Generic field names instead of bring_-prefixed fields. 
This follows the Storage model pattern where the same DB columns serve different connector types and the frontend conditionally shows/hides them with v-if. Renamed `todo_entity` -> `list_id` since it works as a generic identifier for both HA entity IDs and Bring list UUIDs.

## [`bring-api`](https://github.com/miaucl/bring-api) library
Instead of raw `aiohttp` calls. The Bring API only needs 3 endpoints (login, save_item, remove_item), so a direct implementation would be ~120 lines. I went with the library because it's maintained by the Home Assistant community and shields us from auth flow or endpoint changes upstream. Transitive deps are lightweight (mashumaro, orjson). Happy to rewrite if fewer deps is preferred.

## Auth retry limit
After 3 consecutive failures the connector stops trying until the ConnectorConfig is saved again (which triggers a rebuild). Without this, every shopping list operation would hammer the Bring API with bad credentials.

## Password as plaintext
The current implementation stores the password in plain text into the DB. The HA token is also stored in plain text, and as far as I can tell the rest of the project handles this similarly. 

All fields returned regardless of type (Bring connectors return `url: null`, HA returns `email: null`). Considered a to_representation override to strip irrelevant fields, but the Storage serializer does the same "return everything" approach, so I stayed consistent.


i-ilak:
@vabene1111 

As discussed, a PR implementig what we discussed in #911. 

Tbh, Im not particularly familiar with vue, so I just followed the developer guidelines. If any of the files should not have been committed, let me know, Ill remove it.

PR #4562, opened 2026-03-28
fix: allow ingredient amount null by auto-setting no_amount true

### LINKED ISSUE #4416: Improve recepie editor - No ingredient amount
If i accidentally leave the amount of an ingredient empty an marke it as "no amount" then a red poput tells me that it can't be left empty. Therefor it would be a nice quality of life feature if it resets the the amout automaticlly to 0,00 or the pop-up is disabled if no amount is needet.

Matt-Gore:
Resolves https://github.com/TandoorRecipes/recipes/issues/4416

### DIFF
--- a/cookbook/serializer.py
+++ b/cookbook/serializer.py
@@ -131,12 +131,17 @@ class CustomDecimalField(serializers.Field):
     and allow commas as decimal separators
     """
 
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


vabene1111:
hi, thanks for the PR. This approach is different to how I have handled this in the past but its a nice solution. I will think about it and review the PR.

vabene1111:
fixed this another way, still thank for the PR

PR #4628, opened 2026-04-16
Stats

ac615223s5:
shows total properties of logged cooking for each day/week


vabene1111:
thank you for your PR. I want to do something like this but I do not just want to slap on a page and display the stuff. I want to integrate it into the application and am not sure how to best do that yet, so I will likely not merge this. 

Can you post some screenshots so I can take a look and learn from what you did?

ac615223s5:
image
image

my use case is to track nutrients and money per day/week so a separate stats tracking page seemed like the most obvious idea. How else should this be integrated?

vabene1111:
thank you. Feel free to use the page but this is not how I want it implemented. It also seems to conain errors on the screenshot (NAN) so I dont want to spend the time reviwing this as its likely equal to me just implementing it the way I want.

PR #4644, opened 2026-04-24
feat: add support for richer reverse proxy auth and custom logout URL

### LINKED ISSUE #1134: Richer reverse proxy authentication

Being able to use Authelia SSO without much fuss is great, but unfortunately the only piece of user information that recipes grabs right now is the HTTP_REMOTE_USER.

Furthermore, the logout link is hard coded and cannot be overridden with an environment variable. This prevents being able to use an SSO logout url e.g. https://authelia.example.com/logout

Add support for other header fields like:
HTTP_REMOTE_GROUPS
HTTP_REMOTE_NAME
HTTP_REMOTE_EMAIL

Update this information as required when a user logs in e.g. their email address has changed.

Add an environmental variable REVERSE_PROXY_AUTH_LOGOUT to specify a custom logout URL.

I guess I could manage users in Recipes as well as in Authelia, but that means double-handling of information for my use case.

**Additional context**

sne-ha091:
This PR addresses issue #1134 by extending the reverse proxy authentication capabilities and allowing for a custom logout redirect.
Changes:
User Data Sync: Update middleware.py to support HTTP_REMOTE_NAME and HTTP_REMOTE_EMAIL headers. The user's name and email are now automatically updated in the database upon login if these headers are present.

Custom Logout: Added a new environment variable REVERSE_PROXY_AUTH_LOGOUT in settings.py.

Frontend Update: Modified the logout logic to redirect to the custom SSO logout URL if the environment variable is set , preventing the hardcoded redirect to the internal login page.

Fixes: #1134

### DIFF
--- a/recipes/middleware.py
+++ b/recipes/middleware.py
@@ -7,6 +7,31 @@ from django.db import connection
 
 class CustomRemoteUser(RemoteUserMiddleware):
     header = getenv('PROXY_HEADER', 'HTTP_REMOTE_USER')
+    
+    # Add this method to capture extra headers
+    def process_request(self, request):
+        if not settings.REMOTE_USER_AUTH:
+            return
+            
+        # Standard Django RemoteUserMiddleware logic
+        super().process_request(request)
+        
+        # If user is authenticated, sync the new fields
+        if request.user.is_authenticated:
+            email = request.META.get('HTTP_REMOTE_EMAIL')
+            name = request.META.get('HTTP_REMOTE_NAME')
+            
+            updated = False
+            if email and request.user.email != email:
+                request.user.email = email
+                updated = True
+            if name and request.user.first_name != name:
+                request.user.first_name = name
+                updated = True
+            
+            if updated:
+                request.user.save()
+
 
 
 """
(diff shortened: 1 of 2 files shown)


tejash5489-lang:
Looked at this alongside #4737 while triaging #1134. A couple of notes:

- `REVERSE_PROXY_AUTH_LOGOUT` is defined in `settings.py` but I couldn't find it referenced anywhere else in the diff or branch (`git grep REVERSE_PROXY_AUTH_LOGOUT` only turns up the definition). The PR description mentions a "Frontend Update: Modified the logout logic to redirect to the custom SSO logout URL", but that change doesn't appear to be in the diff — as it stands the env var has no effect, so item 3 from the issue (custom logout URL) isn't actually wired up yet.
- The `HTTP_REMOTE_USER`/name/email sync logic in `middleware.py` itself looks correct and does address item 1/2 of the issue.
- No test coverage for the new header-sync behavior.

Might be worth pushing a follow-up commit that actually consumes `REVERSE_PROXY_AUTH_LOGOUT` (e.g. overriding the logout URL/view) to close the gap, plus a test or two for the middleware changes.

PR #4675, opened 2026-05-21
Improve recipe print view layout

ManojPentapati:
## Changes Made

* Improved recipe print layout styling
* Updated ingredient and step display formatting
* Adjusted recipe view page styling for better print readability

## Files Updated

* IngredientsTable.vue
* RecipeView.vue
* StepView.vue
* RecipeViewPage.vue

## Testing

* Verified locally using Docker setup
* Checked print layout rendering in browser

### DIFF
--- a/vue3/src/components/display/IngredientsTable.vue
+++ b/vue3/src/components/display/IngredientsTable.vue
@@ -172,4 +172,19 @@ function addToShopping(ingredient: Ingredient) {
 
 <style scoped>
 
+@media print {
+    :deep(.v-table) {
+        font-size: 12px !important;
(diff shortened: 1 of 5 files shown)


paulovitorbal:
Would you mind to share a screenshot of the result of your changes? I am not actively involved on the project yet, but I was also thinking about improving the print design. I looked through your changes and I think they are relevant.

I was thinking on removing the botton row, where we can find the create at, created by, updated at, etc. 
For a printable recipe, those information are not relevant imho. And also the creator is also stated on top of the printable page.

What do you think?

PR #4684, opened 2026-05-31
feat: add to existing pantry entries

SirEdvin:
## Summary

- add a per-row pantry action to add more stock to an existing inventory entry
- prefill inventory booking add mode from the selected pantry entry, including code/barcode
- handle duplicate inventory codes on create by incrementing the existing entry amount instead of surfacing a database conflict
- reject duplicate-code additions when the submitted unit does not match the existing inventory entry unit
- add focused serializer tests for amount increments, inventory log creation, preserving the original entry details, and unit mismatch validation

## Verification

- `python -m pytest cookbook/tests/api/test_api_inventory_entry.py -q`
- `python -m py_compile cookbook/serializer.py cookbook/tests/api/test_api_inventory_entry.py`
- `git diff --check`
- `yarn build`

## Manual smoke notes

I also used a local Tandoor container to inspect the current pantry flow: setup, empty pantry, adding an item through the API, viewing the pantry row, opening remove mode for the row, and removing the item from stock.

### DIFF
--- a/cookbook/serializer.py
+++ b/cookbook/serializer.py
@@ -11,6 +11,7 @@ from django.forms.models import model_to_dict
 from django.contrib.auth.models import AnonymousUser, Group, User
 from django.core.cache import caches
 from django.core.mail import send_mail
+from django.db import IntegrityError, transaction
 from django.db.models import Q, QuerySet, Sum
 from django.http import BadHeaderError
 from django.urls import reverse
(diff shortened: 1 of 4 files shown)


nilsjaekel:
I just discovered pantry and was like "I would like to be able to increase stock, lets check PRs" and yours came up directly :) Hoping for this to be merged.

SirEdvin:
I hope for this too :) Having ability to increase amount of something is stock instead of creating duplicating records seems quite important to me

vabene1111:
yes that makes sense, wanted that as well myself. Not sure when I will get to it but for now, as a hack, you can remove a negative amount which will add it 😉

SirEdvin:
sorry, forgot to mark it ready. I tested it on my local instance and it is working

PR #4687, opened 2026-06-02
fix: use iCalendar version 2.0 for calendar exports

veeresh4520:
The exported iCalendar files currently use the Tandoor application version as the VCALENDAR VERSION field.

Example:

VERSION:2.6.9

According to the iCalendar specification, the VERSION field should be set to `2.0`.

This change updates calendar exports to use the correct iCalendar version, improving compatibility with clients such as Apple Calendar.

Fixes   4687


vabene1111:
thank you very much, i did not know this

PR #4692, opened 2026-06-06
  fix: handle null description in Mealie import

geoClink:
## Bug
  Importing from Mealie fails with `'NoneType' object is not subscriptable` when any recipe has a null description. The entire import crashes and 0 recipes are imported.

  ## Root Cause
  `mealie1.py` line 114 slices `r['description']` with `[:512]` without checking for `None` first. Mealie allows recipes to have no description, so this field can be null in the export.
  
  ## Fix
  Use `(r['description'] or '')[:512]` to convert `None` to an empty string before slicing. This is the same pattern already used elsewhere in the file.


geoClink:
This is ready for review. Simple one-line fix. Now handles null recipe descriptions in Mealie exports by converting None to an empty string before slicing, following the same pattern used elsewhere in the file.

PR #4697, opened 2026-06-12
Add updated_at fields to more models

### LINKED ISSUE #4672: Create updated_at on Food Model

I am implementing a offline / caching synchronization mechanism for the unofficial client app https://github.com/aimok04/kitshn/pull/396. This requires us to work with the paginated API. Some models such as `recipes` provide a updated_at filter in the API which allow us to do delta-syncronization (pull only newly created items). 

However especially with the Food model the API schema says there should be an updated_at filter. But this is a noop since the Food Model does not have a `updated_at` field.

For us it would be practical to also add to Food items the `update_at` field to allow us to do better synchronization without dos'ing the server.

If we cannot filter for newest items and the pageination is not sorted by e.g. id then we can only sync / partial-sync update each time.

### Additional context

_No response_

jonsch318:
This would resolve [#4672](https://github.com/TandoorRecipes/recipes/issues/4672).

From a client perspective like [kitshn](https://github.com/aimok04/kitshn/pull/396) updated_at filters are nice because they allow for better synchronization without much hassle since we can get only the newest updates after the last synchronization.

I started with `Food`, `SupermarketCategory` and `Unit` because they were already using the StandardFilterModelViewSet which allowed updated_at queries but the models where missing the property, which i added. I would be fine with change only but I thought it would be better if most mutable models would allow for updated_at queries and it would be better to do it in one swoop than multiple times later on requiring a migration.

Since many do not derive the StandardFilterModelViewSet i added a new smaller UpdatedAtFilterBackend. I needed to use the BaseFilterBackend with get_schema_operation_parameters so that the query documentation is correct.

I also added tests for the updated_at filter.

There are some models like Logs, Space, FoodProperty, Space, ConnectorConfig but I think these are fine and not really that change frequent.

I tried generating the vue3 openapi client but many many things changed especially .openapi-generator/FILES which some where removed for some reason idk. So I did not want to clutter up this PR. Just say that I should include it if thats wanted.


vabene1111:
thank you, that makes sense for offline clients. I will review it and try to get it merged with one of the next releases. Sorry that things are a bit slow at the moment as I am extremly busy IRL and also have smilerzs big batch of PRs that I promised to look at before breaking them all with changes first.

jonsch318:
no worries take your time.

PR #4698, opened 2026-06-13
docs: clarify default import file formatDocs tandoor import format

### LINKED ISSUE #4323: Add details for Tandoor import
### Documentation link

https://docs.tandoor.dev/features/import_export/#default

### Affected section

FAQ

### Other

_No response_

### Issue description

The current documentation is unclear about the expected data format for importing into Tandoor.

As a new user, I wanted to import data into Tandoor. To understand the required format, I exported an existing recipe as a reference. However, based on the documentation, it was not clear that the import expects a ZIP file rather than a raw .json file, so I attempted to import the exported JSON directly.

This resulted in an error during import, which made it appear as if the file itself was invalid rather than incorrectly formatted for the import process.

The core issue here is documentation clarity: the import section does not clearly state the required file structure and format (for example, that the JSON must be contained within a ZIP archive). As a result, new users can easily misinterpret the export output as being directly importable.

Please consider updating the documentation to clearly describe the expected file structure and contents

veeresh4520:
Clarifies that the Default Tandoor importer expects the exported ZIP archive rather than an extracted JSON file.

## Changes

* Added documentation note to the Default integration section.
* Explained that importing a standalone JSON file is not supported.
* Clarified that users should import the exported ZIP archive directly.

Closes #4323


vabene1111:
thanks

PR #4704, opened 2026-06-16
feat: add user setting to show ingredient notes inline

### LINKED ISSUE #455: Ingredient Notes Rework

guguinhaaax:
Fixes #455

Previously, ingredient notes were hidden behind a tooltip icon requiring a hover or tap to view. This made notes hard to discover and inconvenient to use, especially on mobile devices or while cooking.

This PR adds a new user preference `show_ingredient_notes` that allows users to choose whether ingredient notes are displayed inline (below the ingredient) or as a tooltip (previous behavior).

Changes:
- cookbook/models.py: added `show_ingredient_notes` BooleanField (default=True)
- cookbook/migrations/0243_userpreference_show_ingredient_notes.py: migration for new field
- cookbook/serializer.py: exposed `show_ingredient_notes` in UserPreferenceSerializer
- vue3/src/components/display/IngredientsTable.vue: render note inline when setting is enabled, tooltip otherwise
- vue3/src/components/settings/CosmeticSettings.vue: added checkbox to toggle the setting
- vue3/src/locales/en.json: added translation strings for the new setting


tejash5489-lang:
Flagging a serious issue before this goes further: the migration renames an existing, unrelated field rather than adding a new one.

```python
migrations.RenameField(
    model_name='userpreference',
    old_name='show_step_ingredients',
    new_name='show_ingredient_notes',
),
```

`show_step_ingredients` is a separate, currently-shipping preference ("show ingredients table next to recipe steps") — it's referenced in `models.py`, the `UserPreferenceSerializer` fields tuple, `CosmeticSettings.vue`'s existing checkbox, and read directly in ~25 recipe-import integrations (`cookbook/integration/paprika.py`, `mealie.py`, `chowdown.py`, `cookmate.py`, etc.) plus `recipe_url_import.py`.

Renaming the underlying column means:
1. Every existing user's saved `show_step_ingredients` value gets silently reinterpreted as their new `show_ingredient_notes` preference — an unrelated setting with no connection to what they actually chose.
2. The model no longer has a `show_step_ingredients` attribute, but ~25 integration files and `recipe_url_import.py` still do `request.user.userpreference.show_step_ingredients` — that's an `AttributeError` on every recipe import (URL import and all third-party format importers).
3. `serializer.py`'s `UserPreferenceSerializer.Meta.fields` still lists `'show_step_ingredients'` literally, alongside the new `'show_ingredient_notes'` — since that's no longer a valid model field name, DRF should raise at serializer introspection time, which would affect the `/api/user-preference/` endpoint broadly.
4. The existing "Show Step Ingredients" checkbox in `CosmeticSettings.vue` is left wired up to a backend field that's now gone.

This has the shape of a `makemigrations` rename-detection prompt ("did you rename `show_step_ingredients` to `show_ingredient_notes`?") answered yes by mistake, rather than an intentional design — the two settings are semantically unrelated (one is per-recipe-step display, the other is ingredient note visibility) and happened to collide only because both are same-shaped booleans. The fix should be a genuinely new field/migration (`AddField`) rather than a rename, so `show_step_ingredients` stays intact.

(Based on static analysis of the diff + a full-branch grep for remaining `show_step_ingredients` references, not a live run against a DB — but the AttributeError/DRF-misconfiguration consequences follow directly from Python/Django's documented attribute-resolution behavior once the model field is gone.)

PR #4718, opened 2026-06-23
feat: add shopping list item notes

### LINKED ISSUE #4624: Add notes to shopping list (items)

While planning the shopping list for a large group, we missed the option to leave notes on items in the shopping list for the ones, who are going to the supermarket. While shopping list items (based on recipes) are very generic, like "jam", "Gouda-Cheese" e.g. we have difficulties to add notes to an item e.g. to provide information about a certain brand to buy or some allergies to check before buying a certain product.

I would like to have the opportunity to leave an additional note on each shopping item to give further instructions for the ones buying it. It is important that such notes in our particular context are limited to the actual shopping list and it not something that is general applicable for the groceries.

I could replace a shopping item with a more precise one, but afaik, I would loose the reference to the underlying recipe/planned meal which is really helpful while double checking the shopping list or while shopping itself.

### Additional context

I would like to discuss this feature with you before starting to make an upstream contribution.

jonsch318:
This adds a note field to the shopping list entry and adds a new textfield to the ShoppingLineItemDialog to configure it per shopping item which then shows in the shopping line next to the food name. I tried it below in the info Row but this would conflict with the tons of  information there already. 
UI feedback is appreciated.

It currently does not connect to the existing note structure of ingredients since I don't know if it wanted or not.

resolves https://github.com/TandoorRecipes/recipes/issues/4624

image

image


vabene1111:
hi, thank you. The dialog is already very overloaded but I also see no other option to add any info there. Will review and see if I merge it. The info row should probably at least include an icon that there is a note on the given row 🤔 other than that thats probly how I would have done it as well. 

Maybe just dont display the text input always but make a small button that you can press so it appears when needed 🤔 

will need to think about this.

jonsch318:
I will create different designs soon if I find time for it

PR #4736, opened 2026-07-01
fix markdown blockquote parsing

tomtjes:
_bleach.clean()_ turns `>` into `&gt;` so markdown parsing it into `<blockquote>` fails. As a fix, I swapped md parsing and sanitizing.

### DIFF
--- a/cookbook/helper/template_helper.py
+++ b/cookbook/helper/template_helper.py
@@ -123,9 +123,6 @@ def render_instructions(step):  # TODO deduplicate markdown cleanup code
         "a": ["href", "alt", "title"],
     }
 
-    # do a first, strict round of cleaning
-    instructions = bleach.clean(instructions, allowed_tags, allowed_attributes)
-
     # parse markdown
     instructions = md.markdown(
         instructions,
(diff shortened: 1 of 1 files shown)


vabene1111:
thank you. It seems like an easy fix but I remember doing this two stage proces for some security reason. I will need to find the old markdown test code and see if it still fails/is safe with this way of parsing. Still blockquotes are nice so I would like to get them working again.

tomtjes:
Yeah, I assumed that was the reason. But since the markdown parser should only add some _legal_ html tags into the code, I could not think of why this order should be unsafe. But I'm no expert. The alternative would be to add a manual parser just for blockquote after the bleach, replacing `&gt; ` at the beginning of a paragraph and wrapping that paragraph in the blockquote tag.

vabene1111:
ok the reason for the double check is in this advisory (I think the comments below are private) https://github.com/TandoorRecipes/recipes/security/advisories/GHSA-89pw-5qxc-7v86 

Need to review why exactly I did the first round of cleaning as I sadly did not document it well enough.

vabene1111:
ok reviewed and tested this again. It seems the first run might not be required after all as non of the malicous payloads I tested back then work. Not merging this as I had to comment out the line for testing but thanks for bringing this up and pointing me to the source, will mention this in the changelog.

PR #4737, opened 2026-07-01
feat: Add richer reverse proxy authentication support

Jbansal2:
## Description
Implements richer reverse proxy authentication to address #1134

This PR adds support for extracting user information from additional reverse proxy headers and enables custom SSO logout URLs.

### New Features
- Support for `HTTP_REMOTE_NAME` header to auto-populate user's first/last name
- Support for `HTTP_REMOTE_EMAIL` header to auto-populate user's email address
- User information automatically updates on each login
- New `REVERSE_PROXY_AUTH_LOGOUT` environment variable for custom SSO logout URLs

### Technical Implementation
- Created `CustomRemoteUserBackend` that extends Django's `RemoteUserBackend`
- Added conditional URL override for logout when using reverse proxy auth
- Maintains backward compatibility - existing setups continue to work

### Documentation
- Updated authentication documentation with examples
- Added nginx configuration samples for Authelia/Authentik
- Added configuration reference for new environment variable

### Testing
- Comprehensive test suite with 13 test cases
- Tests cover name parsing, email updates, edge cases, and error handling

## Usage Example

**.env:**
```ini
REMOTE_USER_AUTH=1
REVERSE_PROXY_AUTH_LOGOUT=https://authelia.example.com/logout

### DIFF
--- a/IMPLEMENTATION_NOTES.md
+++ b/IMPLEMENTATION_NOTES.md
@@ -1,0 +1,160 @@
+# Richer Reverse Proxy Authentication - Implementation Notes
+
+## Overview
+This implementation adds support for richer reverse proxy authentication in Tandoor Recipes, addressing GitHub issue #1134.
+
+## Changes Made
+
+### 1. New Backend: `cookbook/helper/remote_user_helper.py`
+Created a custom `CustomRemoteUserBackend` that extends Django's `RemoteUserBackend` to:
+- Extract and update user email from `HTTP_REMOTE_EMAIL` header
+- Extract and update user first/last name from `HTTP_REMOTE_NAME` header
+- Automatically update user information on each login
+- Only save user object if changes were detected
+
+**Key Features:**
+- Handles names with various formats (single name, multiple spaces, etc.)
+- Strips whitespace from names
+- Splits name into first and last name on first space
+- Updates only when values differ from existing data
+
+### 2. Settings Changes: `recipes/settings.py`
+- Added `REVERSE_PROXY_AUTH_LOGOUT` environment variable support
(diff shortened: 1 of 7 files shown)


tejash5489-lang:
Compared this against #4644 while triaging #1134 — this is a solid, more complete implementation:

- Using a custom `RemoteUserBackend` subclass (`CustomRemoteUserBackend`) instead of hooking into the middleware is the more idiomatic Django extension point for this and keeps `middleware.py` untouched.
- The logout override actually works end-to-end: I traced the `urlpatterns.insert(1, ...)` against Django's URL-resolution order and confirmed the `RedirectView` for `accounts/logout/` is registered ahead of allauth's own logout route (which lives in the `include('allauth.urls')` at what becomes index 2), so it correctly intercepts the request rather than just shadowing the URL name.
- Good test coverage (13 cases) and the docs update to the Authelia nginx example is a nice touch — it directly reflects the workaround the original issue reporter described (the old `location /accounts/logout/ { return 301 ... }` nginx hack), now replaced by the env var.

One thing worth cleaning up before merge: `IMPLEMENTATION_NOTES.md` at the repo root looks like a leftover scratch/notes file rather than something meant to ship — probably worth dropping from the PR.

Minor, non-blocking note on behavior: since this relies on Django's `RemoteUserMiddleware`, the backend's `authenticate()` (and thus the header sync) only runs when a new session is established, not on every request — so a header value changing mid-session won't take effect until the session is renewed. That matches the issue's wording ("update when a user logs in") and Django's own recommended pattern, just flagging it as a known tradeoff vs. #4644's per-request sync.

Issue #4713, opened 2026-06-20
Open Data Import fails when 'Property' is unselected


stleusc:
### Tandoor Version

2.6.9

### Setup

Docker / Docker-Compose

### Reverse Proxy

No reverse proxy

### Bug description

Brand new install. As part of onboarding it offers Open Data Import.
I HATE the properties and do not care for them, so I uncheck them. 
If I do that, import fails with below error.
Once I select it, I can import the data. But them In need to MANUALLY delete the properties from 400 foods.
That makes no sense.
Please allow import without Property

### Relevant logs

```shell
ThreadPoolExecutor-0_0 ERROR 2026-06-20 11:05:24,998 django.request Internal Server Error: /api/import-open-data/

Traceback (most recent call last):

  File "/opt/recipes/venv/lib/python3.13/site-packages/django/core/handlers/exception.py", line 55, in inner

    response = get_response(request)

  File "/opt/recipes/venv/lib/python3.13/site-packages/django/core/handlers/base.py", line 197, in _get_response

    response = wrapped_callback(request, *callback_args, **callback_kwargs)

  File "/opt/recipes/venv/lib/python3.13/site-packages/django/views/decorators/csrf.py", line 65, in _view_wrapper

    return view_func(request, *args, **kwargs)

  File "/opt/recipes/venv/lib/python3.13/site-packages/django/views/generic/base.py", line 105, in view

    return self.dispatch(request, *args, **kwargs)

           ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/opt/recipes/venv/lib/python3.13/site-packages/rest_framework/views.py", line 515, in dispatch

    response = self.handle_exception(exc)

  File "/opt/recipes/venv/lib/python3.13/site-packages/rest_framework/views.py", line 475, in handle_exception

    self.raise_uncaught_exception(exc)

    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^

  File "/opt/recipes/venv/lib/python3.13/site-packages/rest_framework/views.py", line 486, in raise_uncaught_exception

    raise exc

  File "/opt/recipes/venv/lib/python3.13/site-packages/rest_framework/views.py", line 512, in dispatch

    response = handler(request, *args, **kwargs)

  File "/opt/recipes/cookbook/views/api.py", line 3175, in post

    response_obj['food'] = data_importer.import_food().to_dict()

[...]

PR #4743, opened 2026-07-11
fix: skip unimported property types during open data food import (#4713)

TowyTowy:
## Problem
Open Data Import returns a 500 (`KeyError`, e.g. `'property-fats'`) whenever a user imports foods but unselects the "Property" datatype. Reported in #4713.

## Cause
`ImportOpenData.post` runs each importer conditionally on `selected_datatypes`. When "Property" is unselected, `import_property()` never runs, so `slug_id_cache['property']` is empty. `import_food()` still runs and unconditionally iterates each food's `properties['type_values']`, looking up `slug_id_cache['property'][fp['property_type']]` — which raises `KeyError` and aborts the whole import.

## Fix
Guard that lookup: skip any property whose type was not imported, so the food import completes successfully and simply doesn't link that property. This matches the user's intent (import foods without properties).

## Tests
Adds `TestIssue4713_PropertyUnselected` covering both the unselected case (no `KeyError`, food still created) and the selected case (properties still linked, guarding against over-skipping). Full `test_open_data_importer.py` passes (10 tests); the new no-KeyError test fails before the change and passes after.

## Notes
This change was prepared with AI assistance (Claude) and reviewed before submission; the test suite was run locally.

### DIFF
--- a/cookbook/helper/open_data_importer.py
+++ b/cookbook/helper/open_data_importer.py
@@ -416,6 +416,11 @@ class OpenDataImporter:
 
         for k in list(self.data['food'].keys()):
             for fp in self.data['food'][k]['properties']['type_values']:
+                # property types are only imported (and cached) when the 'property' datatype
+                # is selected. If it was not imported, skip linking the property so the food
+                # import still succeeds instead of raising a KeyError. (#4713)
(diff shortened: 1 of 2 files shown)


vabene1111:
thank you, looks good on first sight. will likeley merge

PR #4748, opened 2026-07-15
feat: add per-100g column to recipe nutrition label

JvandeLocht:
To be upfront, I have done most of this with GLM 5.2, but I tried my best to strip the final version of all reimplementations of existing helpers and such.

I have tested this on my live proxmox lxc install of tandoor and found no side effects, during adding or editing of new and or existing recipes. Furthermore I consumed the API with SparkyFitness and could not find any difference in its behaviour.

I have read and signed the CLA.

Add a "per 100g" column to the property view alongside the existing "per serving" and "total" columns.

- FoodPropertyHelper: compute total_weight_grams from ingredient weight conversions during calculate_recipe_properties()
- RecipeSerializer: add food_weight read-only field exposing the computed weight (skipped on create, like food_properties)
- PropertyView.vue: add per_100g column with serving weight subtitle
- API client (Recipe.ts, PatchedRecipe.ts): add foodWeight field

grafik

### DIFF
--- a/cookbook/helper/property_helper.py
+++ b/cookbook/helper/property_helper.py
@@ -1,7 +1,7 @@
 from django.core.cache import caches
 
 from cookbook.helper.cache_helper import CacheHelper
-from cookbook.helper.unit_conversion_helper import UnitConversionHelper
+from cookbook.helper.unit_conversion_helper import BASE_UNITS_WEIGHT, UnitConversionHelper
 from cookbook.models import PropertyType
 
 
@@ -23,6 +23,7 @@ class FoodPropertyHelper:
         """
         ingredients = []
         computed_properties = {}
+        self.total_weight_grams = 0
 
         for s in recipe.steps.all():
             ingredients += s.ingredients.all()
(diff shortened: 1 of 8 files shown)


KPKluge:
Thanks for implementing nutrition values per 100 g. This is a very useful addition.

Would it be possible to support an optional manually entered recipe yield weight as well?

The sum of the ingredient weights is often different from the actual weight of the finished recipe. For example:

* water evaporates during cooking or baking
* rice, pasta, or legumes absorb water
* ingredients may be drained
* bones, shells, cooking liquid, or other parts may be removed

Example:

* calculated ingredient weight: 1,800 g
* actual finished recipe yield: 1,500 g

In this case, the nutritional values per 100 g should be calculated using the finished yield of 1,500 g rather than the original ingredient weight.

A possible implementation could be:

1. Automatically calculate the ingredient weight as proposed in this PR.
2. Add an optional editable field such as “Finished yield weight” or “Recipe yield weight”.
3. Use the manually entered value for the per-100-g calculation when present.
4. Fall back to the automatically calculated ingredient weight when no manual value is set.
5. Scale the manually entered yield proportionally when the recipe is scaled.

This would make the feature accurate for cooked meals, bread, cakes, sauces, soups, and similar recipes while preserving the convenient automatic calculation as the default.

JvandeLocht:
I thought a bit about your proposal and I dont think this is the right approach.

The nutrition values are basically calculated by a weighted sum of the respective foods. 
If we would later scale the whole recipe to compensate for the weight difference of a specific food item, after evaporation or absoprtion of fluids, we would also scale the food items that didnt loose any weight.

I think to accomplish your goal we would need to set the yield weight of each food item individually, but I dont think that fits in the scope of this PR.

vabene1111:
this was discussed a few times before. It sadly does not make any sense to have recipe nuttions per weight as you have already mentioned that the final weight of a recipe is not the sum of its ingredients (at least most times). One would have to weigh the output of cooking to get the ration and that is to cumbersome to be useful. 
For that reason I will not implement/merge this.

PR #4749, opened 2026-07-17
Rework AI step sort around verifiable decisions

tecosaur:
In fa7cc12b992623002a7ba460d18322b50596e784 LLM-automated assignment of ingredients to steps, and step splitting was introduced.

Having just started using Tandoor, I rather liked the idea of this feature, and thought the local LLM I have (Gemma 4 e4b) should be fine for it. Then two minutes and ~6k tokens later I found it timed out, and I was confused by why so much work/output was required. From a read of the prompting design, I think there's substantial room for improvement. Specifically:
- The entire recipe object is sent to the model, despite most fields being irrelevant to the task
- The current model relies on an LLM following a series of "don't do X" instructions ("You cannot under any circumstance change the value of any of the fields." and "Do not change any field value like strings or numbers, or change the sorting, also do not change the language.")
- The LLM is asked to reproduce lots of unmodified information

This relies on big, well-behaved, and somewhat speedy models to work.

From my experience with LLMs, I do think it's much better to:
- Give the LLMs just the information they need to do the task
- Ask for the simplest acceptable form of the output
- Design the output format such that it can be validated/checked
- Allow for easy detection + fixing of common mistakes

With the help of an LLM to do the gruntwork, I've overhauled how the step splitting works to be much simpler for the LLM, turned the "don't do X" instructions into impossibilities, and **reduced the number of tokens needed by ~10-20x**.

1. Steps and ingredients are given IDs
2. We specifically ask which ingredients belong to which steps
3. We validate that the IDs are correct and that all ingredients have been assigned
4. If not all ingredients have been assigned, we emit 1-2 follow-up calls naming the unassigned ingredients. This should be fairly cheap due to prompt caching.
5. We give the LLM the opportunity to split overly large steps, by assigning lines of split steps to new sub-steps + ingredients (also validated)

##### Sample output

```json
{
    "assignments": [
        {"step": "s1_e72", "ingredients":["chicken_6f2", "oil_a31"]},
        {"step": "s2_1a9", "ingredients": ["lemon_74f", "garlic_b3c", "olive_xd1"]},
        {"step": "s3_4b5", "ingredients": ["salt_9c2", "pasta_8d3"]}
    ],
    "should_split": ["s2_1a9"]
}
```

### DIFF
--- a/cookbook/helper/ai_step_sort.py
+++ b/cookbook/helper/ai_step_sort.py
@@ -1,0 +1,416 @@
+"""AI ingredient-to-step sorting via a compact, identifier-based protocol.
+
+The model returns decisions (ingredient→step assignments, steps to split) keyed by
+request-local ids; Tandoor validates them against the original objects and applies them,
+so the model never reproduces recipe data. Fabricated ids are fatal; an incomplete
+assignment is recovered by re-prompting for the missing ingredients and finally by leaving
+any still-unassigned ingredient in its original step. Splitting is a per-step-tolerant
+enhancement. All functions take serializer-shaped recipe dicts and do no I/O.
+"""
+
+import json
+import re
+import secrets
+import string
+
+from django.utils.text import slugify
+
+ASSIGNMENT_INSTRUCTION = (
+    "You are given a recipe as JSON: recipe context, a list of ingredients (each with an id), "
+    "and a list of steps (each with an id). Assign each ingredient to "
+    "exactly one step: the first step where it is used, choosing the most plausible step from the "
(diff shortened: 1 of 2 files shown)


vabene1111:
thank you. I implemented the AI features relatively fast to test them and was not able to optimize since. i will review your code and get back to you.

tecosaur:
I just want to mention that I've been using this for the month with Gemma 4 e4b and now 12b and I'm pretty happy with how well it's working in practice.

vabene1111:
thank you, that is good to hear

PR #4752, opened 2026-07-19
Fix space assignment during social signup

tecosaur:
I ran into an annoying bug while trying to add other people to Tandoor, which I think is best described by the commit message (reproduced below) 

-----

With `SOCIALACCOUNT_AUTO_SIGNUP` enabled, django-allauth bypasses the social signup form and creates the user directly via the social account adapter, so `AllAuthSocialSignupForm.signup()` never ran and `SOCIAL_DEFAULT_ACCESS` users received no UserSpace. They were then prompted to create their own space instead of joining the shared one.

Move the default-space assignment to `TandoorSocialAccountAdapter.save_user()`, the common hook for both automatic (`form=None`) and form-based signup. The form's `signup()` override is removed entirely; allauth's base form already provides a no-op, and dropping it avoids assigning the space twice.

### DIFF
--- a/cookbook/forms.py
+++ b/cookbook/forms.py
@@ -4,7 +4,6 @@ from allauth.account.forms import ResetPasswordForm, SignupForm
 from allauth.socialaccount.forms import SignupForm as SocialSignupForm
 from django import forms
 from django.conf import settings
-from django.contrib.auth.models import Group
 from django.core.exceptions import ValidationError
 from django.forms import widgets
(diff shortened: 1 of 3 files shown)


vabene1111:
thank you. have you tested this to work with proxy, third party auth and just normal user/password auth?

tecosaur:
I've only tested it with OIDC, since the change just affects the `AllAuthSocialSignupForm` and `TandoorSocialAccountAdapter` classes.

PR #4769, opened 2026-08-03
Apply EXIF orientation before resizing images

### LINKED ISSUE #4765: Image rotation not taken into account when downsampling recipe image
### Tandoor Version

2.6.13

### Setup

Docker / Docker-Compose

### Reverse Proxy

Traefik

### Other

_No response_

### Bug description

When attaching this image to any recipe, it gets rotated 90 degrees to the left. But not for the small preview.

The metadata contain the Orientation tag which seems to be honored by the preview, but not by the downsampling operation.

```
$ exiftool 20260801_175525.jpg | grep Orientation
Orientation                     : Rotate 90 CW
```

Preview:

Image

Final:

Image

[20260801_175525.jpg.zip](https://github.com/user-attachments/files/30621041/20260801_175525.jpg.zip)

### Relevant logs

```shell

```

agross:
## What changed

- Apply EXIF orientation before resizing JPEG, PNG, and WebP recipe images.
- Apply EXIF orientation before stripping image metadata from images that do not need resizing.
- Add regression tests for both JPEG processing paths.

## Why

Image pixels were resized or copied without first applying the EXIF Orientation tag. The later save removed that metadata, leaving the final image permanently rotated even though browser previews honored the tag.

## Validation

- Focused pytest run in the repository dev container: 2 passed.
- Flake8 run for the changed Python files in the repository dev container: passed.
- Verified the attached reproduction image (Orientation 6, 4000x3000) produces an upright 1020x1360 result with the orientation metadata removed.

Fixes #4765

### DIFF
--- a/cookbook/helper/image_processing.py
+++ b/cookbook/helper/image_processing.py
@@ -1,11 +1,11 @@
 import os
 from io import BytesIO
 
-from PIL import Image
+from PIL import Image, ImageOps
 
 
 def rescale_image_jpeg(image_object, base_width=1020):
-    img = Image.open(image_object)
+    img = ImageOps.exif_transpose(Image.open(image_object))
     icc_profile = img.info.get('icc_profile')  # remember color profile to not mess up colors
     width_percent = (base_width / float(img.size[0]))
     height = int((float(img.size[1]) * float(width_percent)))
@@ -18,7 +18,7 @@ def rescale_image_jpeg(image_object, base_width=1020):
 
 
 def rescale_image_png(image_object, base_width=1020):
-    image_object = Image.open(image_object)
+    image_object = ImageOps.exif_transpose(Image.open(image_object))
     wpercent = (base_width / float(image_object.size[0]))
     hsize = int((float(image_object.size[1]) * float(wpercent)))
     img = image_object.resize((base_width, hsize), Image.LANCZOS)
@@ -29,7 +29,7 @@ def rescale_image_png(image_object, base_width=1020):
 
 
 def rescale_image_webp(image_object, base_width=1020):
-    image_object = Image.open(image_object)
+    image_object = ImageOps.exif_transpose(Image.open(image_object))
     wpercent = (base_width / float(image_object.size[0]))
     hsize = int((float(image_object.size[1]) * float(wpercent)))
     img = image_object.resize((base_width, hsize), Image.LANCZOS)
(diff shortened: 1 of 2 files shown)


agross:
I'm not a Python dev, so I let Codex take a look at it.

vabene1111:
thank you. Since this was done with AI: have you read the documentation of the command to verify it makes sense and tested this change with some different images on a test instance?

agross:
No and no.

agross:
> If an image has an EXIF Orientation tag, other than 1, transpose the image accordingly, and remove the orientation data.

[Seems](https://pillow.readthedocs.io/en/stable/reference/ImageOps.html#PIL.ImageOps.exif_transpose) legit, though.

agross:
@vabene1111 How can we move this forward?

tejash5489-lang:
I independently verified this while triaging good-first-issue PRs, since @vabene1111's review question ("have you tested this with different images on a test instance?") is exactly the kind of thing I can check.

I extracted `image_processing.py` from `develop` and this branch into an isolated harness (no Django/DB dependency — the module only imports PIL/os/io) and built synthetic images with a positional marker, testing all four distinct EXIF orientation classes (1=normal, 3=180°, 6=90°CW, 8=90°CCW) across JPEG/PNG/WebP, plus the `strip_image_meta` no-resize path. Ground truth for each case was computed independently via Pillow's own `ImageOps.exif_transpose` (not by trusting either branch's code).

Results:
- **`develop` (baseline)**: ignores EXIF orientation entirely in every case — output dimensions and marker position are unchanged regardless of the orientation tag, reproducing the bug from #4765 exactly.
- **This PR**: matches Pillow's reference `exif_transpose` output in all 6 test combinations (dimensions swap correctly, marker lands in the geometrically correct corner, EXIF orientation tag is stripped afterward). Also confirmed the width/height-percent scaling in `rescale_image_jpeg` is computed *after* the transpose, so a 4000x3000 orientation-6 source correctly becomes a portrait-scaled output rather than a landscape one scaled from the wrong dimensions.

This is the documented, canonical way to handle the problem (`ImageOps.exif_transpose`, per the Pillow docs linked upthread), applied consistently across all four image-processing entry points. From a correctness standpoint this looks solid and ready to move forward.

vabene1111:
great, this sounds like even more AI ... sorry but these days maintianing AI PRs is more work that doing stuff yourself .... will get this looked over at some point

agross:
As I said, I'm not a Python developer, AI helped finding the right APIs. It fixes the issue for me. From my perspective, the change is not too complex and should be low risk for you.

vabene1111:
thank you, reviewed the docs and looks good.

PR #4779, opened 2026-08-06
Fix JPEG EXIF orientation during image processing

DiwanshuG:
## Summary

This PR fixes JPEG images being displayed with incorrect orientation after upload by applying the EXIF orientation before resizing.

### Changes

- Import `ImageOps` from Pillow.
- Apply `ImageOps.exif_transpose()` immediately after opening JPEG images in `rescale_image_jpeg()`.

### Testing

- Reproduced the issue reported in #4765.
- Verified that the affected JPEG image is displayed with the correct orientation after upload.
- Verified that a smaller JPEG image also displays correctly.

### DIFF
--- a/cookbook/helper/image_processing.py
+++ b/cookbook/helper/image_processing.py
@@ -1,12 +1,13 @@
 import os
 from io import BytesIO
 
-from PIL import Image
+from PIL import Image, ImageOps
 
 
 def rescale_image_jpeg(image_object, base_width=1020):
     img = Image.open(image_object)
-    icc_profile = img.info.get('icc_profile')  # remember color profile to not mess up colors
+    img = ImageOps.exif_transpose(img)
+    icc_profile = img.info.get('icc_profile')
     width_percent = (base_width / float(img.size[0]))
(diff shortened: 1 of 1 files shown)


agross:
Duplicate of #4769

agross:
Also, this one only handles JPEGs.

DiwanshuG:
Thanks for the review!

You're right. My current fix only handles JPEG images by applying `ImageOps.exif_transpose()` in the JPEG processing path.

I'll take another look at the image processing flow and update the PR if the orientation handling should also be applied in the other relevant code paths.

agross:
@DiwanshuG Could you share why you're eager to duplicate the work?

DiwanshuG:
Hi! I wasn't aware that another PR or implementation already addressed the same approach. I picked up issue #4765, reproduced it locally, investigated the image processing flow, and arrived at this solution independently. I wasn't intentionally trying to duplicate anyone's work. Thanks for pointing me to #4769.

vabene1111:
so closing this in favor of the other one, thanks for everyones work

PR #4785, opened 2026-08-12
Render recipe and step times as hours and minutes

### LINKED ISSUE #382: time display option

dr-froehlich:
Closes #382.

## The problem

Recipe and step durations are stored and displayed as a flat number of minutes. A 72 hour
sourdough proof renders as `4320 min`, the step timer button shows a bare `4320`, and the
recipe card's time chip shows working + waiting summed with no unit at all.

This was agreed in #382 in 2021 — "everything over 60 minutes is converted to hours and
minutes" — and a `timeConvert()` helper was sketched in the thread, but it was never built.
Worth flagging for anyone who lands here from a search: #3751 and #3778 are linked under
that issue and look like the fix. They are not. Together they add a localized `+2 Days`
suffix to `Timer.vue`'s "Finished at" line while a countdown is *running* — correct, and
unrelated to how a recipe page renders its durations.

## What this does

A shared formatter in `vue3/src/utils/duration_utils.ts` converts minutes for display:

| Input | Output |
|---|---|
| 45 | `45 min` |
| 59 | `59 min` |
| 60 | `1 h` |
| 90 | `1 h 30 min` |
| 120 | `2 h` |
| 4320 | `72 h` |

Below an hour it stays minutes; from an hour up it renders `H h`, plus ` M min` when there
is a remainder. **No day rollover** — 4320 stays `72 h`, matching the table in #382.
`Timer.vue` already expresses "+3 Days" for the one case where a calendar day matters, and
long proofs are spoken as "72 hours", not "3 days".

Unit labels are passed in by the caller from translation keys, so no unit string is inlined
in the conversion logic. `min` already existed; `h` is new in `en.json`.

Applied at the read-only duration displays: both `RecipeView` layouts (working and waiting
time), the `RecipeCard` time chip, and the `StepView` timer button. Duration *inputs* keep
plain minutes — entering durations was never the problem, reading them is.

It also removes two hardcoded `min` strings in `RecipeView.vue` that bypassed i18n. They sat
on the lines this change rewrites.

## The preference

`UserPreference.use_readable_time`, mirroring `use_fractions` field for field: one boolean,
one serializer entry, one checkbox in Cosmetic settings, read at the call sites from the
user preference store. One migration, and the generated `UserPreference` /
`PatchedUserPreference` client models regenerated to match.

**It defaults to `false`**, so no existing user's recipe pages change on upgrade. If you
would rather it default on, that is a one-line change in `models.py` plus the migration's
`AddField` — happy to flip it.

## The second commit — vitest

The contribution guidelines mention you are soliciting contributors to set up vue3 testing,
so this sets it up. The formatter is a pure function of one number and a boolean, which
makes it a natural first suite: the whole format table, the preference-off path, and
zero / null / undefined / negative / non-finite input. 20 tests, run with
`yarn --cwd vue3 test`. Vitest is Vite-native so configuration is near-zero, and it is a
devDependency, so it cannot reach the shipped bundle.

That is deliberately the **last** commit and touches only `package.json`, `vitest.config.ts`,
`yarn.lock` and the spec file. If you would rather review test tooling on its own, drop it
with a single reset and I will open it as a separate PR.

## Notes

- The Python tests go into the existing `cookbook/tests/api/test_api_userpreference.py`
  rather than a new module.
- `yarn.lock` was regenerated on top of this branch's base rather than transplanted, so it
  carries your dependency versions, not mine.
- flake8 is clean on the changed files (the five findings in `serializer.py` are
  pre-existing on `develop`). I left the surrounding yapf reformatting alone rather than
  dragging unrelated churn into the diff — happy to run it if you would prefer.
- Translations: English keys only, per `docs/contribute/translations.md`. Weblate can take
  it from there.


vabene1111:
hi, thanks for the PR. While I like the feature I think its a lot more complex than it need to be. I do not think we need the user preference. Just show the times in a human readable format. 

Alternatively one could change the field to a proper duration field, that would likely make a lot more sense but require a migration of existing data. 

Closing this for now, feel free to implement in the recommended way or I will look into this at some point.

dr-froehlich:
Thanks for the quick look, and fair enough — you were right that the preference was the
complicated part.

Redone the way you suggested in #4810: no preference, times are just shown in a
human readable format. It drops the model field, the migration, the serializer entry, the
settings checkbox and the regenerated client models, so the new one touches no Python at all —
ten files, all under `vue3/`.

I have left the duration-field idea alone rather than guessing at it. It breaks the API for
existing clients and needs the data migration you mentioned, so it seemed like something you
would want to shape yourself. If you would rather have that instead, this one is easy to close.

PR #4809, opened 2026-09-07
fix: send active yes/no search filters

### LINKED ISSUE #4788: Yes/No search filters are never applied - "Currently On Hand" (makenow) has no effect
## Description

Search filters that are rendered as a Yes/No `VSelect` never reach the API. Selecting **Yes**
on the *Currently On Hand* filter returns the complete, unfiltered recipe list.

The backend is fine — `makenow` works correctly when it actually arrives. The parameter is
dropped in the frontend, in `isFilterDefaultValue()`.

## Steps to reproduce

1. Put at least one recipe in a state where it is *not* cookable (an ingredient neither in
   inventory nor marked on hand).
2. Go to **Search** → *Add filter* → **Currently On Hand** → set it to **Yes** → **Search**.
3. All recipes are returned, including ones with missing ingredients.
4. Open `/advanced-search?makenow=true` directly — same result, no filtering.
5. Open `/advanced-search?makenow=0` — filtering works as expected.

## Expected

`makenow=true` filters to recipes whose ingredients are all available (equivalent to
`makenow=0`).

## Actual

The `makenow` parameter is never sent. Verified against the API directly with a
session-authenticated client on the same instance:

| request | recipes returned |
|---|---|
| `/api/recipe/` | 162 |
| `/api/recipe/?makenow=true` | 0 |
| `/api/recipe/?makenow=0` | 0 |
| `/api/recipe/?makenow=1` | 2 |
| `/api/recipe/?makenow=false` | 162 |

So the backend behaves correctly. The browser simply omits the parameter.

## Cause

`vue3/src/pages/SearchPage.vue`:

    function isFilterDefaultValue(filter: any) {
        if (Array.isArray(filter.default) && Array.isArray(filter.modelValue)) {
            return filter.default.length == filter.modelValue.length
        } else if (isNaN(filter.default) && isNaN(filter.modelValue)) {
            return true                       // <-- both non-numeric => reported as "default"
        } else {
            return toRaw(filter.default) === filter.modelValue
        }
    }

and the filter definition:

    makenow: {
        id: 'makenow',
        label: t('OnHand'),
        default: "false",
        is: VSelect,
        items: [{value: "true", title: 'Yes'}, {value: "false", title: 'No'}],
        modelValue: useRouteQuery('makenow', "false"),
    }

`default` is the string `"false"` and the selected value is the string `"true"`.
`isNaN("false")` and `isNaN("true")` are both `true`, so the second branch returns `true`
— the filter is reported as being at its default value.

`searchRecipes()` only forwards filters that are *not* at their default:

    Object.values(filters.value).forEach((filter) => {
        if (!isFilterDefaultValue(filter)) {
            searchParameters[filter.id] = filter.modelValue
            hasFiltersApplied.value = true
        }
    })

so `makenow` is dropped. `enableFiltersWithValues()` uses the same helper, which is why
loading a URL with `?makenow=true` does not even mark the filter as enabled in the UI.

`?makenow=1` works because `isNaN("1")` is `false`, so the comparison branch is reached
(`"false" === "1"` → not default) and the parameter is forwarded.

This affects **any** Yes/No filter whose values are non-numeric strings, not just `makenow`.

## Workaround

Use a numeric value in the URL — `?makenow=0` is equivalent to the broken *Yes* option:

    /advanced-search?makenow=0     all ingredients available
    /advanced-search?makenow=1     at most one ingredient missing
    /advanced-search?makenow=2     at most two missing

This is also the only way to reach the "at most N missing" behaviour: `make_now_count`
("At most missing ingredients") exists in the locale files but there is no input field
bound to it, and the backend takes the count in `makenow` itself
(`_parse_makenow` in `cookbook/helper/recipe_search.py` maps `true`/`yes` to `0`).

## Suggested fix

Compare against the default without the `isNaN` shortcut, e.g.

    function isFilterDefaultValue(filter: any) {
        if (Array.isArray(filter.default) && Array.isArray(filter.modelValue)) {
            return filter.default.length == filter.modelValue.length
        }
        return toRaw(filter.default) === filter.modelValue
    }

(The `isNaN` branch appears to be aimed at `undefined`/`null` defaults; those are already
handled by the strict comparison.)

## Version

- Tandoor **2.6.13**, commit `51f1ccfabe61f1951c809f49ee7160effa0a9659`
- Docker `vabene1111/recipes`, PostgreSQL
- Still present in `develop` as of 2026-08-16 — `isFilterDefaultValue` and the `makenow`
  filter definition are unchanged there.

atirna:
## Description

I changed the search filter default check to use `Number.isNaN`. Previously, the coercing global `isNaN` treated the `"false"` and `"true"` Yes/No values as defaults; now an active value is included in the recipe request.

Fixes #4788

## Verification

- before: `makenow` with default `"false"` and value `"true"` was classified as default
- after: the focused SearchPage predicate check classifies that value as active while preserving unset numeric, NaN, and list defaults
- `npm run build` from `vue3`

### DIFF
--- a/vue3/src/pages/SearchPage.vue
+++ b/vue3/src/pages/SearchPage.vue
@@ -353,7 +353,7 @@ function enableFiltersWithValues() {
 function isFilterDefaultValue(filter: any) {
     if (Array.isArray(filter.default) && Array.isArray(filter.modelValue)) {
         return filter.default.length == filter.modelValue.length
-    } else if (isNaN(filter.default) && isNaN(filter.modelValue)) {
+    } else if (Number.isNaN(filter.default) && Number.isNaN(filter.modelValue)) {
         return true
     } else {
         return toRaw(filter.default) === filter.modelValue


vabene1111:
just saw this when making the changelog, sorry, already fixed but same way. Thanks for the PR

PR #4820, opened 2026-09-16
fixed cookmate local recipe image import

LakshmikanthanC:
Implemented the proper fix so Cookmate .mcb local photos are imported during normal import (no separate API pass needed).
Changes
cookbook/integration/cookmate.py:68-98 — Rewrote the image import step:
- Reads <imagepath> and loads the photo directly from the .mcb archive via self.safe_read(), falling back to a basename match if the stored path differs slightly.
- Falls back to the remote <imageurl> fetch only when no local image was found.
- Uses is not None checks instead of lxml truth-testing. This mattered: lxml 6.x reports elements containing only text (like <imagepath>images/a.jpg</imagepath>) as falsy, so a truthy-test would silently skip them. It also removes the AttributeError on self-closing empty elements.
- Detects the filetype (get_filetype) so PNG/WebP photos keep their format.
cookbook/integration/integration.py:191 — Sets self.import_zip = import_zip in the Cookmate branch of do_import (same pattern already used by the Gourmet integration), so get_recipe_from_file can reach the archive.
cookbook/tests/other/test_cookmate_integration.py — New tests covering: local image from archive, remote imageurl fallback (mock safe_request), and empty/self-closing <imagepath>/<imageurl> elements (no crash, no image).


azec19:
Top! It should fix the problem
Thanks for your time !!

